PR #554, opened 2015-02-16
Albers Equal Area projection

QuLogic:
Building on what I learned out of the Azimuthal Equidistant projection, I think I've managed to put this all together with no fuss. Once again, I have implemented the CRS, some tests, and added it to the documented list of projections.

projections-3

This time, I determined the boundaries a bit more rigorously, i.e., [semi-algebraically using SymPy](http://nbviewer.ipython.org/github/QuLogic/map-projections/blob/master/Albers%20Equal%20Area.ipynb#Cartopy-Boundaries). I'm not super happy about the various precisions that had to be used in the boundary tests, but I guess that a consequence of using a relatively coarse sampling of the boundary.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2014, Met Office
+# (C) British Crown Copyright 2011 - 2015, Met Office
 #
 # This file is part of cartopy.
 #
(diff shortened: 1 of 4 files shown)


lib/cartopy/crs.py
```diff
@@ -1531,6 +1531,85 @@ def y_limits(self):
...
+        """
+        Kwargs:
+
+            * central_longitude - The central longitude. Defaults to 0.
+            * central_latitude - The central latitude. Defaults to 0.
+            * false_easting - X offset from planar origin in metres.
+                              Defaults to 0.
+            * false_northing - Y offset from planar origin in metres.
+                               Defaults to 0.
+            * standard_parallels - The one or two latitudes of correct scale.
```
pelson: How about forcing this to be an iterable of length 1 or 2 and completely ruling out the possibility of this being a number?

lib/cartopy/crs.py
```diff
@@ -1531,6 +1531,85 @@ def y_limits(self):
...
+        self._x_limits = bounds[0], bounds[2]
+        self._y_limits = bounds[1], bounds[3]
+
+    @property
+    def boundary(self):
+        return self._boundary
+
+    @property
+    def threshold(self):
+        return 1e5
```
pelson: This could be a proportion of the x and y limits if necessary?

pelson:
Couple of interface questions - but I think this is pretty much good to go. This is really awesome! :+1:

lib/cartopy/crs.py
pelson: Unless I'm mistaken, this has also been called secant_latitudes in https://github.com/SciTools/cartopy/blob/master/lib/cartopy/crs.py#L1011. I'd like to have consistency in terminology - that either means changing this, or deprecating the other.

QuLogic: Yes, it appears that they should be the same thing. I don't really have a
preference either way.

lib/cartopy/crs.py
QuLogic: Do you have a suggestion for the relationship? Some projections make this
variable and others area fixed (e.g., the Lambert  Conformal you mentioned
is fixed.)

lib/cartopy/crs.py
pelson: I prefer standard parallels.

References: http://support.esri.com/en/knowledgebase/GISDictionary/term/standard%20parallel
And: http://www.geography.hunter.cuny.edu/~jochen/GTECH361/lectures/lecture04/concepts/Map%20coordinate%20systems/Lambert%20Conformal%20Conic.htm (specifically the image:
LambertConformalConic
)

lib/cartopy/crs.py
pelson: Work it based on the current limits you have? I'm happy to keep it as it is for now.

pelson:
I'll submit a PR shortly for fixing the Lambert code.

pelson:
Merged in #558. Great addition @QuLogic - thank you!

PR #600, opened 2015-03-31
shapely dependency

marqh:
replaces #598 
shapely dependency


rhattersley:
@pelson & @QuLogic - I've tagged v0.12.0rc2 with this change.

PR #627, opened 2015-06-18
Updated license headers

dkillick:


rhattersley:
1. Travis has gone bonkers.
2. These header failures **should** have been picked up earlier by Travis. ... Double-bonkers!

ajdawson:
The problem is the same as Iris had last year, the package installs are done individually and they keep swapping dependencies out from under one-another. E.g in the Python 2.7 build matplotlib etc are installed, but then the install of pillow swaps out the libpng version for a different one, and so on. The solution for iris was to create a requirements file and install from that.

ajdawson:
And for license header tests, doesn't the test suite need to be run from the git repository? Looks like the tests are always run from a different directory against the installed version, so the license header check will fail.

ajdawson:
>so the license header check will fail.

I don't mean fail, I mean "function correctly"...

PR #629, opened 2015-06-20
Limit libpng version installed on Travis.

QuLogic:
The matplotlib conda packages requires libpng15.so, but does not limit dependencies so that a new version is not installed.

This should reduce the number of build failures, though I'm not sure if they'll all go away.


ajdawson:
This won't work because the later install of pillow will changed the version. The smart way to approach this is probably a requirements file to make sure that dependencies are all resolved at the same time.

QuLogic:
OK, there are still image failures, but at least it's not crashing now.

ajdawson:
OK this will do for now to get us back on track I think.

PR #630, opened 2015-06-21
Update some web tile test images.

QuLogic:
Honestly, I can barely see the difference.


ajdawson:
Can you rebase this to pick up the Travis changes from your other PR and we'll see how we get on.

QuLogic:
Voilà.

ajdawson:
OK I don't really understand what the difference is or what caused it, but I'm content that this achieves peace and harmony re testing so I'm in...

QuLogic:
They're some web tiles, so I _think_ it's something going on on whatever server from which they get loaded.

Issue #648, opened 2015-07-20
Tissot indicatrices


MaxNoe:
Is there a way to get tissots indicatrices with cartopy?

Or another way to project a circle of an angular radius onto a map?
I'm drawing a star map with stereographic projection and want to add the field of view of our telescope.

QuLogic:
A `tissot` method was added to the `GeoAxes` in #668.

MaxNoe:
Great!

PR #668, opened 2015-09-07
Tissot method for geoaxes

scmc72:
Added method to draw Tissot's indicatrices onto the current axes and an example demonstrating this.

I haven't doc'd the **kwargs as I couldn't find a definitive list of what can and cannot be used in `cartopy.feature.ShapelyFeature`. I was wondering if anyone could shed any light on this for me?

#648

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -419,7 +419,7 @@ class GeoAxes(matplotlib.axes.Axes):
 
     def tissot(self, rad_km=5e5, n_samples=80, lon_n=10, lat_n=6, **kwargs):
         """
-        Adds Tissot's indicatrices to the current axes using the geodesic class.
+        Adds Tissot's indicatrices to the current axes with the geodesic class.
 
         Kwargs:
 
@@ -439,7 +439,7 @@ class GeoAxes(matplotlib.axes.Axes):
         for lat in np.linspace(-80, 80, lat_n):
             for lon in np.linspace(-180, 180, lon_n, endpoint=False):
                 circle = geod.circle(lon, lat, rad_km, n_samples=n_samples)
-                geoms.append(sgeom.Polygon(zip(circle[:,0],circle[:,1])))
+                geoms.append(sgeom.Polygon(zip(circle[:, 0], circle[:, 1])))
 
         feature = cartopy.feature.ShapelyFeature(geoms, ccrs.Geodetic(),
                                                  **kwargs)
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -49,7 +49,7 @@ import cartopy.mpl.feature_artist as feature_artist
 import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 from cartopy.vector_transform import vector_scalar_to_grid
-from cartopy import geodesic
+
 
 assert matplotlib.__version__ >= '1.2', ('Cartopy can only work with '
                                          'matplotlib 1.2 or greater.')
@@ -417,9 +417,10 @@ class GeoAxes(matplotlib.axes.Axes):
                                                       resolution, **kwargs)
         return self.add_feature(feature)
 
-    def tissot(self, rad_km=5e5, n_samples=80, lon_n=10, lat_n=6, **kwargs):
+    def tissot(self, rad_km=5e5, n_samples=80, lon_n=6, lat_n=6, lon_max=180,
+               lon_min=-180, lat_max=80, lat_min=-80, **kwargs):
         """
-        Adds Tissot's indicatrices to the current axes with the geodesic class.
+        Adds Tissot's indicatrices to the axes.
 
         Kwargs:
 
@@ -432,14 +433,30 @@ class GeoAxes(matplotlib.axes.Axes):
 
             * lat_n - Number of circles in the latitudinal direction.
 
+            * lon_max - Maximum extent in degrees in the longitudinal
+              direction.
+
+            * lon_min - Minimum extent in degrees in the longitudinal
+              direction.
+
+            * lat_max - Maximum extent in degrees in the latitudinal
+              direction.
+
+            * lat_min - Minimum extent in degrees in the latitudinal
+              direction.
+
+        **kwargs are passed through to `class:ShapelyFeature`.
+
         """
-        geod = cartopy.geodesic.Geodesic()
+        from cartopy import geodesic
+
+        geod = geodesic.Geodesic()
         geoms = []
 
-        for lat in np.linspace(-80, 80, lat_n):
-            for lon in np.linspace(-180, 180, lon_n, endpoint=False):
+        for lat in np.linspace(lat_min, lat_max, lat_n):
+            for lon in np.linspace(lon_min, lon_max, lon_n, endpoint=False):
                 circle = geod.circle(lon, lat, rad_km, n_samples=n_samples)
-                geoms.append(sgeom.Polygon(zip(circle[:, 0], circle[:, 1])))
+                geoms.append(sgeom.Polygon(circle))
 
         feature = cartopy.feature.ShapelyFeature(geoms, ccrs.Geodetic(),
                                                  **kwargs)
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -49,7 +49,7 @@ import cartopy.mpl.feature_artist as feature_artist
 import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 from cartopy.vector_transform import vector_scalar_to_grid
-
+from cartopy import geodesic
 
 assert matplotlib.__version__ >= '1.2', ('Cartopy can only work with '
                                          'matplotlib 1.2 or greater.')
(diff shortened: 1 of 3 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -417,6 +417,34 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
...
+
+            * rad_km - The radius in km of the the circles to be drawn.
+
+            * n_samples - Integer number of points sampled around the
+              circumference of each circle.
+
+            * lon_n - Number of circles in the longitudinal direction.
+
+            * lat_n - Number of circles in the latitudinal direction.
+
```
pelson: You can just state that ``**kwargs`` are passed through to `` `class:ShapelyFeature` ``.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -417,6 +417,34 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
                                                       resolution, **kwargs)
         return self.add_feature(feature)
 
+    def tissot(self, rad_km=5e5, n_samples=80, lon_n=10, lat_n=6, **kwargs):
+        """
+        Adds Tissot's indicatrices to the current axes with the geodesic class.
```
pelson: the word current is not needed here.

pelson: In fact the following would be perfectly good: 
```
Adds aTtissot indicatrix to the axes.
```

lib/cartopy/mpl/geoaxes.py
```diff
@@ -49,7 +49,7 @@
 import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 from cartopy.vector_transform import vector_scalar_to_grid
-
+from cartopy import geodesic
```
QuLogic: There should still be two newlines here for PEP8. Also, I don't know why you import it like this when it gets used as `cartopy.geodesic.Geodesic`.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -417,6 +417,34 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
...
+
+            * lon_n - Number of circles in the longitudinal direction.
+
+            * lat_n - Number of circles in the latitudinal direction.
+
+        """
+        geod = cartopy.geodesic.Geodesic()
+        geoms = []
+
+        for lat in np.linspace(-80, 80, lat_n):
```
QuLogic: It may be easiest to just provide an argument, but I think these limits should be determined by the extent of the map in some way. For example, say someone plots azimuthal equidistant centred on the North pole; they probably don't care too much about -80, but 85 or something would be visible.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -417,6 +417,34 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
...
+            for lon in np.linspace(-180, 180, lon_n, endpoint=False):
+                circle = geod.circle(lon, lat, rad_km, n_samples=n_samples)
+                geoms.append(sgeom.Polygon(zip(circle[:, 0], circle[:, 1])))
```
QuLogic: Is it really necessary to do this `zip` thing? If `shapely` is just iterating through the argument, it'd end up with the same result if it was just `circle`.

lib/cartopy/mpl/geoaxes.py
scmc72: Ahh okay. From a user perspective though, it doesn't seem like it is very easy to find exactly which `**kwargs` can be used. The doc for `ShapelyFeature` just says:

`Kwargs:
    Keyword arguments to be used when drawing this feature. `

How is someone supposed to find a list of usable `**kwargs`? Am I missing something?

lib/cartopy/mpl/geoaxes.py
scmc72: Ah - my bad. Should just be `geodesic.Geodesic`.

Just a thought - does it make more sense to put the import within the tissot method, so that it isn't required for geoaxes to work as discussed in #665?

lib/cartopy/mpl/geoaxes.py
scmc72: Ah - very good point - I hadn't thought of that! I agree that using the extent of the map would be the neatest way to get these numbers, but I'm not sure how easy that would be to implement.. 

I'll have a look into it, but if it seems like too much work, then simply adding an argument is also a viable option.

lib/cartopy/mpl/geoaxes.py
scmc72: Yes.. That seems to work. I was under the impression that Polygon needed a list of tuples, but apparently not.

lib/cartopy/mpl/geoaxes.py
scmc72: :+1:

lib/cartopy/mpl/geoaxes.py
pelson: I was actually considering suggesting that rather than specifying ```lon_n``` ```lat_n``` we simply put the circles ```rad_km``` apart, working out from the equator and central meridian. 

I haven't looked at how well it would work, but I'd also consider putting the tissot on each of the projections in the [projection list](http://scitools.org.uk/cartopy/docs/latest/crs/projections.html). (see https://github.com/SciTools/cartopy/blob/master/docs/make_projection.py#L97)

scmc72: Yeah - that could make the interface easier to use, although it would give users slightly less flexibility. It could default to being radially spaced, but have the option for a user to specify how many they want?

I'll have a look at the projection list and see if might work.

pelson: I discussed the equidistant circles thing with @scmc72, and we decided to leave that for another day.

scmc72: > I haven't looked at how well it would work, but I'd also consider putting the tissot on each of the projections in the projection list. (see https://github.com/SciTools/cartopy/blob/master/docs/make_projection.py#L97)

Having had a fiddle, it seems that it would definitely be possible to put indicatrices on each of the projections - it definitely adds to the images, making distortions more clear in a lot of cases.

There is one thing bugging me, which is that the convention for the indicatrices is to place them on gridline intersections like this:
tissot_world_from_space

I can't work out if there is  a way to get at these intersections. As far as I can tell, when you call `ax.gridlines()` unless you specify it, the spacing of the gridlines is shipped off to matplotlib land and I'm not really sure how to get at it.. Any thoughts?

ajdawson: Perhaps specifying where you want the grid lines to be would produce a more consistent result anyway?

scmc72: That would definitely work - I would just have to specify the grid line spacing for each projection, which isn't a problem. Thanks! 

I would still be interested to know if there is a way to get at the automatically spaced grid lines though. It seems like it should be straight forward...

scmc72:
So I've had a play around with tissot drawings with the possibility of adding them to the projections list. Here are a few of the things I came up with:
projection_list_demo

Would this kind of thing be a good addition to the projections list? If so, should I change anything? Are the grid lines generally a good spacing? Thoughts on colours? 

Cheers!

QuLogic:
Are those ellipses centred at the same longitudes and latitudes of the grid lines? They seem just a slight bit off.

pelson:
> Would this kind of thing be a good addition to the projections list? If so, should I change anything? Are the grid lines generally a good spacing? Thoughts on colours? 

I think it looks pretty good as is. We can always iterate afterwards though.

Now I've seen them, I would definitely like them in the projection list. :+1:

lib/cartopy/mpl/geoaxes.py
pelson: That is a problem with the ShapelyFeature documentation, rather than the tissot documentation. You could be explicit, and state that the kwargs are passed through to the underlying PathCollection (https://github.com/SciTools/cartopy/blob/master/lib/cartopy/mpl/feature_artist.py#L131)?

scmc72:
I think @QuLogic is right. Some of them are slightly off - I haven't worked out why yet...

I'll push up my working projections asap!

QuLogic:
Maybe because the default globe is different between Cartopy and the Geodesic code? I think it should be WGS84.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -417,6 +417,51 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
                                                       resolution, **kwargs)
         return self.add_feature(feature)
 
+    def tissot(self, rad_km=5e5, n_samples=80, lon_n=6, lat_n=6, lon_max=180,
```
pelson: I'm now inclined to suggest simplifying the arguments to:

```
def tissot(self, lons=None, lats=None, radius=5000, circle_n_samples=80):
    if lons is None:
        lons = np.linspace(-180, 180, 6)
    if lats is None:
        lats = np.linspace(-80, 80, 6)

    if lons.ndim == 1 or lats.ndim == 1:
    lons, lats = lons.flatten(), lats.flatten()

    if lons.shape != lats.shape:
        raise ValueError('lons and lats must have the same shape.')

    ...
```

scmc72: >I'm now inclined to suggest simplifying the arguments to...

Yeah that makes sense - gives a more transparent interface, and more flexibility. Did you mean to change the radius to 5000?

>Maybe because the default globe is different between Cartopy and the Geodesic code? I think it should be WGS84.

I hadn't considered that, although I would be surprised if that was the case as I definitely used the WGS84 parameters in geodesic. Does cartopy definitely use this?

pelson: PlateCarree will be a spherical approximation, from memory.

> Did you mean to change the radius to 5000?

Nope. Thanks for asking!

PR #687, opened 2015-10-25
Geodesic optimizations

QuLogic:
Cython correctly warns that bounds checking can be disabled in the geodesic methods. The reason is that the methods check dimensions, allocate arrays, and loop through the same length thus ensuring bounds will never be exceeded.

There's also a minor cleanup of trailing whitespace.


rhattersley:
:+1: Travis-CI failure is nothing to do with this PR.

PR #693, opened 2015-11-09
Fix license typo

QuLogic:
`docs/source/copyright.rst` says _Iris_, not Cartopy. Also, update the dates and word wrapping while I'm there.


QuLogic:
Oh, hmm, I should have put that into v0.13.x...

PR #702, opened 2015-11-19
Remove documentation of 'target_extent' argument to imshow

duncanwp:
This isn't actually an accepted argument by the looks of it


rhattersley:
:+1:

There's no explicit usage within this routine, and the `kwargs` dict is only passed on to `matplotlib.axes.Axes.imshow()` ... but matplotlib doesn't use the target_extent keyword [anywhere](https://github.com/matplotlib/matplotlib/search?utf8=%E2%9C%93&q=target_extent).

rhattersley:
NB. Test failures are due to the web tiles issue.

rhattersley:
Thanks @duncanwp :smile:

duncanwp:
No problem :-)

PR #737, opened 2016-03-14
Aurora example

belteshassar:
aurora2

I thought the colors in the new viridis colormap were similar to aurora so I decided to use it to plot the aurora. It sources data from a gist, but I also included a commented line for downloading live data. I'll probably update the gist when I can catch a file from a day of elevated auroral activity to make the example more impressive. The data comes from [NOAA's Space Weather Prediction Center](http://swpc.noaa.gov).

For compatibility with matplotlib < 1.5 I use an alternative colormap if viridis can't load.

### DIFF
--- a/lib/cartopy/examples/aurora_forecast.py
+++ b/lib/cartopy/examples/aurora_forecast.py
@@ -21,8 +21,11 @@ except ImportError:
 from io import StringIO
 
 import numpy as np
+from datetime import datetime
 import cartopy.crs as ccrs
 import matplotlib.pyplot as plt
+from matplotlib.colors import LinearSegmentedColormap
+import matplotlib.patches as patches
 
 
 def aurora_forecast():
@@ -40,6 +43,8 @@ def aurora_forecast():
         the ``img_proj`` coordinate system.
     origin : str
         The origin of the image to be passed through to matplotlib's imshow.
+    dt : datetime
+        Time of forecast validity.
 
     """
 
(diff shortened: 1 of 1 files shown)


pelson:
Cool. Nice idea! I'm definitely up for having this kind of data in the gallery.

Example

rhattersley:
> I thought the colors in the new viridis colormap were similar to aurora so I decided to use it to plot the aurora.

Nice idea! :+1:

belteshassar:
> I quite like the fact we have the underlying blue marble in the example above though - I wonder if we can add some levels on the data so that we can see blue marble where there are low data values.

I thought about that while working on this, but I didn't come up with any solution. Solving that would really add to the example though so it's worth putting some thought into.

belteshassar:
I pushed another commit to fix the build error in Python 3. Only tested in Python 2.7 on my machine :confused:

lib/cartopy/examples/aurora_forecast.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+__tags__ = ["Scalar data"]
+try:
+    from urllib2 import urlopen
+except ImportError:
+    from urllib.request import urlopen
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
QuLogic: Just use `io` everywhere. It exists on Python 2, too.

lib/cartopy/examples/aurora_forecast.py
```diff
@@ -0,0 +1,88 @@
+"""
+Plotting the Aurora Forecast from NOAA on Orthographic Polar Projection
+---------------------------------------------------
```
QuLogic: Underline not long enough.

lib/cartopy/examples/aurora_forecast.py
```diff
@@ -0,0 +1,88 @@
...
+    # ax1 for Northern Hemisphere
+    ax1 = plt.subplot(1, 2, 1, projection=ccrs.Orthographic(0, 90))
+    # ax2 for Southern Hemisphere
+    ax2 = plt.subplot(1, 2, 2, projection=ccrs.Orthographic(0, -90))
+
+    img, crs, extent, origin = aurora_forecast()
+    for ax in [ax1, ax2]:
+        ax.coastlines()
+        ax.gridlines()
+        plt.sca(ax)
```
QuLogic: Not necessary.

lib/cartopy/examples/aurora_forecast.py
```diff
@@ -0,0 +1,88 @@
...
+        plt.imshow(img, vmin=0, vmax=100, transform=crs,
```
QuLogic: Use `ax.imshow`.

lib/cartopy/examples/aurora_forecast.py
```diff
@@ -0,0 +1,88 @@
...
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+import numpy as np
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+try:
+    from matplotlib.cm import viridis as cmap
```
belteshassar: I should probably use `get_cmap()` instead.

belteshassar:
Geomagnetic activity is high at the moment. I've updated the gist as well as the image at the top of this thread.

belteshassar:
I'll refactor the code according to the suggestions by @QuLogic and experiment a little with custom colormaps to see if I can get something that looks nice on top of the Blue Marble.

I'd also like to add the day-night terminator, but I don't want to bring in astropy so I'll probably approximate the Sun's declination as a sine function with amplitude 23.4 degrees and period of one year. The equinoxes would be off by about four days if I do that, so depending on how visible that is, I might go for something slightly more complicated.

belteshassar:
@pelson I've done some work on this. I'm still not satisfied with the colormap, but it works as proof of concept at least. Do you think I'm on the right path?

aurora3

ocefpaf:
>  I'm still not satisfied with the colormap

Why not? It looks awesome to me. Nice example!

pelson:
> Why not? It looks awesome to me. Nice example!

Agreed. You haven't pushed those changes up yet, did you know? I was curious about the terminator calculation and how you have visualised it.

belteshassar:
> You haven't pushed those changes up yet, did you know? I was curious about the terminator calculation and how you have visualised it.

Finally got my act together and finalized it. I pushed two commits, first is just a cleanup of the old viridis example. Second is the new one with natural earth and shadowed nightside.

The terminator calculation is a bit ugly and not very accurate, but I think it works for illustration purposes. I just plotted a 90x90 square on a rotated pole projection. The parameters of the rotated pole are determined from the position of the suns zenith and the shadow is just a 75% opaque. Declination of the sun is just a cosine with amplitude 23.4 degrees. This ignores the fact that Earth's orbit is slightly elliptical meaning that winters are in fact slightly longer in the northern hemisphere so the equinoxes are off by a few days.

pelson:
> The terminator calculation is a bit ugly and not very accurate, but I think it works for illustration purposes. I just plotted a 90x90 square on a rotated pole projection. The parameters of the rotated pole are determined from the position of the suns zenith and the shadow is just a 75% opaque. Declination of the sun is just a cosine with amplitude 23.4 degrees. This ignores the fact that Earth's orbit is slightly elliptical meaning that winters are in fact slightly longer in the northern hemisphere so the equinoxes are off by a few days.

Good enough for me. One day I'd love to get the terminator in vector form so that we can visualise it as polygons, for now, this is :+1:

pelson:
What a beautiful contribution. Thanks @belteshassar! If you have any more of those up your sleeve, I'd be happy to merge! :+1:

belteshassar:
Thanks @pelson. And thanks for challenging me to improve the original.

PR #773, opened 2016-06-05
Matplotlib 2.0 compatibility

QuLogic:
A few patches to ensure matplotlib 2.0 compatibility:
- Import `matplotlib.testing.setup` instead of `matplotlib.tests.setup` because the latter sometimes doesn't get packaged in conda.
- Enable the "classic" mpl style; I'm not sure what this will do with older versions, so we'll see how Travis copes.
- Set `FeatureArtist` to use a 1.0 line width, since `PathCollection` uses 0 width edges in mpl 2.0.

Those changes leave us with 20 remaining image failures. Some of these are minor textual things, and some are differences which I think arose from the change to the image infrastructure. I haven't looked too deeply at those yet.


QuLogic:
OK, as expected, this doesn't work too well with older releases; I'll look into that tomorrow, then.

But the files that do still have issues seem odd to me. For examples, this diff here gives an RMS of 14.5921332259:
result-lambert_conformal_south-failed-diff
but if I run the two images through ImageMagick's compare program (@200%), there are far less differences:
diff

I wonder if there's a problem with the image comparison routine.

QuLogic:
Oops, scratch that last one; it's down to font issues, 
result-gridliner_labels-failed-diff
image handling?
result-imshow_natural_earth_ortho-failed-diff
pcolormesh interpolation (seems like a bug in the classic style):
result-pcolormesh_limited_area_wrap-failed-diff
and the correction of the streamplot zorder:
result-streamplot-failed-diff

pelson:
Amazing! Thanks @QuLogic! Ping me if there is anything you need from me, but it looks like this is progressing nicely! 👍 👍

QuLogic:
I just realized that the pcolormesh image doesn't work because the `Set1` colour map is supposed to be categorical, so it comes out "blocky" now. Changing colour maps is is probably the easiest fix there.

QuLogic:
Updated images except for the ones that change text. The reason is that Matplotlib 2.0.x branch has some tweaks to the text baseline in the classic style, and I want to see if that will fix the text to match our existing images. If that's the case, it'd be better to just modify the tolerance for Matplotlib 2.0.0 only.

dopplershift:
Is this still a WIP? Anything else holding this up?

Kind of annoyed right now because wanting an conda environment (from conda-forge) with matplotlib 2.0 and cartopy means having cartopy 0.14.3 and numpy 1.11 (which rolls back a bunch of other stuff). So anything I can do to help push along?

QuLogic:
I am waiting for matplotlib/matplotlib#7970 to see whether or not we need to change figures for the text movements or just up the tolerance temporarily.

lib/cartopy/mpl/gridliner.py
```diff
@@ -29,7 +29,7 @@
 from cartopy.crs import Projection, _RectangularProjection
 
 
-degree_locator = mticker.MaxNLocator(nbins=9, steps=[1, 2, 3, 6, 15, 18])
+degree_locator = mticker.MaxNLocator(nbins=9, steps=[1, 1.5, 1.8, 2, 3, 6, 10])
```
marqh: i am intrigued what these values do and how changing them is better for mpl2 but no worse for mpl1.5

this change seems to have resulted in a number of image changes.

please could you talk me through the logic for this change?

marqh:
the change to 
lib/cartopy/tests/mpl/baseline_images/mpl/test_images/imshow_regional_projected.png 
appears to be a subtle shift in location of image to coastline, which is a bit worrying, do you know why this might be?

marqh:
the colour change in
...rtopy/tests/mpl/baseline_images/mpl/test_mpl_integration/pcolormesh_limited_area_wrap.png 
is quite surprising

marqh:
@QuLogic 
please can you give me an extra sense of you approach here.

This PR is testing against mpl1.3 and mpl1.5 but not mpl2.  Are you proposing a set of 'no change' changes such that the test suite may be extended to use mpl2 without further test changes required?

thank you

lib/cartopy/mpl/gridliner.py
QuLogic: The values are documented as and required to start at 1 and end at 10; they only worked with mpl1.5 this way by chance. It does mean a change in the result, but they are consistent between mpl1.5 and mpl2.0.

dopplershift:
Is this ready to start testing against Matplotlib 2.0?

djkirkham:
Enabling the 'classic' style hides an issue with the axis margins. In Matplotlib 2.0 the default margin is 0.05 whereas in the old style there is no margin by default, although scatter plots (and perhaps others) do have a margin in both styles.

The default margin causes issues plotting maps; non-global plots have an ugly white area between the plot and the border, and global plots are smaller than before, and hovering off the edge of the map shows an extrapolation of the coordinates in the lower status bar.

Global plot:
```
>>> import matplotlib as mpl
>>> print mpl.__version__
2.0.0
>>> import matplotlib.pyplot as plt
>>> x = np.arange(0, 361, 10)
>>> y = np.arange(-90, 91, 10)
>>> c = x[None,:-1] * y[:-1,None] 
>>> plt.axes(projection=ccrs.PlateCarree())
<cartopy.mpl.geoaxes.GeoAxesSubplot object at 0x7f358bdeb910>
>>> plt.pcolormesh(x, y, c)
<matplotlib.collections.QuadMesh object at 0x7f358c158710>
>>> plt.show()
```
extra

Non-global:
```
>>> plt.axes(projection=ccrs.PlateCarree())
<cartopy.mpl.geoaxes.GeoAxesSubplot object at 0x7f358c18cb50>
>>> plt.pcolormesh(x[4:9], y[4:9], c[4:8,4:8])
<matplotlib.collections.QuadMesh object at 0x7f358c18c410>
```
non_global

QuLogic:
@djkirkham I have corrected that issue in 03e0214; the autolimits actually produced a view that was outside the projection's limits. This was a latent bug really, since old Matplotlib could have done the same margin thing, just with a different setting.

This PR now works on 1.5.3, 2.0.0, 2.0.1, 2.0.2, and the `v2.0.x` branch of Matplotlib. It almost passes on the `master` branch also, failing only two tests. Travis also shows it continues to work on 1.3.1, but I did not bother to test anything in between, and really would prefer if we dropped support for the old ones anyway.

Travis is currently failing due to external downloads/maps/etc., but none of those should be related to the tests from this PR.

QuLogic:
It's been almost a year since I opened this, though only about a month since it was complete.

ajdawson:
OK this one needs to go in finally. @QuLogic - can you fix the conflict and I'll merge.

QuLogic:
Rebased; I believe it should pass, though I didn't run the entire suite after.

ajdawson:
Thanks @QuLogic, much appreciated.

QuLogic:
Thanks for merging @ajdawson!

PR #775, opened 2016-06-06
Add jupyter notebooks based on existing examples

### LINKED ISSUE #388: Add examples & tutorials
Hi Guys,

Whenever you don't have time, it would be cool to add new examples and tutorials. I should be able to provide a notebook with the tutorial I gave in Brussels last summer, but it needs tweaking. Bill showed also a few nice examples during his presentation there, those were really cool !

And in tutorials: some like "how to create a map with a Belgian Lamber 72 (pyproj string is known)", etc... :-)

ahaberlie:
This was mentioned at least once in the issues (eg, closed #388). These notebooks could serve as a supplement/alternative to the existing examples (and the related gallery page).


QuLogic:
I think the main problem is not converting them to notebooks, but in getting those notebooks to show up in the compiled documentation.

PR #776, opened 2016-06-10
TST: Fix caching test error messages.

QuLogic:
The strings use %-markers, but calls .format() on the string.


QuLogic:
Ping? This is a pretty small fix.

PR #782, opened 2016-07-17
Fix invisible pcolormesh

QuLogic:
When running `ax.pcolormesh` with a grid that is fully transformed off the plot, it becomes infinity and is cut out into an empty array. This causes an issue with determining the extent.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1480,10 +1480,21 @@ class GeoAxes(matplotlib.axes.Axes):
             # END OF PATCH
             ##############
 
-        minx = np.amin(X)
-        maxx = np.amax(X)
-        miny = np.amin(Y)
-        maxy = np.amax(Y)
+        ########################
+        # PATCH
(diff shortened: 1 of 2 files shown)


marqh:
i think i understand this, although I have not experienced the failure mode

i'm happy to take it as a bug fix

thanks @QuLogic

marqh:
@corinnebosley 

is this a similar case to the one you were chasing?  please could you evaluate this fix compared to your failure cases?
it will be in the 0.15 release

thank you

corinnebosley:
@marqh I am a little concerned about this.  It seems a bit like a fix for another symptom of the same major bug, rather than solving the problem itself.  What I would like to be doing is fixing the proj4 calls so that we don't get any np.infs at all.

QuLogic:
If the mesh is not visible, then what is it supposed to transform into, other than `np.inf`?

corinnebosley:
@QuLogic I wasn't slating your fix.  I don't fully understand the bug that you fixed because you haven't posted an example of it.  I was really referring to the problems I have experienced in different scenarios which return np.infs when a real number should be returned instead.

PR #788, opened 2016-08-01
Add WMTS kwargs to add_wmts() method

batsc:
Following the structure of the add_wms() method, I've added 'gettile_extra_kwargs' to the add_wmts() call, which allows one to pass arguments to the owslib.wmts.gettile() method. This is useful as it allows the specification of the 'time' dimension when requesting WMTS layers that support it.

I've added an example to demonstrate the functionality.

A couple more minor changes:

* Added support for EPSG 3413 (WGS 84 / NSIDC Sea Ice Polar Stereographic North) for 'add_wmts()'
* Corrected comments in the geostationary image example

Cheers

### DIFF
--- a/docs/source/contributors.rst
+++ b/docs/source/contributors.rst
@@ -34,6 +34,7 @@ the package wouldn't be as rich or diverse as it is today:
  * Louis Tiao
  * Zachary Tessler
  * Daniel Eriksson
+ * Crispian Batstone
 
 
 Thank you!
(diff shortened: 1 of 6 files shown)


batsc:
I forgot to mention: OWSlib needs to be 0.8.11 otherwise owslib.wmts.gettile fails when supplying the time keyword.

batsc:
I could be wrong but I don't think those Travis errors/failures are related to my changes.

marqh:
Hello @batsc 

I have been through your changes and they look great.

the test failures are nothing to do with your code and I have not yet found a fix to push to master to get the tests on this PR to pass

I'm keen to see this change adopted, so I'm going to merge this, rather than wait until i've finished my investigation into mapquest images (tl;dr test failing)

PR #794, opened 2016-09-09
correct unit error in geoaxes.tissot() (m->km)

skjaeve:
The tissot() docstring says the radius input is in km, but the underlying Geodesic.circle() function takes it to be meters. This patch adds a conversion from meters to kilometers.

An alternative solution would be to change the docstring to say meters; this wouldn't break existing code using tissot().


lib/cartopy/mpl/geoaxes.py
```diff
@@ -418,7 +418,7 @@ def coastlines(self, resolution='110m', color='black', **kwargs):
                                                       resolution, **kwargs)
         return self.add_feature(feature)
 
-    def tissot(self, rad_km=5e5, lons=None, lats=None, n_samples=80, **kwargs):
+    def tissot(self, rad_km=5, lons=None, lats=None, n_samples=80, **kwargs):
```
QuLogic: The equivalent is 500, I guess, not 5.

QuLogic:
Ping @scmc72.

QuLogic:
Ping @scmc72.

QuLogic:
Ping @skjaeve (have I been pinging the wrong person all this time?)

lib/cartopy/mpl/geoaxes.py
skjaeve: Oops, you're absolutely right.

skjaeve:
Hello. Sorry, I'm not familiar with how pull requests work. Should I change the request?
Would  you prefer that the number is correct, or change the docstring?

skjaeve:
Fixed default argument committed.

ajdawson:
This PR has gone a bit wrong, looks like an accidental merge of master into your branch. @skjaeve - are you able to get this back into shape, we should only be seeing the commits you made for this PR (I think 0d2d518 and 6ddbd92, but please check).

ajdawson:
Also, we'll need you to complete the [CLA](http://scitools.org.uk/documents/cla_form.pdf) before we can merge the PR @skjaeve. For general information on CLAs see http://oss-watch.ac.uk/resources/cla.

skjaeve:
Yup, I bungled it. I didn't realise merging changes from the upstream branch would also mess up the pull request (I thought commits already in the upstream code would be filtered out, since there's nothing to merge.) The simplest solution is probably to withdraw this pull request and make another one.

skjaeve:
See request #904 instead.

PR #809, opened 2016-10-19
numpy 1.10+ linspace nan fix

bjlittle:
This PR smooths over a corner case exception that can be raised within `img_transform.mesh_projection` for a **single point** that requires to be projected and using numpy **v1.10+**.

In this case, for numpy **v1.9** we have:
```python
>>> import numpy as np
>>> np.__version__
'1.9.3'
>>> np.linspace(-1, 1, 1, retstep=True, endpoint=False)
(array([-1.]), 2.0)
>>> 
```
and for numpy **v1.10+** we have:
```python
>>> import numpy as np
>>> np.__version__
'1.10.4'
>>> np.linspace(-1, 1, 1, retstep=True, endpoint=False)
(array([-1.]), nan)
>>> 
```

### DIFF
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -81,6 +81,14 @@ def mesh_projection(projection, nx, ny,
     y, ystep = np.linspace(y_lower, y_upper, ny, retstep=True,
                            endpoint=False)
 
+    # Deal with single point corner case and the difference
+    # between np.linspace v1.9 and v1.10+ retstep nan result.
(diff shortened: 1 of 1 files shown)


bjlittle:
@marqh Ping ...

marqh:
all test failures are recognised #810

this is a bug fix with limited scope, but needed for up to date numpy versions for some cases

I think this is a useful bug fix and we should put it on a 0.14.3 as soon as practical

dkillick:
The fix looks good to me, if a little annoying to have to include in Cartopy...

PR #814, opened 2016-11-04
BUG: Fix contourf bounds calculation (Fixes #811)

### LINKED ISSUE #811: Problem with contourf when displaying data with Cartopy
In certain cases (including but not necessarily limited to increasing the # of fill intervals and adding the "extend" option for the colorbar), I end up with a frame that contains just a single color, with no geography showing.

The problem is illustrated with this Jupyter notebook:

http://nbviewer.jupyter.org/url/www.atmos.albany.edu/facstaff/ktyle/notebooks/Cartopy_Issue.ipynb?flush_cache=true

matplotlib: v 1.5.1
cartopy: v0.14.2

dopplershift:
Previously, if a path returned from contourf was empty, the calculation
of the bounds would go wrong, resulting in poor bounds (*extremely*
zoomed in view). This would often show when specifying arbitrary contour
levels.

### DIFF
--- a/lib/cartopy/tests/mpl/test_contour.py
+++ b/lib/cartopy/tests/mpl/test_contour.py
@@ -1,0 +1,45 @@
+# (C) British Crown Copyright 2016, Met Office
+#
+# This file is part of cartopy.
+#
+# cartopy is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cartopy is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


dopplershift:
Test failures/errors look unrelated to me.

lib/cartopy/tests/mpl/test_contour.py
```diff
@@ -0,0 +1,45 @@
+# (C) British Crown Copyright 2015 - 2016, Met Office
```
QuLogic: Technically, this file is 2016 only...

dopplershift:
Fixed the copyright year.

ajdawson:
@dopplershift - Have you completed a CLA form before? I don't see your id listed at http://scitools.org.uk/about. CLA form: http://scitools.org.uk/documents/cla_form.pdf.

dopplershift:
I hadn't before. Now that I finally have UCAR legal's blessing, I emailed a signed copy last night.

QuLogic:
CLA is in, finally.

PR #827, opened 2016-12-10
Fix testing against OWSLib 0.13.0.

QuLogic:
The `WebMapService` is a function in that version, so we can't run `isinstance` against it.

I assume this is not triggered on Travis because we're still using the scitools channel which doesn't have an updated owslib. Would it make sense to switch to conda-forge now?


marqh:
thank you for the update @QuLogic 

this looks fine to me

PR #830, opened 2016-12-21
removed import of missing PROJ4_VERSION

corinnebosley:
I don't know when this changed, but today I fetched upstream master to cut a branch, and all of a sudden I got import errors because PROJ4_VERSION is not now living in '_crs.pyx'.

I just removed the import, and that worked for me.  Not tested this extensively yet.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -33,7 +33,7 @@ import shapely.geometry as sgeom
 from shapely.prepared import prep
 import six
 
-from cartopy._crs import CRS, Geocentric, Geodetic, Globe, PROJ4_VERSION
+from cartopy._crs import CRS, Geocentric, Geodetic, Globe
 import cartopy.trace
 
 


corinnebosley:
I can now see that this will fail as PROJ4_VERSION is referenced in class Robinson.

corinnebosley:
@QuLogic It looks like this was changed in April, when you factored out duplicate parsing of the proj4 version.  

I am confused though, have you had any import problems with this since you made the change?
(https://github.com/SciTools/cartopy/commit/e77ceadbaf6513c8bbfa688fc348fc729b7294de)

ajdawson:
This all looks fine to me, I don't think any change is necessary... Are you sure you re-compiled the `_crs` cython module before testing? If you had the old `_crs` module you would see an import error.

corinnebosley:
@ajdawson No, that's not something I did.  Thank you.

PR #831, opened 2017-01-03
test added to show faulty proj4 params

corinnebosley:
For several projections, the call to proj4 in cartopy geoaxes.py transform_points uses params which are out of bounds.  This causes proj4 to return null coordinates and hence not be able to produce a plot, although no error is shown to the user.

This test should fail because it uses an example of the invalid coordinates which are passed to proj4.


corinnebosley:
I am having a little trouble working out which order lats and lons need to be in to return them in the right order from proj4 direct.  Working on it...

PR #834, opened 2017-01-05
Travis' python 3 restricted to 3.5 and badge added for build status

corinnebosley:


README.md
```diff
@@ -2,6 +2,7 @@
 
 Cartopy
 -------
+[![Build Status](https://api.travis-ci.org/repositories/SciTools/cartopy.svg?branch=master)](http://travis-ci.org/SciTools/cartopy/branches)
```
corinnebosley: @marqh I thought this would be better underneath the title, rather than above it.  Seems a bit more consistent with Iris this way.

QuLogic: https link.

marqh: the http just redirects to https at travis, so I don't think this matters.

This is needed so that we can respin tests, which are currently failing due to this test matrix change.

i'm going to merge this

it's easy to alter the URI in the readme as required

@QuLogic this is the same as on the Iris readme:
https://raw.githubusercontent.com/SciTools/iris/master/README.md
which this pattern was taken from.
would you prefer that the URIs were consistent and using https on both of these repositories?

mark

QuLogic: Yes, one less http link is one less link that can be MITM'd and one less request for the user that clicks it.

PR #847, opened 2017-02-01
Fix gallery examples for release.

lbdreyer:
There have been a few problems with MapQuest and SRTM requiring user accounts to use the service - not good for gallery examples!

This replaces the MapQuest tilers with OpenStreetMap or Stamen.

This removes SRTM altogether as I can't think of a simple workaround and we we want to get this Cartopy release out soon. Hopefully we will be able to come up with a solution and reinstate the SRTM example in a later release.

The new images are as follows:
effects_of_the_ellipse

effects_of_the_ellipse_00_00
eyja_volcano

eyja_volcano_00_00

tube_stations
tube_stations_00_00


bjlittle:
@lbdreyer I'm :+1: for this change, as it avoids the use account issue but still allows us to showcase the capability in the gallery.

@marqh @ajdawson Any objections?

ajdawson:
:+1:

marqh:
looks like happy faces all round

mergetastic

PR #850, opened 2017-02-02
Update other version numbers .

lbdreyer:
Found a couple other version numbers that also need updating


setup.py
```diff
@@ -335,7 +335,7 @@ def get_config_var(name):
 # ==========
 setup(
     name='Cartopy',
-    version='0.14.4.dev0',
+    version='0.15.0',
```
bjlittle: @lbdreyer Totally not your doing btw, but it seems a little crazy to me that devs need to remember to do this on release in several places ...

The `cartopy.__version__` should suffice here. It's completely discoverable.

If you were to implement a scheme similar to iris, see [here](https://github.com/SciTools/iris/blob/master/setup.py#L183-L193) and [here](https://github.com/SciTools/iris/blob/master/setup.py#L198) 

I think this approach might be a step forward in the right direction, and perhaps prevent any future mishaps .. what do you think? :thinking:

lbdreyer: I agree, I will add the Iris approach.

Unfortunately this doesn't get around the issue of having to update the version switcher

bjlittle: @lbdreyer What if the `version_switch.js` was to read a json file in order to initialise the `all_versions` ?

The json file could be generated as part of `make html` or thought some other mechanism?

Perhaps not part of this PR, but I'll raise an issue to that this isn't forgotten.

bjlittle:
@lbdreyer Great thanks! :+1:

PR #857, opened 2017-02-17
Add OSGB as OGC SRS

dkillick:
Add the OSGB projection to the dictionary of CRS' that web services can be returned natively in.


lbdreyer:
Looks good! Thanks @dkillick

dkillick:
Thanks @lbdreyer :grin:

PR #860, opened 2017-02-27
Update v0.15.x

dkillick:
Sucks a couple of bugfixes onto v0.15.x, where they should have been merged to originally. I feel some dev notes on bugfixes and releases coming on...

Changes:

 * add `lon_0` argument to Gnomic projection.
 * add OSGB to native SRS' for WMS/WMTS retrieval.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2016, Met Office
+# (C) British Crown Copyright 2011 - 2017, Met Office
 #
 # This file is part of cartopy.
 #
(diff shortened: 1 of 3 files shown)


ajdawson:
> where they should have been merged to originally

Neither of these are really bug fixes are they?

> add lon_0 argument to Gnomic projection.

This isn't at least (actually it is an API break, but let's not get into that).

dkillick:
Maybe they aren't bug fixes, but they aren't breaking changes either. In fact, I think it is easier to argue the Gnomic change is a bugfix because it is correcting a part of the API **that should have been there in the first place**.

Pragmatically, they're both tiny fixes that _people need_. Shall we tell them they're going to have to wait three months for us to do a proper release?

ajdawson:
Well, I hope it won't be 3 months until 0.16, but that's probably another discussion! I just wanted to make sure you were aware that the Gnomic change could break existing code, so you can make sure you have considered that properly before putting it in a bugfix release.

dkillick:
> that's probably another discussion

Quite...

> you were aware that the Gnomic change could break existing code

We had discussed this. If the were adding a new arg I wouldn't even consider it but given it adds a kwarg I'm not as concerned as kwargs are very opt-in.

ajdawson:
Specifically, things that look like this will be broken by this change:

```python
central_latitude = 55.0
globe = ccrs.Globe(...)

proj = Gnomic(central_latitude, globe)
```

because the `globe` argument will now be interpreted as a central longitude (and fail).

dkillick:
Ah, the globe. Alright, can I at least release my change?

ajdawson:
You can do whatever you like, I'm not here to forbid you from doing this, I just wanted to make sure the decision had been fully considered.

Your change looks extremely innocent, I can't think of a way it will break anything.

dkillick:
>  I just wanted to make sure the decision had been fully considered

Which it evidently hadn't :scream: Thanks @ajdawson!

PR #864, opened 2017-02-28
Bump version for post-release development

lbdreyer:
In a similar vein to #763 this bumps the version number to be 0.15.2dev0 (or should it be 0.15.2dev????)

### DIFF
--- a/lib/cartopy/__init__.py
+++ b/lib/cartopy/__init__.py
@@ -17,7 +17,7 @@
 
 from __future__ import (absolute_import, division, print_function)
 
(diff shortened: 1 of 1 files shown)


dkillick:
Personally, I'd have gone for `v0.15.2-DEV0`... but it's much of a muchness. @bjlittle any thoughts? There seems little point getting much hung up on this.

bjlittle:
@lbdreyer has gone for [pep440](https://www.python.org/dev/peps/pep-0440/#developmental-releases), which looks good to me :+1:

PR #870, opened 2017-03-16
adopt conda-forge

marqh:
I think that we can adopt conda-forge for ongoing continuous integration testing, if we adopt the test fixes in this PR at the same time.

the only non-backwards compatible one is the numerical changes in: `lib/cartopy/tests/test_crs.py` `test_epsg`
I still haven't found out why this might be occurring

closely related to
https://github.com/SciTools/iris/pull/2440

So, is now the time to move to conda-forge for Travis CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ install:
 
   # Customise the testing environment
   # ---------------------------------
-  - conda config --add channels scitools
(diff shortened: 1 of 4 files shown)


bjlittle:
Thanks @marqh :+1:

Issue #883, opened 2017-05-18
contourf: ValueError: A LinearRing must have at least 3 coordinate tuples


akrherz:
I've been migrating from basemap to cartopy and a code path that works in basemap, will now error in cartopy like so:
```
Traceback (most recent call last):
  File "current_bias.py", line 111, in <module>
    main()
  File "current_bias.py", line 108, in main
    doit(ts, model)
  File "current_bias.py", line 76, in doit
    m.contourf(lons, lats, vals, range(-10, 11, 2), units='F', cmap=cmap)
  File "/opt/miniconda2/lib/python2.7/site-packages/pyiem/plot/geoplot.py",
line 863, in contourf
    extend='both', transform=ccrs.PlateCarree())
  File "/opt/miniconda2/lib/python2.7/site-packages/cartopy/mpl/geoaxes.py",
line 1339, in contourf
    if col.get_paths()])
  File
"/opt/miniconda2/lib/python2.7/site-packages/matplotlib/collections.py",
line 213, in get_datalim
    paths = [transform.transform_path_non_affine(p) for p in paths]
  File
"/opt/miniconda2/lib/python2.7/site-packages/matplotlib/transforms.py", line
2386, in transform_path_non_affine
    return self._a.transform_path_non_affine(path)
  File "/opt/miniconda2/lib/python2.7/site-packages/cartopy/mpl/geoaxes.py",
line 183, in transform_path_non_affine
    getattr(self, 'force_path_ccw', False))
  File "/opt/miniconda2/lib/python2.7/site-packages/cartopy/mpl/patch.py",
line 192, in path_to_geos
    geom = sgeom.Polygon(path_verts[:-1, :])
  File
"/opt/miniconda2/lib/python2.7/site-packages/shapely/geometry/polygon.py",
line 229, in __init__
    self._geom, self._ndim = geos_polygon_from_py(shell, holes)
  File
"/opt/miniconda2/lib/python2.7/site-packages/shapely/geometry/polygon.py",
line 508, in geos_polygon_from_py
    geos_shell, ndim = geos_linearring_from_py(shell)
  File "shapely/speedups/_speedups.pyx", line 214, in
shapely.speedups._speedups.geos_linearring_from_py
(shapely/speedups/_speedups.c:4017)
ValueError: A LinearRing must have at least 3 coordinate tuples
```

[...]

PR #885, opened 2017-05-22
polygons should have 4+ points #883

akrherz:


lib/cartopy/mpl/patch.py
```diff
@@ -185,13 +185,10 @@ def path_to_geos(path, force_ccw=False):
                                      axis=1)
         if all(verts_same_as_first):
             geom = sgeom.Point(path_verts[0, :])
-        elif (path_verts.shape[0] > 2 and
+        elif (path_verts.shape[0] > 3 and
                 (path_codes[-1] == Path.CLOSEPOLY or
                  verts_same_as_first[-1])):
-            if path_codes[-1] == Path.CLOSEPOLY:
```
QuLogic: For `CLOSEPOLY`, the last location should be ignored; see [Matplotlib's Path tutorial](https://matplotlib.org/users/path_tutorial.html). Thus, I don't think this condition should be removed, unless you can guarantee that the last point is at the first point (which seems unlikely given the `or verts_same_as_first[-1]` bit.)

QuLogic:
Looks reasonable, except for the comment above. I believe you still need to sign the CLA though.

akrherz:
Thanks @QuLogic , your suggestion is safer than my assumption that the effectively ignored last point associated with `Path.CLOSEPOLY` may not always equal the first point.  Hopefully properly signed the CLA as well.

ajdawson:
@marqh - please confirm when a CLA from @akrherz has been processed, thanks.

akrherz:
@ajdawson I thought signing the CLA involved adding my name to `docs/source/contributors.rst`, which I did in the commit above.  Otherwise, what needs to be done?

ajdawson:
@akrherz - You need to fill in the [CLA form](http://scitools.org.uk/documents/cla_form.pdf) and the Met Office will process it, shouldn't take long. If you are unfamiliar the the CLA concept see here: http://oss-watch.ac.uk/resources/cla

lbdreyer:
We have received the cla from @akrherz so this is now ready to go in (also see https://github.com/SciTools/scitools.org.uk/pull/155)

PR #886, opened 2017-05-23
Multiple small optimizations

QuLogic:
This optimizes a few small bits of repetitive work. Using a test of the following form (generating a bunch of maps on a grid like this is something I do often):
```python
import matplotlib as mpl
mpl.use('Agg')
import matplotlib.pylab as plt
import cartopy.crs as ccrs

transform = ccrs.PlateCarree(central_longitude=0)
fig, axs = plt.subplots(10, 10, figsize=(20, 20), subplot_kw=dict(projection=transform))
for ax in axs.flat:
    gridlines = ax.gridlines()

fig.savefig('tmp.png')
```
these changes reduce runtime from 8.32 s (average of 3 runs) to 3.73 s, approximately 2 times faster.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2016, Met Office
+# (C) British Crown Copyright 2011 - 2017, Met Office
 #
 # This file is part of cartopy.
 #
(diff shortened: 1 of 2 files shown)


QuLogic:
These changes help optimize two hotspots in `Gridliner._axes_domain` and `GeoAxesSubplot.cla`:
multiax

QuLogic:
Looks like I might have broken a test or two, though.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -270,6 +284,27 @@ def __init__(self, *args, **kwargs):
...
+
+        GeoAxes does not use the Axis for ticking or labelling, though
+        Matplotlib still uses it internally. So we need to keep it around, but
+        we disable the ticks using these small wrapper classes. We also do not
+        bother to register the Axis with any Spines.
+        """
+        self.xaxis = _TicklessXAxis(self)
+        self.yaxis = _TicklessYAxis(self)
+
+    def _gen_axes_spines(self, locations=None, offset=0.0, units='inches'):
```
dkillick: @QuLogic this method appears to be a duplicate of the one at [L1195-L1204](https://github.com/QuLogic/cartopy/blob/8a17a897f2dbb8d071803cb0806d15217cdbdde0/lib/cartopy/mpl/geoaxes.py#L1195-L1204)?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -219,6 +219,20 @@ def inverted(self):
                                         self.source_projection)
 
 
+# A small wrapper that does not create a "proto-Tick", which is expensive and
+# unused with GeoAxes.
+class _TicklessXAxis(matplotlib.axis.XAxis):
+    def reset_ticks(self):
+        del self.majorTicks[:]
```
dkillick: It always looks a little dramatic when `del` is used in module code! From a very small amount of testing it looks like you could instead set the major and minor ticks either to `None` or to `[]`, which would avoid the use of `del`.

dkillick:
@QuLogic I wonder if you could add a little testing for the new code you've introduced?

lib/cartopy/mpl/geoaxes.py
QuLogic: Yes, you are right, but it turns out I'll have to drop this patch anyway.

QuLogic:
So it turns out that ticks are still used on cylindrical projections and ticks also require spines, so I've had to drop the last commit. Alas, this means the final time is only about 6.20s with the speedup being 75% of original. And I'll have to go back to Matplotlib to try and figure out how to quicken the tick stuff.

@dkillick What I've changed here is already hit by existing tests, unless you want to check speed, but that seems tricky.

pelson:
The failing tests here are known failures, and so I'm 👍  on merging. @dkillick - are you happy to press merge?

dkillick:
@pelson happy to press merge! Thanks for this optimisation @QuLogic 👍

Issue #889, opened 2017-06-08
Fix failing build: pyshp error, http 503 error


lbdreyer:
The current build of the master branch is failing due to two errors:
* pyshp: There seems to have been a change in pyshp between versions 1.2.10 and 1.2.11 which changes the printout of a record from a shapefile (see [relevant comment](https://github.com/SciTools/cartopy/pull/780#issuecomment-307054268)). This is causing cartopy.tests.test_shapereader.TestRivers.test_record to fail
* http Error 503: During the build we are getting two `HTTPError: HTTP Error 503: Service Unavailable`

ajdawson:
>http Error 503: During the build we are getting two HTTPError: HTTP Error 503: Service Unavailable

Looks like #781. Previously we have removed the offending tests (#818; @marqh).

marqh:
there are issues with
http://oatile1.mqcdn.com/tiles/1.0.0/sat/0/0/0.jpg

I have made a small change to the code to see what is not resolving
https://github.com/marqh/cartopy/pull/2/commits/9299e830f040225c908f01b394ce86acffaaac2c
https://travis-ci.org/marqh/cartopy/jobs/243683742

the whole subdomain returns a HTTP 503
http://oatile1.mqcdn.com/

I agree with @ajdawson we can remove the test and this data source but my branch doesn't fix this and I am out of time (sorry)

ajdawson:
The integration of #758 has removed the remaining test failures (by setting them as expected fails). This is still a sticky plaster and not a fix for the API, but the tests run green now and this tracking issue can be closed.

PR #890, opened 2017-06-09
BUG: Fix incorrect ellipse boundary for Geostationary

dopplershift:
Need to calculate boundary out to where the satellite viewing angle is
tangent to the spheroid (not out to +/- 90 latitude and longitude). Also
should always use `a + h` in the denominator, since this needs to be total
distance of the satellite above the earth center point.

Also go ahead and use numpy math functions since values are converted to
numpy floats.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1619,8 +1619,8 @@ class _Satellite(Projection):
         a = np.float(self.globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS)
         b = np.float(self.globe.semiminor_axis or a)
         h = np.float(satellite_height)
-        max_x = h * math.atan(a / (a + h))
-        max_y = h * math.atan(b / (b + h))
+        max_x = h * np.arcsin(a / (a + h))
+        max_y = h * np.arcsin(b / (a + h))
 
         coords = _ellipse_boundary(max_x, max_y,
                                    false_easting, false_northing, 61)
--- a/lib/cartopy/tests/crs/test_geostationary.py
+++ b/lib/cartopy/tests/crs/test_geostationary.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2013 - 2016, Met Office
+# (C) British Crown Copyright 2013 - 2017, Met Office
 #
 # This file is part of cartopy.
 #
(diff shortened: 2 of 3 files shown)


dopplershift:
So without this fix, looking at GOES-16 data we see a ring of data symmetrically outside the boundary ellipse (it's currently dark in Antarctica, so no data to see in visible there)--as well as maps prematurely clipped. Here are examples before and after.

Incorrect:
unknown

Fixed:
unknown-1

dopplershift:
And if you really want to double check my geometry, here's the diagram that finally helped me figure out where the hell that %$@%@# ring came from:

img_0217

We want theta_2, not theta_1 as was previously done. Using theta_1 restricts the data to where that ray *first* intersects the spheroid (not out to +/- 90).

dopplershift:
You can also see this problem pretty clearly if you modify the `geostationary.py` example to plot in the image projection.

QuLogic:
This looks reasonable to me, but you'll need to update the date in the license headers.

Do you have any input from a different latitude? IIUC, that would make some difference with an ellipse, though finding the boundary is not as simple.

QuLogic:
According to [the table here](http://www.progonos.com/furuti/MapProj/Normal/ProjAz/projAz.html#GenVertPers), the GOES visible angle is 81°18', and assuming you have 20° gridlines, that seems about right.

dopplershift:
Copyrights updated.

pelson:
Thanks @dopplershift. Awesome! Appreciate the diagram too 😉 

So we are currently in the extremely uncomfortable situation where master is currently failing on TravisCI. I've reviewed the risk, and personally would be happy to merge this PR. @SciTools/cartopy-devs - any objections? 👎  if so and we will hold fire until we get a green tick.


@SciTools/cartopy-devs - in general, we're going to need to focus on getting the tests passing again ASAP so that we can have a bit of confidence wrt breakages. I've been out of the loop somewhat - is there an issue covering the subject I should be aware of?

lbdreyer:
>  is there an issue covering the subject I should be aware of?

I raised the issue #889 to capture the 2 erroring tests and the 1 failing test
I notice there are also 16 skipped tests due to SRTM login in not being supported.

QuLogic:
Works for me locally, so I don't think we need Travis. I don't think the failing tests use an ellipse boundary anyway.

PR #897, opened 2017-06-23
Support `sweep` parameter for Geostationary projection

dopplershift:
Proj.4 (added in 4.8) has a `sweep` parameter that designates which axis is fixed and which is scanning, which impacts the formula for the projection. (See [here](https://github.com/OSGeo/proj.4/wiki/Geostationary) for more info.)

I've added this parameter to the `Geostationary` and `_Satellite` projections. Owing to the Euro roots of the project, I made 'y' the default to match Meteosat. The upshot of this change is that GOES-16 data (making use of the `sweep_angle_axis` attribute) goes from:

screen shot 2017-06-22 at 9 02 46 pm

to:

screen shot 2017-06-22 at 9 03 33 pm

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1608,11 +1608,14 @@ class InterruptedGoodeHomolosine(Projection):
 class _Satellite(Projection):
     def __init__(self, projection, satellite_height=35785831,
                  central_longitude=0.0, central_latitude=0.0,
-                 false_easting=0, false_northing=0, globe=None):
+                 false_easting=0, false_northing=0, globe=None,
+                 sweep_axis=None):
         proj4_params = [('proj', projection), ('lon_0', central_longitude),
                         ('lat_0', central_latitude), ('h', satellite_height),
                         ('x_0', false_easting), ('y_0', false_northing),
                         ('units', 'm')]
+        if sweep_axis:
+            proj4_params.append(('sweep', sweep_axis))
         super(_Satellite, self).__init__(proj4_params, globe=globe)
 
         # TODO: Let the globe return the semimajor axis always.
@@ -1652,7 +1655,8 @@ class Geostationary(_Satellite):
 
     """
     def __init__(self, central_longitude=0.0, satellite_height=35785831,
-                 false_easting=0, false_northing=0, globe=None):
+                 false_easting=0, false_northing=0, globe=None,
+                 sweep_axis='y'):
         super(Geostationary, self).__init__(
             projection='geos',
             satellite_height=satellite_height,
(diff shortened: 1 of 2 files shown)


lib/cartopy/crs.py
```diff
@@ -1608,11 +1608,13 @@ def y_limits(self):
 class _Satellite(Projection):
     def __init__(self, projection, satellite_height=35785831,
                  central_longitude=0.0, central_latitude=0.0,
-                 false_easting=0, false_northing=0, globe=None):
+                 false_easting=0, false_northing=0, globe=None, sweep_axis=None):
```
ajdawson: Put the `sweep_axis` argument on a new line.

lib/cartopy/crs.py
```diff
@@ -1652,15 +1654,16 @@ class Geostationary(_Satellite):
 
     """
     def __init__(self, central_longitude=0.0, satellite_height=35785831,
-                 false_easting=0, false_northing=0, globe=None):
+                 false_easting=0, false_northing=0, globe=None, sweep_axis='y'):
```
ajdawson: Same here, new line for `sweep_axis`.

ajdawson: Should the default be `'y'` or `None`? I'm a bit confused about this I must admit. The document you linked to suggests the proj4 default is `'y'`, so is that what we have been getting anyway since proj4.8? In other words, is using `'y'` as the default backwards compatible?

The last paragraph of https://github.com/OSGeo/proj.4/wiki/Geostationary appears to contradict the previous 2, suggesting meteosat uses sweep `'x'` and GOES uses sweep `'y'`, which doesn't help my undertanding!

lib/cartopy/crs.py
dopplershift: Oops.

lib/cartopy/crs.py
dopplershift: Yeah, in my logic before, I missed the fact that the best reason to use a default of `'y'` is that it's identical to what was used before.

On my read of that paragraph, it seems consistent. Most importantly:
1. "Thus, the scanning geometry of the Meteosat series satellite should take sweep as x, and GOES should take sweep as y."
2. The files I have for GOES-16 specify `sweep_angle_axis` as `'x'`, and using this value makes the image line up perfectly with maps.

dopplershift:
Fixed the bad lines. To be clear, this change is entirely backwards compatible (only test changes were to expect the additional Proj.4 parameter)

This should be good now I think.

ajdawson:
Excellent, thanks @dopplershift.

PR #904, opened 2017-07-02
correct unit error in geoaxes.tissot() (m->km)

skjaeve:
The tissot() docstring says the radius input is in km, but the underlying Geodesic.circle() function takes it to be meters. This patch adds a conversion from meters to kilometers.


lbdreyer:
We have now received the signed copy of cla from @skjaeve

QuLogic:
Please update [contributors list](https://github.com/SciTools/scitools.org.uk/blob/gh-pages/contributors.json).

lbdreyer:
> Please update contributors list.

The PR to do this is up: https://github.com/SciTools/scitools.org.uk/pull/156

PR #916, opened 2017-08-07
ENH: Set default zorder for LAND and OCEAN

dopplershift:
This should help keep them behind most plot elements by default. This
fixes a problem where wind barbs end up hidden by these features.

### DIFF
--- a/lib/cartopy/feature.py
+++ b/lib/cartopy/feature.py
@@ -377,13 +377,13 @@ LAKES = NaturalEarthFeature('physical', 'lakes', '110m',
 
 LAND = NaturalEarthFeature('physical', 'land', '110m',
                            edgecolor='face',
-                           facecolor=COLORS['land'])
+                           facecolor=COLORS['land'], zorder=-1)
 """Small scale (1:110m) land polygons, including major islands."""
 
 
 OCEAN = NaturalEarthFeature('physical', 'ocean', '110m',
                             edgecolor='face',
-                            facecolor=COLORS['water'])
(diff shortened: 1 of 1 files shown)


QuLogic:
Re-trigger Travis.

dopplershift:
Travis failure is complaining about libgfortran. Can someone try restarting it?

QuLogic:
Didn't work; did conda-forge drop support for 3.4?

dopplershift:
That sounds familiar...

ocefpaf:
> That sounds familiar...though the packages that were there should still be there.

And that is exactly the problem. The remaining packages never for updated to `defaults` latest `libgfortran` and they are expecting an old one that probably was not pinned down at the time they were built.

I can re-build the stack for cartopy on Python 3.4 if that is really important here. If not I recommend to just drop Python 3.4 from the tests.

ajdawson:
Yeah we should probably just be testing 3.5 and 3.6 and drop 3.4 as @ocefpaf says

dopplershift:
I can change 3.4 -> 3.5 and 3.5 ("latest everything") to 3.6 in `.travis.yml`. Does that need to be a separate PR?

ajdawson:
Well, probably. It may be useful to establish that changing the test setup doesn't break anything independently from the changes here.

dopplershift:
Done in #926.

ajdawson:
Can you rebase this please @dopplershift

dopplershift:
Done.

ajdawson:
Thanks @dopplershift

PR #926, opened 2017-08-16
MNT: Drop Python 3.4 from test matrix

dopplershift:
Conda-forge packages for 3.4 aren't updated any more. Move 3.5 to where
3.4 was used, use 3.6 for "latest everything".


dopplershift:
So replacing 3.4 in the build matrix with 3.5 and conda packages means:
- NumPy 1.9
- Matplotlib 1.4.3
- SciPy 0.16.0

Is it ok just to update these?

dopplershift:
And no clue why running on 3.6 breaks 2 tests....

QuLogic:
I personally don't think we should bother with 1.3.1, but I assume this is something that the Met Office has previously had need for.

ajdawson:
1.3.1 seems a little bit too conservative, 1.4.3 is 2 major versions behind the latest and that should satisfy most needs.

QuLogic:
Any idea what's up with the `libgfortran` install?

dopplershift:
Need to install libgfortran 1.0 package--which unfortunately isn't an explicit dependence of scipy 0.16 in defaults.

dopplershift:
Well, hey! That last commit at least got Python 3.5 with old packages to pass...

QuLogic:
I think we kind of skipped testing with 1.4.3...

dopplershift:
Down to 1 failure on latest everything for both 2 and 3. 

Baseline:
web_tiles

Result for me locally (which also fails with same RMS):
result-web_tiles

Diff:
result-web_tiles-failed-diff

Weird.

QuLogic:
I think these are remote images that might just need to be updated.

dopplershift:
I was just looking at that. We could also bump the tolerance for Matplotlib 2.x from 2 to 2.9.

.travis.yml
```diff
@@ -26,9 +30,9 @@ install:
   # Install miniconda
   # -----------------
   - if [[ "$PYTHON_VERSION" == 2* ]]; then
-      wget http://repo.continuum.io/miniconda/Miniconda-3.7.0-Linux-x86_64.sh -O miniconda.sh;
+      wget http://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
     else
-      wget http://repo.continuum.io/miniconda/Miniconda3-3.7.0-Linux-x86_64.sh -O miniconda.sh;
+      wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
```
QuLogic: Can you make these `https` too?

lib/cartopy/tests/test_polygon.py
```diff
@@ -270,10 +270,13 @@ def test_plate_carree_partial_wrap(self):
         # Check the structure
         assert len(multi_polygon) == 2
         # Check the rough shape
-        polygon = multi_polygon[0]
-        self._assert_bounds(polygon.bounds, 170, 0, 180, 10)
-        polygon = multi_polygon[1]
-        self._assert_bounds(polygon.bounds, -180, 0, -170, 10)
+        poly1, poly2 = multi_polygon
```
QuLogic: This was sooo annoying.

.travis.yml
dopplershift: Done.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -31,10 +31,10 @@
 from cartopy.tests.mpl import MPL_VERSION, ImageTesting
 
 
-_ROB_TOL = 0.5 if ccrs.PROJ4_VERSION < (4, 9) else 0.1
+_ROB_TOL = 0.5 if ccrs.PROJ4_VERSION < (4, 9) else 0.111
 if MPL_VERSION >= '2':
     _STREAMPLOT_IMAGE = 'streamplot'
-elif MPL_VERSION >= '1.5':
+elif MPL_VERSION >= '1.4.3':
     _STREAMPLOT_IMAGE = 'streamplot_1.5'
```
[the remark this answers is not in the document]
dopplershift: Done here and in the other file named for 1.5 below.

djkirkham:
I'm getting failures when I run the tests locally: three related to ShapeReader, and a couple of image test failure when running with the latest available Matplotlib (2.1.0). I'll restart the Travis tests here.

PR #931, opened 2017-09-12
Prevent memory leak in CRS objects

djkirkham:
Cartopy CRS objects don't free the proj4 object they own. This PR fixes that.


QuLogic:
Ping @pelson ?

PR #953, opened 2017-11-15
Tiny fix for errors not being caught in set_extent

corinnebosley:
Sometimes when plotting extents are very close to the poles, they are not being plotted due to some numerical issues with the geometries.  A fix had been implemented for this, but for some reason one of the conditions to try the workaround was not being met (that is, when crs was equal to self.projection, it was still being evaluated as False) so the workaround was being skipped.

This alteration should catch those conditions and apply the workaround.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -641,7 +641,7 @@ class GeoAxes(matplotlib.axes.Axes):
         # extents are obviously the same as the projection domain.
         try_workaround = ((crs is None and
                            isinstance(self.projection, ccrs.PlateCarree)) or
-                          isinstance(crs, ccrs.PlateCarree))
+                          isinstance(crs, self.projection.__class__))
         if try_workaround:
             boundary = self.projection.boundary
             if boundary.equals(domain_in_crs):
@@ -1115,9 +1115,9 @@ class GeoAxes(matplotlib.axes.Axes):
             # clip the image. This does not work as the patch moves with mouse
             # movement, but the clip path doesn't
             # This could definitely be fixed in matplotlib
-        #        if result.get_clip_path() in [None, self.patch]:
-        #            # image does not already have clipping set, clip to axes patch
-        #            result.set_clip_path(self.outline_patch)
+            # if result.get_clip_path() in [None, self.patch]:
+            # image does not already have clipping set, clip to axes patch
+            #  result.set_clip_path(self.outline_patch)
         return result
 
     def gridlines(self, crs=None, draw_labels=False, xlocs=None,
@@ -1443,7 +1443,7 @@ class GeoAxes(matplotlib.axes.Axes):
         collection.set_alpha(alpha)
         collection.set_array(C)
         if norm is not None:
-            assert (isinstance(norm, mcolors.Normalize))
+            assert isinstance(norm, mcolors.Normalize)
         collection.set_cmap(cmap)
         collection.set_norm(norm)
         collection.set_clim(vmin, vmax)
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -51,10 +51,8 @@ import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 from cartopy.vector_transform import vector_scalar_to_grid
 
-
 assert mpl.__version__ >= '1.3', ('Cartopy is only supported with '
                                   'Matplotlib 1.3 or greater.')
-
 
 _PATH_TRANSFORM_CACHE = weakref.WeakKeyDictionary()
 """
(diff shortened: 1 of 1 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -642,7 +641,7 @@ def set_extent(self, extents, crs=None):
         # extents are obviously the same as the projection domain.
         try_workaround = ((crs is None and
                            isinstance(self.projection, ccrs.PlateCarree)) or
-                          crs == self.projection)
+                          isinstance(crs, ccrs.PlateCarree))
```
dkillick: What if `crs` _isn't_ PlateCarree?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1112,12 +1112,12 @@ def imshow(self, img, *args, **kwargs):
...
-        # clip the image. This does not work as the patch moves with mouse
-        # movement, but the clip path doesn't
-        # This could definitely be fixed in matplotlib
-#        if result.get_clip_path() in [None, self.patch]:
-#            # image does not already have clipping set, clip to axes patch
-#            result.set_clip_path(self.outline_patch)
+            # clip the image. This does not work as the patch moves with mouse
+            # movement, but the clip path doesn't
+            # This could definitely be fixed in matplotlib
+        #        if result.get_clip_path() in [None, self.patch]:
```
dkillick: I have to say, this rendering of the diff suggests that these comment characters still aren't indented correctly...

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1443,7 +1443,7 @@ def _pcolormesh_patched(self, *args, **kwargs):
         collection.set_alpha(alpha)
         collection.set_array(C)
         if norm is not None:
-            assert(isinstance(norm, mcolors.Normalize))
+            assert (isinstance(norm, mcolors.Normalize))
```
dkillick: I don't think this line needs the outer brackets at all. There's no line continuation or explicit operation order going on here...

corinnebosley: I actually didn't touch that... I'm confused...

lib/cartopy/mpl/geoaxes.py
corinnebosley: I also didn't touch that... what's happening?

lib/cartopy/mpl/geoaxes.py
corinnebosley: Yeah good point.  Didn't think that through, sorry.  I'll try again.

lib/cartopy/mpl/geoaxes.py
dkillick: Has your editor gained sentience??

lib/cartopy/mpl/geoaxes.py
ajdawson: This doesn't check the same thing does it? It was supposed to be checking if the given crs is the same as the projection (i.e. native coordinates, don't care what the projection actually is). Now it is just checking if the crs is `PlateCarree`, which seems wouldn't be useful unless we also knew the projection was `PlateCarree`. Can you explain the motivation here?

dkillick: @ajdawson the motivation is that according to @corinnebosley the original is not working...

ajdawson: That doesn't help at all. What I want to know is why are we changing the logic (i.e. the question we are asking) instead of the implementation of the logic?

It seems like the question we want the answer to is "is the specified crs the same as our projection?". Instead this change checks if the crs is `PlateCarree`, but why? How does this help determine if the workaround is needed? What if the specified crs matches the projection but isn't `PlateCarree`?

corinnebosley: Yes thank you @ajdawson, Pete already pointed out that observation.  It was an oversight on my part which I am now fixing.  I do still want to test that the crs is the same as the projection no matter what they are.

ajdawson: OK, timing difference between starting comment and posting, PR had 0 comments when I began.

lib/cartopy/mpl/geoaxes.py
corinnebosley: I think it might have done.  It's certainly started being a bit stroppy and irrational lately...

corinnebosley:
@dkillick This should be a bit more general as a check that the crs is the same as the projection.

I am having some major problems reproducing the failure mode though, so I can't get a new test to fail with the old behaviour (which would pass with the replacement behaviour).  Not sure what to do next...

corinnebosley:
I discovered at 5pmish yesterday that the only way I could reproduce this problem was if the call to set_extent came from the ADAQ-Python code that I am working with, so I looked at that instead and found that some floats needed to be turned into ints, and that fixed it.

It still makes me a little nervous that the check I was trying to fix didn't work, when everything looked the same and the comparison should have returned True, but given the circumstances and the fact that I can't make a test to reproduce the problem, I'm just going to close this and consider it a teeny tiny corner case which will not happen again.

Issue #899, opened 2017-06-23
Improvements to documentation / examples


choldgraf:
I'm helping somebody transcribe a bunch of R code into python for map plotting, and I have found cartopy to be fairly powerful, but really, really hard to figure out the API. There seem to be a lot of things exposed to users that don't have an example showing them off (e.g., I see that @dopplershift just added a user-friendly way to plot state boundaries).

I think it'd be really useful if there were a collection of examples / "intro to cartopy" tutorial that showed off the many things that this package can do!

I can potentially put some time into this as I figure out how this package works, though it'll be a little bit of time because I'm still figuring things out. Just putting this here for now to see if this is something that folks are already planning to implement.

choldgraf:
Note - some of this could be improved by swapping out the current gallery build with something like sphinx-gallery and displaying titles of examples so they're more informative from that page.

choldgraf:
per #900 a useful example would be creating a geoaxis with `plt.subplots`:

```
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
proj = ccrs.Geostationary()
fig, axes = plt.subplots(2, 2, subplot_kw=dict(projection=proj)))
```

(just leaving this here for future reference)

lwasser:
i'd like to second this request! i'm working on these lessons with @choldgraf and he told me about cartopy. i'd love to see some vignettes and could also potentially contribute once i have a grasp of how it works - i'd potentially like to teach with it. thanks!!

lgolston:
This was addressed by #969

PR #954, opened 2017-11-20
Example added to display all 60 zones of UTM projection side-by-side

kaedonkers:
The projection list in Cartopy documentation displays only one UTM zone instead of all sixty:
proj_list_utm_crop

This is an example demonstrating how to display all sixty zones in one figure:
utm_all_zones

A followup to this will be to include a similar figure in the projection list.


pelson:
Awesome addition @kaedonkers! I don't think I've ever actually seen them all side-by-side before!
FYI We have a few issues with continuous integration that I'm going to try to get on top of before merging this.

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+    # Create a list of integers from 1 - 60
+    zones = range(1, 61)
+
+    # Create a figure and array of 60 subplot Axes objects
+    fig, axarr = plt.subplots(nrows=1,
+                              ncols=len(zones),
+                              figsize=(18, 6))
+
+    # Loop through each zone in the list
+    for zone in zones:
```
pelson: I'd have probably ```for zone, ax in zip(zones, axarr):```, but your approach works too.

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def main():
+    # Create a list of integers from 1 - 60
+    zones = range(1, 61)
+
+    # Create a figure and array of 60 subplot Axes objects
+    fig, axarr = plt.subplots(nrows=1,
+                              ncols=len(zones),
+                              figsize=(18, 6))
```
pelson: Might be worth taking the approach in https://github.com/SciTools/cartopy/issues/899#issuecomment-310792210

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Loop through each zone in the list
+    for zone in zones:
+
+        # Change the Axes object to a GeoAxes object with a specific UTM zone projection
+        axarr[zone - 1] = plt.subplot(1, len(zones), zone,
+                                      projection=ccrs.UTM(zone=zone, southern_hemisphere=True))
+
+        # Add coastlines, gridlines and zone number for the subplot
+        axarr[zone - 1].coastlines(resolution='110m')
```
pelson: Did you take a look at adding land? How did it look?

lib/cartopy/examples/utm_all_zones.py
QuLogic: But you need a different constructor for every subplot, due to different `zone`s.

lib/cartopy/examples/utm_all_zones.py
QuLogic: `ax` is getting replaced anyway.

lib/cartopy/examples/utm_all_zones.py
QuLogic: Since you aren't using the existing subplots, it seems a bit better to create just the figure here. `fig = plt.figure(figsize=(18, 6))`

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+    # Create a figure and array of 60 subplot Axes objects
+    fig, axarr = plt.subplots(nrows=1,
+                              ncols=len(zones),
+                              figsize=(18, 6))
+
+    # Loop through each zone in the list
+    for zone in zones:
+
+        # Change the Axes object to a GeoAxes object with a specific UTM zone projection
+        axarr[zone - 1] = plt.subplot(1, len(zones), zone,
```
QuLogic: And then use `fig.add_subplot` here instead.

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+                                      projection=ccrs.UTM(zone=zone, southern_hemisphere=True))
+
+        # Add coastlines, gridlines and zone number for the subplot
+        axarr[zone - 1].coastlines(resolution='110m')
+        axarr[zone - 1].gridlines()
+        axarr[zone - 1].set_title(zone)
+
+    # Add a supertitle for the figure
+    plt.suptitle("UTM Projection - Zones")
```
QuLogic: `fig.suptitle`

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+Finally we add a supertitle and display the figure.
+"""
+__tags__ = ['Lines and polygons']
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+
+
+def main():
+    # Create a list of integers from 1 - 60
+    zones = range(1, 61)
```
QuLogic: This is not a list on Python 3; I'd just inline it into the `for`-loop.

lib/cartopy/examples/utm_all_zones.py
kaedonkers: There are anomalies when plotting coastlines on in a number of the zones, which is made more obvious when land is added:
utm_all_zones_land

The anomalous zones are [2, 5, 8, 32, 35, 36, 38, 41, 52]:
utm_anomalies_land

So I chose to keep land out of the example until this has been fixed.

kaedonkers: Errors are returned when using `cartopy.feature.OCEAN` and `ax.stock_img()`, to the extent that nothing is displayed.

lib/cartopy/examples/utm_all_zones.py
kaedonkers: 1. In Python 3 `range()` is not a list but it will work in the `for` loop just the same, right?

2. I separated the list of zones from the `for` loop to make it easy for someone to try out their own list of zones.
e.g. List of anomalous zones from the [comment](https://github.com/SciTools/cartopy/pull/954#discussion_r152208271) above

lib/cartopy/examples/utm_all_zones.py
kaedonkers: Good suggestion, thanks

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Create a figure and array of 60 subplot Axes objects
+    fig, axarr = plt.subplots(nrows=1,
+                              ncols=len(zones),
+                              figsize=(18, 6))
+
+    # Loop through each zone in the list
+    for zone in zones:
+
+        # Change the Axes object to a GeoAxes object with a specific UTM zone projection
```
bjlittle: @kaedonkers Could you make this line [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) compliant, by splitting it and ensuring that the line length is <80 characters.

lib/cartopy/examples/utm_all_zones.py
```diff
@@ -0,0 +1,47 @@
...
+        axarr[zone - 1] = plt.subplot(1, len(zones), zone,
+                                      projection=ccrs.UTM(zone=zone, southern_hemisphere=True))
```
bjlittle: @kaedonkers Same here, the pep8 tests are failing as this line is too long, thanks!

QuLogic:
There's still one PEP8 failure though:
```
.../cartopy/examples/utm_all_zones.py:28:80: E501 line too long (80 > 79 characters)
```

QuLogic:
Can you rebase this against current `master`?

PR #970, opened 2017-11-28
Configure + enable stickler testing.

pp-mo:


lib/cartopy/crs.py
```diff
@@ -746,7 +747,6 @@ def quick_vertices_transform(self, vertices, src_crs):
                 mod = np.diff(src_crs.x_limits)[0]
                 bboxes, proj_offset = self._bbox_and_offset(src_crs)
                 x_lim = xs.min(), xs.max()
-                y_lim = ys.min(), ys.max()
                 for poly in bboxes:
```
corinnebosley: @pp-mo What's happened to y_lim?

lib/cartopy/crs.py
```diff
@@ -1184,7 +1184,6 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
                                                         globe=globe)
 
         a = np.float(self.globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS)
-        b = np.float(self.globe.semiminor_axis or WGS84_SEMIMINOR_AXIS)
         lon, lat = central_longitude + 180, - central_latitude + 0.01
```
corinnebosley: @pp-mo And also what's happened to b (the semiminor axis)?

lib/cartopy/__init__.py
```diff
@@ -108,4 +108,4 @@
 # Commonly used sub-modules. Imported here to provide end-user
 # convenience.
 import cartopy.crs
-import cartopy.feature
+import cartopy.feature  # noqa: F401  (imported but not used)
```
corinnebosley: Does this comment stop Stickler complaining?

lib/cartopy/sphinxext/gallery.py
```diff
@@ -202,8 +201,6 @@ def gallery_code(examples_mod_name):
             # XXX The path is currently determined out of process by
             # the plot directive. It would be nice to figure out the
             # naming scheme to handle multiple plots in a single example.
-            img_path = 'examples/{}_00_00.png'.format(
-                mod_name.split('.')[-1])
             thumb_path = 'examples/{}_00_00.thumb.png'.format(
```
corinnebosley: Do we not need this any more (img_path)?

lib/cartopy/crs.py
pp-mo: 'y_lim' was not used, so flake8 complained about it, so I got rid of it.

There's a lot of that going on in this PR !

lib/cartopy/crs.py
pp-mo: 'b' also not used anywhere.
As it was identical, I suspect this one was a mistake from the first.

QuLogic: Identical to what?

lib/cartopy/mpl/gridliner.py
```diff
@@ -335,9 +336,10 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
         if self.xlines:
             lines = []
             for x in x_gridline_points:
-                l = list(zip(np.zeros(n_steps) + x,
-                         np.linspace(min(y_ticks), max(y_ticks), n_steps)))
-                lines.append(l)
+                ticks = list(zip(
```
QuLogic: I could've sworn I'd written some optimization for this unfortunate type conversion; maybe I never made a PR out of it, but I can see if I can dig it up.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -249,7 +249,7 @@ def test_pcolormesh_global_with_wrap1():
 @ImageTesting(['pcolormesh_global_wrap2'])
 def test_pcolormesh_global_with_wrap2():
     # make up some realistic data with bounds (such as data from the UM)
-    nx, ny = 36, 18
+    nx = 36
     xbnds, xstep = np.linspace(0, 360, nx - 1, retstep=True, endpoint=True)
     ybnds, ystep = np.linspace(-90, 90, nx - 1, retstep=True, endpoint=True)
```
QuLogic: Seems better to make this `ny` (but that would require changing `ny` since it was not used before.)

lib/cartopy/__init__.py
pp-mo: Yes, it's a specific indication to the flake8 tester, actually the "pyflakes" component

If anyone is interested ...
It's really not very clearly explained, but after a bit of digging http://flake8.pycqa.org/en/latest/user/error-codes.html
So, "F" numbers are errors from pyflakes, and "E" numbers are from pycodestyle (pep8-successor).

lib/cartopy/mpl/gridliner.py
pp-mo: I don't see a need to connect this with the issue at hand  ?
We rather want to get this onto master so we can get on right now, since stickler is now complaining about various open PRs.

lib/cartopy/tests/mpl/test_mpl_integration.py
dopplershift: Really probably just `n`, but I don’t think it’s really worth worrying about.

pp-mo: Ok, yes I hadn't spotted it -- it looks like that was the error.
Fixed..

pp-mo:
Hi @QuLogic if we have your interest, would you consider merging this ?
If not, fine, but if you just sign off your look at it.
Then I think I'm happy to merge myself now, as I think it's had enough other eyes now.

lib/cartopy/crs.py
pp-mo: Identical to 'a'.
If it was going to be useful, I'd expect it to be different in some way, but it's all just guesswork.

QuLogic: It's not identical to 'a', though. But I'm not sure it needs to be used anyway.

lib/cartopy/mpl/gridliner.py
QuLogic: No, just a reminder to myself.

lib/cartopy/crs.py
pp-mo: Oh I see, it is not after all : more care required !

PR #974, opened 2017-11-30
Fix for 'linestring being readded' bug

djkirkham:
When attaching lines to the boundary, the points where the lines meet the boundary, plus the boundary points, are put into a list and sorted by their distance along the boundary. If two lines happen to have an end-point at the same location then there is a risk that the end-points of the two lines will be interleaved, which causes the algorithm to fail later on with the 'linestring being re-added' error. This PR adds extra elements to the key used for the sort to prevent this happening.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -429,6 +429,11 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                     ax.text(coords[-1, 0], coords[-1, 1],
                             '{}.'.format(thing.data[0]))
 
+        def filter_last(t):
+            return t.kind or t.data[1] == 'first'
+
+        edge_things = list(filter(filter_last, edge_things))
+
         processed_ls = []
         while remaining_ls:
             # Rename line_string to current_ls
@@ -441,16 +446,6 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                 print()
                 print('Processing: %s, %s' % (i, current_ls))
 
-            # We only want to consider boundary-points, the starts-and-ends of
-            # all other line-strings, or the start-point of the current
-            # line-string.
-            def filter_fn(t):
-                return (t.kind or
-                        t.data[0] != i or
-                        t.data[1] != 'last')
-
-            edge_things = list(filter(filter_fn, edge_things))
-
             added_linestring = set()
             while True:
                 # Find out how far around this linestring's last
@@ -459,7 +454,7 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                 d_last = boundary_distance(current_ls.coords[-1])
                 if debug:
                     print('   d_last: {!r}'.format(d_last))
-                next_thing = _find_first_gt(edge_things, d_last)
+                next_thing = _find_first_ge(edge_things, d_last)
                 # Remove this boundary point from the edge.
                 edge_things.remove(next_thing)
                 if debug:
@@ -473,7 +468,7 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                                        [(boundary_point.x, boundary_point.y)])
                     current_ls = sgeom.LineString(combined_coords)
 
-                elif next_thing.data[0] == i and next_thing.data[1] == 'first':
+                elif next_thing.data[0] == i:
                     # We've gone all the way around and are now back at the
                     # first boundary thing.
                     if debug:
@@ -493,8 +488,6 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                         remaining_ls.pop(j)
                     coords_to_append = list(line_to_append.coords)
 
-                    if next_thing.data[1] == 'last':
-                        coords_to_append = coords_to_append[::-1]
                     # Build up the linestring.
                     current_ls = sgeom.LineString((list(current_ls.coords) +
                                                    coords_to_append))
@@ -1937,9 +1930,9 @@ class _BoundaryPoint(object):
                                                self.data)
 
 
-def _find_first_gt(a, x):
+def _find_first_ge(a, x):
     for v in a:
-        if v.distance > x:
+        if v.distance >= x:
             return v
     # We've gone all the way around, so pick the first point again.
     return a[0]
--- a/lib/cartopy/tests/test_polygon.py
+++ b/lib/cartopy/tests/test_polygon.py
@@ -193,6 +193,27 @@ class TestMisc(object):
         assert 81330 < area < 81340, \
             'Got area {}, expecting ~81336'.format(area)
 
+    def test_same_points_on_boundary_1(self):
+        source = ccrs.PlateCarree()
+        target = ccrs.PlateCarree(central_longitude=180)
+
+        geom = sgeom.Polygon([(-20, -20), (20, -20), (20, 20), (-20, 20)],
+                             [[(-10, 0), (0, 20), (10, 0), (0, -20)]])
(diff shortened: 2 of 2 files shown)


dopplershift:
Can you add a test case that demonstrates the problem?

djkirkham:
I've changed the fix to a cleaner implementation. The basic idea now is to only consider 'first' points when looking for new points to add to the current linestring. The justification for this is that, assuming the exterior linestring of a polygon has the opposite orientation (clockwise or counter-clockwise) to all interior linestrings, then when traversing the boundary looking for the end point of the next line to add to the current linestring, we _should_ only encounter 'first' points; if we encounter a 'last' point then it can only be that it is at the same location as a 'first' point. As to whether the above assumption is justified, the algorithm already implicitly assumes this: when beginning to process a new linestring in the first iteration of [the loop at line 455 ](https://github.com/SciTools/cartopy/blob/a46498c1a9f72b3150501d79d19647e2ee7c2c65/lib/cartopy/crs.py#L455) it chooses to process the line from 'first' to 'last'. This is only valid if the above assumption is true.

It turns out that the root cause of the bug doesn't always result in an error. It can simply result in the polygon encompassing the whole map. I've added two tests which demonstrates the two cases.

The test `test_same_points_on_boundary_2` demonstrates the original issue (the error). When running the content of the test with `debug = True` at [line 340 in crs.py](https://github.com/SciTools/cartopy/blob/a46498c1a9f72b3150501d79d19647e2ee7c2c65/lib/cartopy/crs.py#L340) the following output is produced in the original code:

```
Edge things
    _BoundaryPoint(0.0, True, POINT (-180 -90))
    _BoundaryPoint(360.0, True, POINT (180 -90))
    _BoundaryPoint(430.0, False, (1, 'last', (179.99996185302734, -20.0)))
    _BoundaryPoint(430.00000953674316, False, (5, 'first', (179.9999904632568, -19.999990463256836)))
    _BoundaryPoint(440.0, True, POINT (180 -10))
    _BoundaryPoint(449.99999046325684, False, (5, 'last', (179.9999904632568, -9.5367431640625e-06)))
    _BoundaryPoint(450.00000953674316, False, (2, 'first', (179.9999904632568, 9.5367431640625e-06)))
    _BoundaryPoint(460.0, True, POINT (180 10))
    _BoundaryPoint(469.99999046325684, False, (2, 'last', (179.9999904632568, 19.999990463256836)))
    _BoundaryPoint(470.0, False, (1, 'first', (179.99996185302734, 20.0)))
    _BoundaryPoint(540.0, True, POINT (180 90))
    _BoundaryPoint(900.0, True, POINT (-180 90))
    _BoundaryPoint(970.0, False, (0, 'last', (-180.0, 20.0)))
    _BoundaryPoint(970.0, False, (3, 'first', (-180.0, 20.0)))
    _BoundaryPoint(980.0, True, POINT (-180 10))
    _BoundaryPoint(990.0, False, (3, 'last', (-180.0, 0.0)))
    _BoundaryPoint(990.0, False, (4, 'first', (-180.0, 0.0)))
    _BoundaryPoint(1010.0, False, (0, 'first', (-179.99999999992724, -20.0)))
    _BoundaryPoint(1010.0, False, (4, 'last', (-180.0, -20.0)))
+
Processing: 0, LINESTRING (-179.9999999999272 -20, -160 -20, -160 20, -180 20)
   d_last: 970.0
   next_thing: _BoundaryPoint(980.0, True, POINT (-180 10))
   adding boundary point
   d_last: 980.0
   next_thing: _BoundaryPoint(990.0, False, (3, 'last', (-180.0, 0.0)))
   adding line
   d_last: 970.0
   next_thing: _BoundaryPoint(990.0, False, (4, 'first', (-180.0, 0.0)))
   adding line
   d_last: 1010.0
   next_thing: _BoundaryPoint(0.0, True, POINT (-180 -90))
   adding boundary point
   d_last: 0.0
   next_thing: _BoundaryPoint(360.0, True, POINT (180 -90))
   adding boundary point
   d_last: 360.0
   next_thing: _BoundaryPoint(430.0, False, (1, 'last', (179.99996185302734, -20.0)))
   adding line
   d_last: 470.0
   next_thing: _BoundaryPoint(540.0, True, POINT (180 90))
   adding boundary point
   d_last: 540.0
   next_thing: _BoundaryPoint(900.0, True, POINT (-180 90))
   adding boundary point
   d_last: 900.0
   next_thing: _BoundaryPoint(970.0, False, (3, 'first', (-180.0, 20.0)))
   adding line
  File "<stdin>", line 1, in <module>
  File ".../lib/cartopy/crs.py", line 176, in project_geometry
  File ".../lib/cartopy/crs.py", line 327, in _project_polygon
    rings.extend(self._attach_lines_to_boundary(multi_lines, is_ccw))
  File ".../lib/cartopy/crs.py", line 509, in _attach_lines_to_boundary
    raise RuntimeError('Unidentified problem with '
RuntimeError: Unidentified problem with geometry, linestring being re-added. Please raise an issue.
```

Notice that the end points of linestrings enumerated 0 and 4 are interleaved. Because of this interleaving, when processing linestring 0, the algorithm 'jumps over' point (0, 'first') via linestring 4. It then goes all around the boundary and re-encounters linestring 3, which results in the error.

pelson:
P.S. I tested this with the code @djkirkham provided:

```
import cartopy.crs as ccrs

source = ccrs.PlateCarree()
target = ccrs.PlateCarree(central_longitude=180)

geom = sgeom.Polygon([(-20, -20), (20, -20), (20, 20), (-20, 20)],
                     [[(-10, 0), (0, 20), (10, 0), (0, -20)]])
```

Before: Error (linestring being re-added)
After: 
readded

lib/cartopy/crs.py
```diff
@@ -429,6 +429,11 @@ def boundary_distance(xy):
                     ax.text(coords[-1, 0], coords[-1, 1],
                             '{}.'.format(thing.data[0]))
 
+        def filter_last(t):
+            return t.kind or t.data[1] == 'first'
```
pelson: I do wonder whether this should be ``!= 'last'``.

pp-mo:
Actually, I wasn't quite done with this...
Apart from a desire for better testing, I had a suspicion that the whole of [crs.py line 397](https://github.com/djkirkham/cartopy/blob/7a75ca96c7f73e79f221adeaf151976c1255bc91/lib/cartopy/crs.py#L397) onward is now obsolete, 
i.e. `not edge_thing.kind and not prev_thing.kind and edge_thing.data[0] == prev_thing.data[0]` can no longer happen at all...
To be continued...

pelson:
> Actually, I wasn't quite done with this...

Please feel free to use the reviewing / assignment functionality.

I actually believe the testing *is* sufficient, though you may be correct about some of this making subsequent code obsolete. Happy to review a PR from yourself / @djkirkham to clear that bit up.

pp-mo:
> some of this making subsequent code obsolete
No its okay, I checked it through again and I was wrong !

> the testing is sufficient, 
My concern is that this code has an especially high potential for complicated special cases.
For instance, polygons may have inner holes that touch one another or the outer ring at a point, and which potentially can also intersect or contact the projection boundary.  Also, projections can turn a polygon inside out around an inner point if it surrounds a pole.
As such "awkward" cases are not explored in the tests, we don't know which ones were handled properly either before or after these changes.  I would like to have more confidence when refactoring code.
What we've done here is to focus on one less-than-usual case that was definitely wrong, add a testcase and fix that.  We don't really have much confidence that it didn't introduce any new problems elsewhere, at least for other less-than-common cases.

PR #993, opened 2018-01-04
Egs 2 gallery

pp-mo:
Addresses #991.
Does the same thing for inline code in the 'gridlines' documentation.

### DIFF
--- a/docs/source/matplotlib/advanced_plotting.rst
+++ b/docs/source/matplotlib/advanced_plotting.rst
@@ -14,117 +14,45 @@ of different approaches to data loading.
 Contour plots
 -------------
 
+A very common way of displaying scalar data over a map is to draw contour
+lines, or filled contours with a colormap.  This will display continuous
+coloured regions, smoothing over any gridcell edges even if they are large.
 
-.. plot::
-    :include-source:
+See the :ref:`contour plot example <examples-contour-plot>`
 
-    import os
-    import matplotlib.pyplot as plt
-    from netCDF4 import Dataset as netcdf_dataset
-    import numpy as np
-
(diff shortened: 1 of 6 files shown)


QuLogic:
What is egs? I'll also note that you don't need to create anchors manually; sphinx-gallery adds one for every example, see [the getting started page](https://sphinx-gallery.github.io/getting_started.html#know-your-gallery-files)

pp-mo:
> What is egs? 

Examples, as, in "e.g.".
I don't think I used that anywhere in the codebase ?


> you don't need to create anchors manually

Thanks, I'll check that out.  As usual, I was boilerplating (?verb?) from something I found elsewhere.

pp-mo:
> you don't need to create anchors manually

A much more serious matter is the test failures, due to the example code intoducing dependencies on netCDF4 and iris (!arrgh!).  We surely don't want to add Iris to the test dependencies, so I may need to revert that

.. or work out how to stop pytest loading the examples? 
I *think* this occurs because "pytest --doctest-modules" attempts to import everything it can find.
But I'm not too clear why, as I think we don't actually ***have*** any doctests.
**Anyone know about that  ?**

pp-mo:
> test failures ... we don't actually have any doctests.

From offline discussion with @pelson , we absolutely *do* have doctests, and in his opinion (now also mine :wink:) we shouldn't *either* stop running doctests *or* restrict what we put in examples just to fix this.
Instead, to keep CI speed up, we should probably do this as part of a CI docs-build.
That could be via #995 -- or if that causes problems maybe an additional travis matrix element ?

I'm going to close this PR until one of those provides a better way to expand examples dependencies.

PR #1025, opened 2018-02-16
kav7 crs

pierre-prandi:
based on current implementation of Robinson projection

## Rationale

Implementing the Kav7 projection

## Implications

Should be limited to this crs

## Checklist

 * simply plotting data on a map (looks better than Robinson for my application), 
kav7


pelson:
Thanks for submitting this @pierre-prandi. Looks like a great projection to have on board.
I'm going to be quite unavailable for all of March, and am going to struggle to do this PR justice before then.

A few thoughts from me, and hopefully others will be able to help with getting this merged:

 * KavrayskiyVII is probably a better name than Kav7 (thank goodness for autocomplete 😄)
 * I think you've included an extra crs.py accidentally into your commit.
 * Add some rudimentary testing to ensure that we can use this projection to transform coordinates (i.e. it works with proj4).


That's about all I have for now, but I haven't yet reviewed the code properly.


Thanks!

pierre-prandi:
@pelson  the nan treatment was simply copied from the current implementation of the Robinson projection, I'm not certain if/why it is needed here, i'll try to look at it.

pierre-prandi:
Basic testing shows this seems to work well, I'm able to plot gridded data with no apparent problems
kav7_map

lib/cartopy/crs.py
```diff
@@ -1563,6 +1563,18 @@ def transform_points(self, src_crs, x, y, z=None):
...
+    def __init__(self, central_longitude=0, globe=None):
+        proj4_params = [('proj', 'kav7'), ('lon_0', central_longitude)]
+        super(KavrayskiyVII, self).__init__(
+            proj4_params,
+            central_longitude,
+            globe=globe)
+
+    @property
+    def threshold(self):
+        return 1e4
```
pelson: FYI, these thresholds are ripe for improving in cartopy. I'm confident that we will be changing how this works before we cut v1.0 of cartopy. No action on you here, just wanted you to be aware.

pelson:
> the nan treatment was simply copied from the current implementation of the Robinson projection

Great. I have a feeling that Nan treatment for Robinson was the result of some pain in older proj.4 versions. If you are interested, a subsequent activity worth pursuing might be to see what happens when we remove the code from the Robinson projection too. Let's not do that in this PR though 😄

pierre-prandi:
I'll try to give the Robinson projection a look over the summer, 
It appears my last commit did not pass the tests, however, looking at the error messages, I'm not sure this is actually related to the changes I made.
Could somebody look into this ?

pelson:
> It appears my last commit did not pass the tests, however, looking at the error messages, I'm not sure this is actually related to the changes I made.
> Could somebody look into this ?

Agreed. I'll be able to put some effort onto the tests over the next week I hope. That should unblock a number of PRs that are good to go. While we are waiting for that, would you mind adding a simple test (asserting that the proj4 string is what you expect it to be is good enough), and a what's new so that we document this new change in the release notes?

pierre-prandi:
Basic (really basic) testing added, and I updated the what's new file.

pierre-prandi:
Seems to me that checks do not fail because of the python code (ci conf problem?).
Is there something I can do so that this PR can be merged ?

lib/cartopy/tests/crs/test_kavrayskiy_vii.py
```diff
@@ -0,0 +1,28 @@
+# (C) British Crown Copyright 2015 - 2017, Met Office
```
dopplershift: '2015 - 2017' -> '2018'

dopplershift:
Would it be possible for you to rebase this on current master? That will likely clean up the testing issues.

pelson:
This appears to have stalled. Would be very happy to see this projection in cartopy, and there is probably not much work beyond what has already been done to get it up to speed. Happy if anybody else wants to take this on and get it over the finish line.

Issue #803, opened 2016-10-01
Errors in coastline filling in Lambert Conformal conic projection


dennissergeev:
I noticed that coastlines are not filled correctly when I use Lambert conic projection. For example,

```python
import cartopy
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

fig = plt.figure()                                                
                                                                                                                                                                   
ax = fig.add_subplot(111, projection=ccrs.LambertConformal())                         
feature = cartopy.feature.NaturalEarthFeature(name='coastline', category='physical',
                                              scale='110m',
                                              edgecolor='#000000', facecolor='#AAAAAA')
ax.add_feature(feature)
plt.show()
```
results in
![coastline_fill_bug](https://cloud.githubusercontent.com/assets/12111288/19017147/e537474a-8827-11e6-8788-4196081e44a3.png)

It happens at all coastline scales and with different extents of the map, so I guess it's a bug. Maybe it's related to #579

dopplershift:
I'm getting the same problem with the oceans on master right now:

```python
proj = cartopy.crs.LambertConformal(central_latitude=25, central_longitude=-97,
                                                               standard_parallels=[25])
ax = plt.axes(projection=proj)
ax.add_feature(cartopy.feature.OCEAN)
ax.coastlines()
ax.set_extent([-140, -70, 20, 60])
```
![image](https://cloud.githubusercontent.com/assets/221526/21908715/20de636e-d8d2-11e6-9c2a-4ca668ac66b0.png)

Does not occur if we use `PlateCaree()` instead:

![image](https://cloud.githubusercontent.com/assets/221526/21908728/2f06b310-d8d2-11e6-9026-e248899212aa.png)

dopplershift:
Also works fine if we use `standard_parallels=[35]` for the Lambert projection.

akrherz:
May I add `LambertAzimuthalEqualArea` to the list or reproducers with current master + matplotlib2
```python
proj = cartopy.crs.LambertAzimuthalEqualArea(central_latitude=45,
                                             central_longitude=-100)
ax = plt.axes(projection=proj)
ax.add_feature(cartopy.feature.OCEAN)
ax.coastlines()
ax.set_extent([-140, -70, 20, 60])
```

![test](https://user-images.githubusercontent.com/210858/31137800-e7e69100-a831-11e7-80ae-634a336f19c4.png)

byersiiasa:
[...]

PR #1029, opened 2018-02-21
Better support colour overrides for Features.

QuLogic:
If a Feature defines an `edgecolor` or `facecolor` but the user supplies plain `color` when add the feature to a plot, then which one wins depends on the dictionary and keyword sort order. This makes it more explicit by forcing `color` to override both.

See for example, https://github.com/SciTools/cartopy/issues/803#issuecomment-366686339 for a case where this difference comes up.

### DIFF
--- a/lib/cartopy/mpl/feature_artist.py
+++ b/lib/cartopy/mpl/feature_artist.py
@@ -96,6 +96,12 @@ class FeatureArtist(matplotlib.artist.Artist):
             kwargs = {}
         self._kwargs = dict(kwargs)
 
+        if 'color' in self._kwargs:
+            # We want the user to be able to override both face and edge
+            # colours if the original feature already supplied it.
+            color = self._kwargs.pop('color')
+            self._kwargs['facecolor'] = self._kwargs['edgecolor'] = color
+
         # Set default zorder so that features are drawn before
         # lines e.g. contours but after images.
         # Note that the zorder of Patch, PatchCollection and PathCollection


lib/cartopy/mpl/feature_artist.py
```diff
@@ -96,6 +96,12 @@ def __init__(self, feature, **kwargs):
             kwargs = {}
         self._kwargs = dict(kwargs)
 
+        if 'color' in self._kwargs:
+            # We want the user to be able to override both face and edge
+            # colours if the original feature already supplied it.
+            color = self._kwargs.pop('color')
+            self._kwargs['facecolor'] = self._kwargs['edgecolor'] = color
```
pelson: I think these should be using ``setdefault``, not setitem...

```
Feature(..., edgecolor='red', color='black')._kwargs == {'edgecolor': 'black', 'facecolor': 'black'}
```

What I'd expect with the above call is ```{'edgecolor': 'red', 'facecolor': 'black'}```

QuLogic: I was 50/50 on which one should override the other. I couldn't figure out which one has precedence in Matplotlib from a quick glance.

QuLogic:
> I'd advocate doing this at the final stage of the stylisation (once we have computed the compound style), rather than at the __init__.

The point of doing this now is that it _will_ affect the compound style. If we propagate `color` after, we don't know if that's because the user set it or the feature did. For example, `BORDERS` sets `edgecolor`, so if the user sets `color`, then this will propagate over it. But if a `Feature` used `color` and the user set `edgecolor`, then doing it at the final stage will override the user's setting.

lib/cartopy/mpl/feature_artist.py
QuLogic: @WeatherGod pointed out that there is a precedence in Matplotlib which is applied [here](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/artist.py#L986) and gives `color` precedence over the others, so this should be working in the correct way.

pelson:
> But if a Feature used color (though none of ours do) and the user set edgecolor, then doing it at the final stage will override the user's setting.

The problem is that the behaviour is ambiguous (and we are inheriting that from matplotlib itself).

How about we take the stance that is anybody sets the ```color``` argument at *any point of stylization* (be it in the Feature definition [never in core cartopy!], or later on in the call stack, we **always** apply ``color`` at a higher precedence than ``edgecolor`` and ``facecolor``. As you say, that means once ```color``` is set, ```edgecolor``` and ```facecolor``` can **never** be overridden.

I'm advocating this approach as I don't feel that the following case is particularly consistent with matplotlib's precedence:

```
> f = Feature(..., edgecolor='red', color='black')
> ax.add_feature(f, edgecolor='blue')
```

Should, IMO, produce:

```
===> feature style: {'edgecolor': 'black', 'facecolor': 'black'}
```

Not:

```
===>  feature style: {'edgecolor': 'blue', 'facecolor': 'black'}
```

But:

```
> f = Feature(..., edgecolor='red', color='black')
> ax.add_feature(f, color='pink')
```

should produce

```
{'edgecolor': 'pink', 'facecolor': 'pink'}
```

FWIW, I agree with you completely regarding consistency with mpl - I think were saying that on its own ```f``` style should definitely be ```{'edgecolor': 'black', 'facecolor': 'black'}```.

pelson:
So I'm a little stuck here. Some other perspectives would be extremely helpful (e.g. @weathergod / @dopplershift / @ajdawson).

Essentially, I'm really uncomfortable with ignoring keywords as is the case with:

```
Feature(..., edgecolor='red', color='black')._kwargs == {'edgecolor': 'black', 'facecolor': 'black'}
```

However, it is pointed out that this is the behaviour of matplotlib (I haven't actually confirmed that myself). I'm very tempted to introduce warnings for this somewhat smelly 💩 behaviour.

Furthermore, this is complicated by the complexity that I know this will introduce to things like #1019, where colorization may be done at multiple levels (FEATURE defaults, and as a callback for functional control of style). 

All of this leads to the next obvious issue of handling stylization aliases (e.g. ``lw`` vs ``linewidth``). My approach so far has been to ignore the issue, but it is probably time to address it.

**Note to self: I've found writing this helpful in clearing up what the issue is that I'm addressing here, hope it has been as useful for you 😃.**

### Proposal

Cartopy needs to gain some functionality to merge/compose styles from one or more dictionaries, and for the composed result to be consistent whether done all at once vs iteratively. 

This suggests we need alias expansion (e.g. ``lw -> linewidth``) and keyword normalization (e.g. ``colour -> edgecolor + facecolor``). This can be the canonical place to document precedence (e.g. whether ``color`` trumps ``[edge|face]color``), and the place to implement bespoke styling (e.g. it should be possible for a Feature to define a ``facecolor`` that unless explicitly overridden, setting a ``color`` will not change the default). This final constraint suggests we need a finalization step.

I don't think https://het.as.utexas.edu/HET/Software/Matplotlib/users/style_sheets.html#composing-styles cuts the mustard in terms of the complexity that I've covered here.

Unless we have explicitly intercept all GeoAxes calls, we need to ensure that we remain consistent with the matplotlib interface. Of course, we are in uncharted territory when it comes to things like ``.add_feature``, ``.add_geometries``, ``.coastlines()`` because they introduce the idea of pre-defined styles that can be overridden.

### API

#### cartopy.mpl.style

Where one goes to find out about how cartopy's matplotlib stylisation works.

#### cartopy.mpl.style.merge

Join together one or more style dictionaries.
This expands aliases and has well defined precedence.

#### cartopy.mpl.style.finalize

Given a single (merged) style dictionary, returns a style dictionary that can be passed on to matplotlib, removing any cartopy implementation details (such as the ability to override precedence for things like facecolor).


# Next steps

 * @QuLogic and I to agree whether to merge and move forwards with this PR as it stands, or to implement a first-pass of the above API
 * Implementing the API above

pelson:
I've taken a shot at implementing the proposed API, but taking into account the precedent that @QuLogic has put forwards here. The implication of that PR (and @QuLogic's proposal) is that you can get different results for your style depending on whether you chain your styles or do them in one shot:

```
In [1]: from cartopy.mpl.style import merge                                                                                                                          

In [2]: merge({'color': 'red'}, {'facecolor': 'blue'})                                                                                                               
Out[2]: {'edgecolor': 'red', 'facecolor': 'blue'}

In [3]: merge({'color': 'red', 'facecolor': 'blue'})                                                                                                                 
Out[3]: {'facecolor': 'red', 'edgecolor': 'red'}
```

On reflection, whilst not ideal, I'm OK with this constraint given the consistency that we get with matplotlib with regards to the behaviour of color and [face|edge]color.

WeatherGod:
Why would `facecolor` be red in the second example? It was explicitly set to blue in the given dictionary?

I do totally get why merging two dictionaries would produce a different result than passing in a single dictionary of the same values. I see it as a feature. So, if `color` was in the second dictionary, it would make sense to me to have that override the previous setting of `facecolor`.

pelson:
> Why would facecolor be red in the second example? It was explicitly set to blue in the given dictionary?

That's what matplotlib does.

```
import matplotlib.patches as mpatches
ax = plt.axes()
ax.add_patch(mpatches.Circle([0, 0], 1, facecolor='red', color='blue', lw=4))
plt.show()
```
no_red

dopplershift:
I'm going to merge this for 0.17, and hopefully we can get #1190 in 0.18.

PR #1042, opened 2018-02-28
Normalize grid coordinates used for interpolation.

QuLogic:
Most coordinates are in metres and thus fairly large. This can cause precision issues in Qhull as it calculates distances via squares and square roots, at least on non-x86_64 architectures.

This is equivalent to the `rescale=True` argument in SciPy, but doing it ourselves means a) we don't need to bump requirements to 0.14 and b) re-scaling is not done for every scalar/vector property.

Note that while this does cause some changes in results, the results are now consistent across all Fedora-supported arches.


ajdawson:
The changes to the test images look a bit strange over Greenland, what is going on there?

dopplershift:
The changes to the barbs/quiver look like an improvement. I also wonder about the changes to the stream plot images. I *guess* I can squint and see how the new stream plot images are consistent with the new barbs/quiver...

Either way the actual code changes look innocuous enough so I guess that's further evidence of the instabilities in the current code.

QuLogic:
If you believe the barbs are okay, then it does seem pretty consistent. There is definitely a change in direction between those top two layers, but I guess where the transition happens is a bit hard to say. Here are the two overlaid in the old mode:
old
and if we look at the old streamlines over the new barbs, they are definitely inconsistent:
old-over-new
whereas new streamlines and new barbs seem to align better:
new

QuLogic:
Works for me.

pelson:
Rebasing this now will remove some of the image files.
Merge when green. 👍

pelson:
Looks like a few genuine image failures remaining @QuLogic?

pelson:
OK, I'm going to take this one on the chin and merge #1042 even though the tests are failing. The reason I'm doing this is that:

a) master's tests have been failing for way too long
b) I want to get masters test's working again, and there are tendrils to doing that in this PR

pelson:
Thanks @QuLogic. This is a really good change 👍

PR #1047, opened 2018-03-07
shapely.geometry.LinearRing needs 4+ points

akrherz:
avoids shapely error that leads to segmentation fault:
ERROR:shapely.geos:IllegalArgumentException: Invalid number of points in LinearRing found 3 - must be 0 or >= 4

## Rationale

A `LinearRing` should have at least four points, not three.

## Implications

I stop seeing segmentation faults!

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -510,7 +510,7 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
 
         # filter out any non-valid linear rings
         processed_ls = [linear_ring for linear_ring in processed_ls if
-                        len(linear_ring.coords) > 2]
+                        len(linear_ring.coords) > 3]
 
         linear_rings = [sgeom.LinearRing(line) for line in processed_ls]
 
--- a/lib/cartopy/tests/mpl/test_contour.py
+++ b/lib/cartopy/tests/mpl/test_contour.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2016 - 2017, Met Office
+# (C) British Crown Copyright 2016 - 2018, Met Office
 #
 # This file is part of cartopy.
 #
@@ -21,6 +21,8 @@ import matplotlib.pyplot as plt
 from matplotlib.testing.decorators import cleanup
 import numpy as np
 from numpy.testing import assert_array_almost_equal
+from scipy.interpolate import NearestNDInterpolator
+from scipy.signal import convolve2d
 
 import cartopy.crs as ccrs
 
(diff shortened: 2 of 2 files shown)


QuLogic:
It seems like this broke some tests.

akrherz:
pinging @dopplershift to fulfill an action item from a recent meeting with him :)

I am unsure how to proceed with this.  I continue to need this patch to keep cartopy from sometimes segfaulting on contouring scripts, but breaking tests is not good either.  I could remove the failing tests with an additional commit, hehe :)

My current [travis-ci builds](https://travis-ci.org/akrherz/pyIEM) of my hacky library crash due to this issue (I have not efforted providing a patched cartopy for my travis-ci builds).  So perhaps that would be a means to reproduce this?
```
pyiem/tests/test_plot.py ........../home/travis/.travis/job_stages: line 57:  7720 Segmentation fault      (core dumped) coverage run --source=pyiem setup.py test
```

pelson:
@akrherz - do you have a simple test case that we can add by any chance? Ideally, we would avoid the need to generate contours to reproduce the issue. I wrote up some notes many moons ago about doing such a thing: https://github.com/SciTools/cartopy/wiki/DEVEL:-Isolating-misbehaving-polygons-in-cartopy's-matplotlib-interface. Note: I'm not saying it will be easy or quick, so make a judgement about how far down the rabbit-hole you go (and let me know) 😃 

I completely agree with the LinearRing condition for 4 or more coords. However, the one thing more dangerous than no comments and bad variable names is incorrect comments and variable names. We therefore need to be absolutely certain that we *really do* only have linear rings... (by validating the code, rather than introducing isinstances into the code).

Given the above, I'm far more interested in ensuring that the docs build correctly - they are a pretty good indicator of representative real-world code. For that reason, I'm going to close + reopen this PR to see if I can get CircleCI to re-run (not sure if that actually triggers it or not, so bare with me 😃)

QuLogic:
CircleCI only runs on PR HEAD, not the merge, so unfortunately closing and re-opening will not fix it. A rebase is necessary to get the fixes from `master`.

akrherz:
@QuLogic rebased, thank you

akrherz:
I did it, a repo :)
```python
import cartopy.crs as ccrs
from scipy.interpolate import NearestNDInterpolator
from scipy.signal import convolve2d

import matplotlib
matplotlib.use('agg')
import numpy as np


def contourf(ax, lons, lats, vals, clevs):
    xbnds = ax.get_xlim()
    ybnds = ax.get_ylim()
    ll = ccrs.Geodetic().transform_point(xbnds[0], ybnds[0], ax.projection)
    ul = ccrs.Geodetic().transform_point(xbnds[0], ybnds[1], ax.projection)
    ur = ccrs.Geodetic().transform_point(xbnds[1], ybnds[1], ax.projection)
    lr = ccrs.Geodetic().transform_point(xbnds[1], ybnds[0], ax.projection)
    xi = np.linspace(min(ll[0], ul[0]), max(lr[0], ur[0]), 100)
    yi = np.linspace(min(ll[1], ul[1]), max(ul[1], ur[1]), 100)
    xi, yi = np.meshgrid(xi, yi)
    nn = NearestNDInterpolator((lons, lats), vals)
    vals = nn(xi, yi)
    lons = xi
    lats = yi
    window = np.ones((6, 6))
    vals = convolve2d(vals, window / window.sum(), mode='same',
                      boundary='symm')
    ax.contourf(lons, lats, vals, clevs,
                transform=ccrs.PlateCarree())



ax = plt.axes([0.01, 0.05, 0.898, 0.85], projection=ccrs.Mercator(),
              aspect='equal')
ax.set_extent([-99.6, -89.0, 39.8, 45.5])
contourf(ax, np.arange(-94, -85), np.arange(36, 45), np.arange(9),
         np.arange(9))
plt.savefig('test.png')
```

```
$ python pr1047.py
IllegalArgumentException: Invalid number of points in LinearRing found 3 - must be 0 or >= 4
Segmentation fault
```
with my patch, no segfault and we get this lovely thing
test

dopplershift:
@akrherz I added the test, rebased, and pushed up to your branch in the hopes that we can get this merged in a few minutes.

dopplershift:
So this is causing breakage elsewhere, so I don't think it's the correct fix, at least by itself.

akrherz:
Away from my desk, but did you see my comments above about the tests being
wrong?

dopplershift:
See #1090 for an alternate fix that seems to work (at least for me locally). @akrherz you may want to test with some of the proper data examples you've have crash.

dopplershift:
Linear ring can take 3 points, in which case it will automatically close it, if first and last are different. Your failing case causes 3 identical points.

pelson:
Thanks for this PR @akrherz. @dopplershift has kindly taken this under his wing and managed to isolate the issue further in #1090, which has now been merged. The fix will be available in v0.17.0. 👍

PR #1050, opened 2018-03-09
Reorder Stereographic args to avoid API breaking 0.15 -> 0.16.

pp-mo:
In #929 a new argument was added into `Stereographic.__init__`, but not at the end.
This affects the meaning of usages that rely on argument-order, and broke a couple of tests in Iris.
See : https://github.com/SciTools/iris/pull/2971

**Do we want to consider putting this back ?**

I realise 
  1. it may look neater to put `globe` last ? (though other classes now don't)
  1. it's another breaking change (!)


dopplershift:
Oops. I think we need @pelson to weigh in  on which to break things.

PR #1052, opened 2018-03-16
force string conversion for wmts row, col and tile set id

acrosby:
## Rationale

Fix for `owslib` wmts get tiles action error coming from row, col as `int` rather than `str` which seems to work.

## Implications

This should probably also be fixed in https://github.com/geopython/OWSLib, but we can protect ourselves from these errors here if they won't be handled in `owslib`. Found this issue when trying to use external WMTS sources, related to Issue #1049


acrosby:
Sure, I will open an issue and link back.

PR #1053, opened 2018-03-16
add epsg 3857 support to wmts

acrosby:
## Rationale

Support external WMTS sources that use EPSG 3857 Google Maps/Pseudo-Mercator projections. This includes ESRI basemaps and other commercial vendor sources.

## Implications

This should only allow additional urn projection string matches to the existing `crs.GOOGLE_MERCATOR` projection using a commonly found EPSG code.


acrosby:
Just rebased against the master (I think), any movement in getting this in a release?

PR #1089, opened 2018-07-15
Add inline grid meridian/parallel labeling

ahuang11:
Work in progress after the newest merge; I know I haven't fully updated the tests! (Ignore the red s, I was using that as a reference)

image
image
image

PR #1117, opened 2018-09-20
Add support for all projections to lon/lat labelling

stefraynaud:
## Rationale

Longitude / latitude labelling is restricted Mercator and PlateCarree projections.
This restriction can be removed by computing the position of labels as the intersection of grid lines with the map and plot boundaries. This make it compatible with all projections.

In the current version, only labels that are the border of the plot are displayed, not those at the border of the map. This limitation can be overcome by better checking if a label is plottedover another object, like a label or the map.  

The code below produces the attached figures.

````python
import cartopy.crs as ccrs


plain_crs = ccrs.PlateCarree()
mercator_crs = ccrs.Mercator()
rotated_crs = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=45.0)
igh_crs = ccrs.InterruptedGoodeHomolosine()
polar_crs = ccrs.SouthPolarStereo()
uk_extent = [-10, 3, 49, 59]
north_atlantic_extent = [-90, 0, -10, 70]
polar_extent = [-180, 180, -90, -60]

to_test = [(plain_crs, uk_extent),
           (mercator_crs, uk_extent),
           (rotated_crs, uk_extent),
           (rotated_crs, north_atlantic_extent),
           (igh_crs, north_atlantic_extent),
           (polar_crs, polar_extent)]

for projection, extent in to_test:

    fig = plt.figure(figsize=(6, 6))
    ax = fig.add_subplot(1, 1, 1, projection=projection)
    ax.set_extent(extent, crs=plain_crs)
    ax.coastlines(resolution='50m', color='k')
    ax.gridlines(color='lightgrey', linestyle='-', draw_labels=True)
    plt.savefig(__file__[:-2] + projection.__class__.__name__.lower() + '.png')

plt.show()
````
ticks platecarree
ticks mercator
ticks rotatedpole
ticks interruptedgoodehomolosine
ticks southpolarstereo

## Implications

Longitude and latitude labels can be plotted for all projections.
Another implication is that x/y labels are no longer associated with respectively longitudes/latitudes. This must be clarified in the internal API at least.

Note : This PR must be merged with PR #1089

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -23,6 +23,7 @@ import matplotlib.text as mtext
 import matplotlib.ticker as mticker
 import matplotlib.transforms as mtrans
 import numpy as np
+import shapely.geometry as sgeom
 
 import cartopy
 from cartopy.crs import Projection, _RectangularProjection
@@ -90,6 +91,24 @@ LATITUDE_FORMATTER = mticker.FuncFormatter(lambda v, pos:
                                            _north_south_formatted(v))
 
 
+def _text_artists_overlaps_(artista, artistb, axis):
+    """Check that two text artists don't overlap (approximately)"""
+    xa, ya = artista.get_position()
+    xb, yb = artistb.get_position()
+    trans = artista.get_transform()
+    xa, ya = trans.transform_point((xa, ya))
+    xb, yb = trans.transform_point((xb, yb))
+    size = artista.get_size()
+    if axis == 'x':
+        factor = 0.8 if 'monospace' not in artista.get_family() else 1
+        dxa = size * len(artista.get_text()) / 2 * factor
+        dxb = size * len(artistb.get_text()) / 2 * factor
+        return not ((xa + dxa) < (xb - dxb) or (xb + dxb) < (xa - dxa))
+    dya = size / 2
+    dyb = size / 2
+    return not ((ya + dya) < (yb - dyb) or (yb + dyb) < (ya - dya))
+
+
 class Gridliner(object):
     # NOTE: In future, one of these objects will be add-able to a GeoAxes (and
     # maybe even a plain old mpl axes) and it will call the "_draw_gridliner"
@@ -137,12 +156,10 @@ class Gridliner(object):
         self.ylocator = ylocator or degree_locator
 
         #: The :class:`~matplotlib.ticker.Formatter` to use for the x labels.
-        self.xformatter = mticker.ScalarFormatter()
-        self.xformatter.create_dummy_axis()
+        self.xformatter = LONGITUDE_FORMATTER
 
         #: The :class:`~matplotlib.ticker.Formatter` to use for the y labels.
-        self.yformatter = mticker.ScalarFormatter()
-        self.yformatter.create_dummy_axis()
+        self.yformatter = LATITUDE_FORMATTER
 
         #: Whether to draw labels on the top of the map.
         self.xlabels_top = draw_labels
@@ -177,12 +194,6 @@ class Gridliner(object):
         self.ypadding = 5
 
         self.crs = crs
-
-        # if the user specifies tick labels at this point, check if they can
-        # be drawn. The same check will take place at draw time in case
-        # public attributes are changed after instantiation.
-        if draw_labels:
-            self._assert_can_draw_ticks()
 
         #: The number of interpolation points which are used to draw the
         #: gridlines.
@@ -219,7 +230,7 @@ class Gridliner(object):
             transform = transform._as_mpl_transform(self.axes)
         return transform
 
-    def _add_gridline_label(self, value, axis, upper_end):
+    def _add_gridline_label(self, loc, text, axis, edge):
         """
         Create a Text artist on our axes for a gridline label.
 
@@ -230,69 +241,60 @@ class Gridliner(object):
             value, and is positioned centred at that point.
         axis
             Which axis the label is on: 'x' or 'y'.
-        upper_end: bool
-            If True, place at the maximum of the "other" coordinate (Axes
-            coordinate == 1.0).  Else 'lower' end (Axes coord = 0.0).
+        edge: str
+            If "top" or "right", place at the maximum of the "other"
+            coordinate (Axes coordinate == 1.0).  Else 'lower' end
+            (Axes coord = 0.0).
 
         """
-        transform = self._crs_transform()
-        if upper_end:
-            shift_scale = 1
-        else:
-            shift_scale = -1
+        # Get label specs
         if axis == 'x':
-            x = value
-            y = 1.0 if upper_end else 0.0
-            h_align = 'center'
-            v_align = 'bottom' if upper_end else 'top'
-            tr_x = transform
-            tr_y = self.axes.transAxes + \
-                mtrans.ScaledTranslation(
-                    0.0,
-                    shift_scale * self.xpadding * (1.0 / 72),
-                    self.axes.figure.dpi_scale_trans)
-            str_value = self.xformatter(value)
+            x = loc
+            if edge == 'top':
+                y = 1.0
+                meth = self.axes.get_xaxis_text2_transform
+            else:
+                y = 0.0
+                meth = self.axes.get_xaxis_text1_transform
+            label_transform, v_align, h_align = meth(self.xpadding)
             user_label_style = self.xlabel_style
         elif axis == 'y':
-            y = value
-            x = 1.0 if upper_end else 0.0
+            y = loc
+            if edge == 'right':
+                x = 1.0
+                meth = self.axes.get_yaxis_text2_transform
+            else:
+                x = 0.0
+                meth = self.axes.get_yaxis_text1_transform
+            label_transform, v_align, h_align = meth(self.xpadding)
             if matplotlib.__version__ > '2.0':
                 v_align = 'center_baseline'
             else:
                 v_align = 'center'
-            h_align = 'left' if upper_end else 'right'
-            tr_y = transform
-            tr_x = self.axes.transAxes + \
-                mtrans.ScaledTranslation(
-                    shift_scale * self.ypadding * (1.0 / 72),
-                    0.0,
-                    self.axes.figure.dpi_scale_trans)
-            str_value = self.yformatter(value)
             user_label_style = self.ylabel_style
         else:
             raise ValueError(
                 "Unknown axis, {!r}, must be either 'x' or 'y'".format(axis))
 
-        # Make a 'blended' transform for label text positioning.
-        # One coord is geographic, and the other a plain Axes
-        # coordinate with an appropriate offset.
-        label_transform = mtrans.blended_transform_factory(
-            x_transform=tr_x, y_transform=tr_y)
-
+        # Create and add a Text artist with these properties
         label_style = {'verticalalignment': v_align,
                        'horizontalalignment': h_align,
                        }
         label_style.update(user_label_style)
-
-        # Create and add a Text artist with these properties
-        text_artist = mtext.Text(x, y, str_value,
+        text_artist = mtext.Text(x, y, text,
                                  clip_on=False,
                                  transform=label_transform, **label_style)
-        if axis == 'x':
-            self.xlabel_artists.append(text_artist)
-        elif axis == 'y':
-            self.ylabel_artists.append(text_artist)
+
+        # Check that this artist don't overlap this one
+        artists = getattr(self, '{}label_{}_artists'.format(axis, edge))
+        for ta in artists:
+            if _text_artists_overlaps_(ta, text_artist, axis):
+                return
+
+        # Ok, register it
+        artists.append(text_artist)
         self.axes.add_artist(text_artist)
+        return text_artist
 
     def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
         """Create Artists for all visible elements and add to our Axes."""
@@ -307,6 +309,10 @@ class Gridliner(object):
 
         x_ticks = self.xlocator.tick_values(x_lim[0], x_lim[1])
         y_ticks = self.ylocator.tick_values(y_lim[0], y_lim[1])
+
+        #####################
+        # Gridlines drawing #
+        #####################
 
         # XXX this bit is cartopy specific. (for circular longitudes)
         # Purpose: omit plotting the last x line, as it may overlap the first.
@@ -327,27 +333,28 @@ class Gridliner(object):
         collection_kwargs.setdefault('linestyle', rc_params['grid.linestyle'])
         collection_kwargs.setdefault('linewidth', rc_params['grid.linewidth'])
 
-        if self.xlines:
-            lines = []
-            for x in x_gridline_points:
-                ticks = list(zip(
-                    np.zeros(n_steps) + x,
-                    np.linspace(min(y_ticks), max(y_ticks), n_steps)))
-                lines.append(ticks)
+        # Longitude lines
+        lon_lines = []
+        for x in x_gridline_points:
+            ticks = list(zip(
+                np.zeros(n_steps) + x,
+                np.linspace(min(y_ticks), max(y_ticks), n_steps)))
+            lon_lines.append(ticks)
 
-            x_lc = mcollections.LineCollection(lines, **collection_kwargs)
+        if self.xlines:
+            x_lc = mcollections.LineCollection(lon_lines, **collection_kwargs)
             self.xline_artists.append(x_lc)
             self.axes.add_collection(x_lc, autolim=False)
 
+        # Latitude lines
+        lat_lines = []
+        for y in y_ticks:
+            ticks = list(zip(
+                np.linspace(min(x_ticks), max(x_ticks), n_steps),
+                np.zeros(n_steps) + y))
+            lat_lines.append(ticks)
         if self.ylines:
-            lines = []
-            for y in y_ticks:
-                ticks = list(zip(
-                    np.linspace(min(x_ticks), max(x_ticks), n_steps),
-                    np.zeros(n_steps) + y))
-                lines.append(ticks)
-
-            y_lc = mcollections.LineCollection(lines, **collection_kwargs)
+            y_lc = mcollections.LineCollection(lat_lines, **collection_kwargs)
             self.yline_artists.append(y_lc)
             self.axes.add_collection(y_lc, autolim=False)
 
@@ -355,30 +362,68 @@ class Gridliner(object):
         # Label drawing #
         #################
 
-        # Trim outside-area points from the label coords.
-        # Tickers may round *up* the desired range to something tidy, not
-        # all of which is necessarily visible.  We must be stricter with
-        # our texts, as they are drawn *without clipping*.
-        x_label_points = [x for x in x_ticks if x_lim[0] <= x <= x_lim[1]]
-        y_label_points = [y for y in y_ticks if y_lim[0] <= y <= y_lim[1]]
+        self.xlabel_bottom_artists = []
+        self.xlabel_top_artists = []
+        self.ylabel_left_artists = []
+        self.ylabel_right_artists = []
+        if not (self.ylabels_left or self.ylabels_right or
+                self.xlabels_bottom or self.xlabels_top):
+            return
 
-        if self.xlabels_bottom or self.xlabels_top:
-            self._assert_can_draw_ticks()
-            self.xformatter.set_locs(x_label_points)
-            for x in x_label_points:
-                if self.xlabels_bottom:
-                    self._add_gridline_label(x, axis='x', upper_end=False)
-                if self.xlabels_top:
-                    self._add_gridline_label(x, axis='x', upper_end=True)
+        # Get the real map boundaries
+        x0, x1 = self.axes.get_xlim()
+        y0, y1 = self.axes.get_ylim()
+        plot_boundary = sgeom.Polygon([[x0, y0], [x1, y0],
+                                       [x1, y1], [x0, y1]])
+        map_boundary_vertices = self.axes.outline_patch.get_path().vertices
+        map_boundary = sgeom.Polygon(map_boundary_vertices)
+        map_boundaries = plot_boundary.intersection(map_boundary)
 
-        if self.ylabels_left or self.ylabels_right:
-            self._assert_can_draw_ticks()
-            self.yformatter.set_locs(y_label_points)
-            for y in y_label_points:
-                if self.ylabels_left:
-                    self._add_gridline_label(y, axis='y', upper_end=False)
-                if self.ylabels_right:
-                    self._add_gridline_label(y, axis='y', upper_end=True)
+        # Loop on longitude and latitude lines and collect what to draw
+        to_draw = {}
+        for lonlat, lines, line_ticks, formatter in (
+                ('lon', lon_lines, x_ticks, self.xformatter),
+                ('lat', lat_lines, y_ticks, self.yformatter)):
+
+            to_draw[lonlat] = {'y': {'left': [], 'right': []},
+                               'x': {'bottom': [], 'top': []}}
+
+            for line, tick_value in zip(lines, line_ticks):
+
+                ls = sgeom.LineString(line)
+                lsp = self.axes.projection.project_geometry(ls)
+
+                if lsp.intersects(map_boundaries):
+
+                    lsb = lsp.intersection(map_boundaries)
+                    if lsb.is_empty:
+                        continue
+
+                    lsb = lsb.boundary
+                    for point in lsb:
+                        x = point.x
+                        y = point.y
+                        text = formatter(tick_value)
+                        if x == x0:
+                            loc, axis, edge = y, 'y', 'left'
+                        elif x == x1:
+                            loc, axis, edge = y, 'y', 'right'
+                        elif y == y0:
+                            loc, axis, edge = x, 'x', 'bottom'
+                        elif y == y1:
+                            loc, axis, edge = x, 'x', 'top'
+                        else:
+                            continue
+
+                        if getattr(self, '{}labels_{}'.format(axis, edge)):
+                            to_draw[lonlat][axis][edge].append((loc, text))
+
+        # Draw in the prefered order
+        for lonlat, axis in (('lon', 'x'), ('lat', 'y'),
+                             ('lon', 'y'), ('lat', 'x')):
+            for edge, lts in to_draw[lonlat][axis].items():
+                for loc, text in lts:
+                    self._add_gridline_label(loc, text, axis, edge)
 
     def _assert_can_draw_ticks(self):
         """

### CODE REVIEW

File: lib/cartopy/mpl/gridliner.py [hunk 1]
```diff
@@ -219,7 +237,7 @@ def _crs_transform(self):
             transform = transform._as_mpl_transform(self.axes)
         return transform
 
-    def _add_gridline_label(self, value, axis, upper_end):
+    def _add_gridline_label(self, loc, text, axis, edge):
```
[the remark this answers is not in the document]
stefraynaud: Loc is no longer in degrees but in projection coordinates. 
The corresponding value in degrees used for the label cannot be exactly inferred by inverse projection because is computed as the intersection an approximate grid line with the edge of the plot.
The formatted label is now directly passed because otherwise it would be neccessary to pass the formatter too.

lib/cartopy/mpl/gridliner.py
```diff
@@ -230,69 +248,60 @@ def _add_gridline_label(self, value, axis, upper_end):
             value, and is positioned centred at that point.
```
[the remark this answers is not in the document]
stefraynaud: Docstring is fixed

lib/cartopy/mpl/gridliner.py
```diff
@@ -230,69 +248,60 @@ def _add_gridline_label(self, value, axis, upper_end):
...
+        # Check that this artist don't overlap this one
+        artists = getattr(self, '{}label_{}_artists'.format(axis, edge))
+        for ta in artists:
+            if _text_artists_overlaps_(ta, text_artist, axis):
+                return
+
+        # Ok, register it
+        artists.append(text_artist)
         self.axes.add_artist(text_artist)
+        return text_artist
```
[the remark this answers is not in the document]
stefraynaud: I removed the return statement.

stefraynaud:
Her is the kind of tick labels the last commit allows.
ticks mercator_-3 8--2 9-50 49-50 74
Degrees, minutes, and seconds can be auto hidden when not required.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1161,8 +1161,8 @@ def imshow(self, img, *args, **kwargs):
 #            result.set_clip_path(self.outline_patch)
         return result
 
-    def gridlines(self, crs=None, draw_labels=False, xlocs=None,
-                  ylocs=None, **kwargs):
+    def gridlines(self, crs=None, draw_labels=False, mlocs=None,
```
QuLogic: This is part of the public API, so please don't change the names, or you'll have to add some deprecations, etc.

lib/cartopy/mpl/gridliner.py
```diff
@@ -90,13 +95,32 @@ def _north_south_formatted(latitude, num_format='g'):
...
+
+
 class Gridliner(object):
     # NOTE: In future, one of these objects will be add-able to a GeoAxes (and
     # maybe even a plain old mpl axes) and it will call the "_draw_gridliner"
     # method on draw. This will enable automatic gridline resolution
     # determination on zoom/pan.
-    def __init__(self, axes, crs, draw_labels=False, xlocator=None,
-                 ylocator=None, collection_kwargs=None):
+    def __init__(self, axes, crs, draw_labels=False, mlocator=None,
```
QuLogic: Same here.

lib/cartopy/mpl/gridliner.py
```diff
@@ -130,37 +160,51 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
 
         #: The :class:`~matplotlib.ticker.Locator` to use for the x
         #: gridlines and labels.
-        self.xlocator = xlocator or degree_locator
+        if mlocator is not None:
+            if not isinstance(mlocator, mticker.Locator):
+                mlocator = mticker.FixedLocator(mlocator)
+                print('fixed lon')
```
QuLogic: Remove.

lib/cartopy/mpl/gridliner.py
```diff
@@ -130,37 +160,51 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
...
+        else:
+            self.mlocator = classic_locator
 
         #: The :class:`~matplotlib.ticker.Locator` to use for the y
         #: gridlines and labels.
-        self.ylocator = ylocator or degree_locator
+        if plocator is not None:
+            if not isinstance(plocator, mticker.Locator):
+                plocator = mticker.FixedLocator(plocator)
+                print('fixed lat')
```
QuLogic: Remove.

lib/cartopy/mpl/gridliner.py
```diff
@@ -130,37 +160,51 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
...
+            self.plocator = plocator
+        elif isinstance(crs, cartopy.crs.PlateCarree):
+            self.plocator = LatitudeLocator()
+        else:
+            self.plocator = classic_locator
 
         #: The :class:`~matplotlib.ticker.Formatter` to use for the x labels.
-        self.xformatter = mticker.ScalarFormatter()
-        self.xformatter.create_dummy_axis()
+        self.mformatter = mformatter or LongitudeFormatter()
```
QuLogic: This is all public API too, so can't change the names.

lib/cartopy/mpl/ticker.py
```diff
@@ -98,13 +192,20 @@ def _hemisphere(self, value, value_source_crs):
 
 class LatitudeFormatter(_PlateCarreeFormatter):
     """Tick formatter for latitude axes."""
-    def __init__(self, degree_symbol=u'\u00B0', number_format='g',
+    def __init__(self,
```
QuLogic: Please do not change the order of arguments on public API.

lib/cartopy/mpl/ticker.py
```diff
@@ -98,13 +192,20 @@ def _hemisphere(self, value, value_source_crs):
...
+                 decimal=False,
+                 minute_symbol=u"'",
+                 second_symbol=u"''",
+                 number_format='g',
+                 seconds_number_format='g',
+                 auto_hide=True,
                  transform_precision=1e-8):
         """
-        Tick formatter for a latitude axis.
+        Tick formatter for a latitudes.
```
QuLogic: Remove 'a'.

lib/cartopy/mpl/ticker.py
```diff
@@ -262,3 +439,65 @@ def _hemisphere(self, value, value_source_crs):
         if value in (-180, 180) and not self._dateline_direction_labels:
             hemisphere = ''
         return hemisphere
+
+
+class LongitudeLocator(MaxNLocator):
+    """A locator for longitude degrees that works even at very small scale
+    on minutes and seconds
```
QuLogic: First line of a docstring should not wrap.

lib/cartopy/mpl/ticker.py
```diff
@@ -262,3 +439,65 @@ def _hemisphere(self, value, value_source_crs):
...
+
+    Parameters
+    ----------
+    decimal: bool
+        Force decimal degrees so the local does not stop specifically on
```
QuLogic: the local?

lib/cartopy/mpl/ticker.py
```diff
@@ -262,3 +439,65 @@ def _hemisphere(self, value, value_source_crs):
...
+        self.set_params(steps=np.array(steps))
+
+    def _raw_ticks(self, vmin, vmax):
+        self._guess_steps(vmin, vmax)
+        return MaxNLocator._raw_ticks(self, vmin, vmax)
+
+
+class LatitudeLocator(LongitudeLocator):
+    """A locator for latitude degrees that works even at very small scale
+    on minutes and seconds
```
QuLogic: First line of a docstring should not wrap.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -93,14 +93,16 @@ def test_gridliner():
...
+    ps = [-90, -60, -30, 0, 30, 60, 90]
     ax = mock.Mock(_gridliners=[], spec=GeoAxes)
-    gl = GeoAxes.gridlines(ax, xlocs=xs, ylocs=ys)
-    assert isinstance(gl.xlocator, mticker.FixedLocator)
-    assert isinstance(gl.ylocator, mticker.FixedLocator)
-    assert gl.xlocator.tick_values(None, None).tolist() == xs
-    assert gl.ylocator.tick_values(None, None).tolist() == ys
+    gl = GeoAxes.gridlines(ax, mlocs=ms, plocs=ps)
+    print(gl.mlocator)
+    print(gl.plocator)
```
QuLogic: Remove.

lib/cartopy/tests/mpl/test_ticker.py
```diff
@@ -127,6 +132,8 @@ def test_LongitudeFormatter_central_longitude_0():
     result = [formatter(tick) for tick in test_ticks]
     expected = [u'180\u00B0W', u'120\u00B0W', u'60\u00B0W', u'0\u00B0',
                 u'60\u00B0E', u'120\u00B0E', u'180\u00B0E']
+    print(result)
+    print(expected)
```
QuLogic: Remove.

lib/cartopy/mpl/gridliner.py
stefraynaud: Ok. 
However, these lines originally in the geoaxes.py module: https://github.com/SciTools/cartopy/blob/696e6df5ca09079e51d22c8c8686ebf7441340b6/lib/cartopy/mpl/geoaxes.py#L1204
Maybe this piece of preprocessing can be directly integrated onto gridliner.py?

QuLogic: I meant the `print`...

stefraynaud:
I'm working on a more universal way to plot labels, not only on plot axes limits, but also everywhere grid lines intersects the map boundary, whatever its shape.
I also added a clip keyword to the `GeoAxes.set_boundary()`, so it will be possible to easily make plots like this one : https://oceancurrents.rsmas.miami.edu/atlantic/img_mgsva/north-atlantic-drift-YYY.gif

So I'll amend this PR in the future.

corinnebosley:
Hi @stefraynaud, I see that you're working on a new PR to achieve this goal a little bit better, but I think this PR along with #1089 are incredibly valuable.  

Can you please let me know what's going on with this one now?  I am happy to review it and to help progress it (because I really want this in cartopy, finally, and you've done a lot of work which I don't want to lose), but if it's dead and there is a replacement in progress then I will wait for that instead.

Also, I noticed that this PR must be merged with #1089; is that because of something functional or something to do with the docs?  What I want to know is; if I borrow this branch to test it locally, will it work?

stefraynaud:
The following code produces the following figures:

```python


from __future__ import (absolute_import, division, print_function)

import matplotlib
import cartopy.crs as ccrs

matplotlib.rc('font', size=6)

pcar = ccrs.PlateCarree()

# Global projections

ncol = 4
sy = 3.5
sx = 3.5
global_projs = [pcar,
              ccrs.Mercator(),
              ccrs.AlbersEqualArea(),
              ccrs.AzimuthalEquidistant(central_longitude=90),
#              ccrs.EquidistantConic(),
              ccrs.LambertConformal(),
              ccrs.LambertCylindrical(),
              ccrs.Miller(),
              ccrs.Mollweide(),
              ccrs.Orthographic(),
              ccrs.Robinson(),
              ccrs.Sinusoidal(),
              ccrs.Stereographic(),
              ccrs.TransverseMercator(),
              ccrs.InterruptedGoodeHomolosine(),
              ccrs.OSGB(),
              ccrs.EuroPP(),
              ccrs.Geostationary(),
              ccrs.NearsidePerspective(central_latitude=50.72,
                                       central_longitude=-3.53,
                                       satellite_height=10000000.0),
#              ccrs.EckertI(),
#              ccrs.EckertIII(),
#              ccrs.EqualEarth(),
              ccrs.Gnomonic(),
              ccrs.LambertAzimuthalEqualArea(),
              ccrs.NorthPolarStereo(),
              ccrs.OSNI(),
              ]

nrow = (len(global_projs)-1)//ncol + 1
fig = plt.figure(figsize=(sx*ncol, nrow * sy))
print('GLOBAL')
for ip, proj in enumerate(global_projs):
    print(' '+proj.__class__.__name__)
    ax = plt.subplot(*(nrow, ncol, ip+1), projection=proj)
    ax.gridlines(pcar, draw_labels=True)
    for feature in [cfeature.OCEAN, cfeature.LAND, cfeature.COASTLINE]:
        ax.add_feature(feature)
    ax.set_title(proj.__class__.__name__ + ' global')
fig.savefig(__file__[:-2]+'global.png')  # to make tight_layout working
fig.tight_layout(h_pad=1.5, w_pad=1.5, pad=0.1)
fig.savefig(__file__[:-2]+'global.png')


# Special cases

ncol = 2
sy = 3
sx = 4
uk_extent = [-10, 3, 49, 59]
north_atlantic_extent = [-90, 0, -10, 70]
special_projs = [(pcar, uk_extent, False),
                 (ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=45.0),
                  north_atlantic_extent, False),
                 (ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=45.0),
                  north_atlantic_extent, True)]

nrow = (len(special_projs)-1)//ncol + 1
fig = plt.figure(figsize=(sx*ncol, nrow * sy))
for ip, (proj, extents, clip) in enumerate(special_projs):
    ax = plt.subplot(*(nrow, ncol, ip+1), projection=proj)
    ax.set_extent(extents, crs=pcar, clip=clip)
    ax.gridlines(pcar, draw_labels=True)
    for feature in [cfeature.OCEAN, cfeature.LAND, cfeature.COASTLINE]:
        ax.add_feature(feature)
    ax.set_title(proj.__class__.__name__ + ' with extent [clip={}]'.format(clip))
fig.savefig(__file__[:-2]+'special.png')  # to make tight_layout working
fig.tight_layout(h_pad=1.5, w_pad=1.5, pad=0.1)
fig.savefig(__file__[:-2]+'special.png')
```
With this update, the position of labels is extrapolated from the intersection of a gridline with the map boundary, wherever it occurs. The visibility of each labels is checked so that the do not overlap themselve and do not overlap the map boundary. When an overlapping occurs, the other label directions are plotted within a given limit. Despite the algorithm can be improved, it should provide a decent basis.

Plots with global projections:
test_labels_all_projs global
Special cases:
test_labels_all_projs special

corinnebosley:
Hi @stefraynaud, I've been taking a look at your work this morning.  Those are some very pretty plots you've produced.  

As you can see by the checks, I couldn't look at any test results because you need to resolve some merge conflicts before Appveyor can run them.  I didn't find any test failures when I ran them on your branch locally though, so that bodes well.

There are a couple of sections in your code additions which I would like some clarification on, but I'll dig into that when we get some tests running.  

Thank you for your continuing effort with this.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -427,10 +427,45 @@ def draw(self, renderer=None, inframe=False):
...
-
+        self._cachedRenderer = renderer
         return matplotlib.axes.Axes.draw(self, renderer=renderer,
                                          inframe=inframe)
 
+    def _update_title_position(self, renderer):
+        """
+        Update the title position based on the bounding box enclosing
+        all the ticklabels and x-axis spine and xlabel...
+
```
corinnebosley: Why the dots here?  Is this docstring unfinished?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -427,10 +427,45 @@ def draw(self, renderer=None, inframe=False):
...
+        """
+        matplotlib.axes.Axes._update_title_position(self, renderer)
+        if not self._gridliners:
+            return
+
+        if self._autotitlepos is not None and not self._autotitlepos:
```
corinnebosley: I don't quite understand what this statement checks for; can you explain please?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -427,10 +427,45 @@ def draw(self, renderer=None, inframe=False):
...
+        # Get the max ymax of all top labels
+        top = -1
+        for gl in self._gridliners:
+            if gl.has_labels():
+                for label in (gl.top_label_artists +
+                              gl.left_label_artists +
+                              gl.right_label_artists):
+                    bb = label.get_tightbbox(renderer)
+                    top = max(top, bb.ymax)
+        if top < 0:
```
corinnebosley: Can you add a short comment about why you are returning without adjustment in the case in which `top < 0` and that in which `yn <= 1` please?  It's not very clear.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -731,6 +768,19 @@ def set_extent(self, extents, crs=None):
             if boundary.equals(domain_in_crs):
                 projected = boundary
 
+        if clip:
+            n = 100
+            x = np.linspace(x1, x2, n)
+            y = np.linspace(y1, y2, n)
+            xx = np.concatenate((x, [x[-1]]*n, x[::-1], [x[0]]*n))
+            yy = np.concatenate(([y[0]]*n, y, [y[-1]]*n, y[::-1]))
```
corinnebosley: Can you please explain what `xx` and `yy` are, and add a comment to the same effect?  There's a lot of information to unpack in those two lines and it's making my eyes go funny taking it apart to work out what these variables are and what they represent.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1229,14 +1280,14 @@ def gridlines(self, crs=None, draw_labels=False, xlocs=None,
...
+            used to determine the locations of the meridian gridlines in the
+            coordinate of the given CRS. Defaults to None, which
             implies automatic locating of the gridlines.
         ylocs: optional
             An iterable of gridline locations or a
             :class:`matplotlib.ticker.Locator` instance which will be
-            used to determine the locations of the gridlines in the
-            y-coordinate of the given CRS. Defaults to None, which
+            used to determine the locations of the parallel gridlines in the
+            coordinate of the given CRS. Defaults to None, which
```
corinnebosley: Good words.  I like it.

lib/cartopy/mpl/geoaxes.py
stefraynaud: This docstring as taken from matplotlib:
https://github.com/matplotlib/matplotlib/blob/529410fa0c9379cd60b1e4d8fdd0478bb2f7e892/lib/matplotlib/axes/_base.py#L2468
Should I remove it or copy the __doc__ attribute?

lib/cartopy/mpl/geoaxes.py
stefraynaud: I also comes from matplotlib:
https://github.com/matplotlib/matplotlib/blob/529410fa0c9379cd60b1e4d8fdd0478bb2f7e892/lib/matplotlib/axes/_base.py#L2471

I just remove the logging call.

lib/cartopy/mpl/gridliner.py
```diff
@@ -90,13 +97,42 @@ def _north_south_formatted(latitude, num_format='g'):
                                            _north_south_formatted(v))
 
 
+def _text_angle_to_specs_(angle):
+    """Get appropriate kwargs for a rotated label from its angle in degrees"""
+    if False and matplotlib.__version__ > '3':
+        # deactivated because does not work properly
+        # with rotation_mode='anchor'
```
corinnebosley: Can you please create an issue in matplotlib to address this, and also create an issue in cartopy referencing the matplotlib issue.  And can you then reference the cartopy issue in this comment (and vice versa) so that we can revert the deactivation here when appropriate.

lib/cartopy/mpl/gridliner.py
```diff
@@ -90,13 +97,42 @@ def _north_south_formatted(latitude, num_format='g'):
...
+        kw.update(ha='right', va=va_align_center)
+        kw['rotation'] -= np.sign(angle) * 180
+        loc = 'left'
+    elif angle > 45:
+        kw.update(ha='center', va='bottom', rotation=angle-90)
+        loc = 'top'
+    else:
+        kw.update(ha='center', va='top', rotation=angle+90)
+        loc = 'bottom'
+    return kw, loc
```
corinnebosley: I'd like to see some clearer separations in this block please.  I can't see anything wrong with it functionally, but it would be clearer if there was a blank line above each variable that is not set within an `if` statement.  It would also make the logic and flow more apparent.

lib/cartopy/mpl/geoaxes.py
stefraynaud: When `top < 0`, the labels are all below the axes upper bound and they basically can't overlap the tile.

lib/cartopy/mpl/gridliner.py
```diff
@@ -121,45 +157,69 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
             to determine the locations of the gridlines in the y-coordinate of
             the given CRS. Defaults to None, which implies automatic locating
             of the gridlines.
+        xformatter: optional
+            A :class:`matplotlib.ticker.Formatter` instance to format
+            longitude labels.
+        yformatter: optional
+            A :class:`matplotlib.ticker.Formatter` instance to format
+            latitude labels.
```
corinnebosley: These descriptions seems a little misleading when we consider L.202 and L.205, which cause the default values to become `LongitudeFormatter` and `LatitudeFormatter` rather than None.  I'd like these docstrings to reflect that behaviour please.

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
-        x_label_points = [x for x in x_ticks if x_lim[0] <= x <= x_lim[1]]
-        y_label_points = [y for y in y_ticks if y_lim[0] <= y <= y_lim[1]]
+        self.bottom_label_artists = []
+        self.top_label_artists = []
+        self.left_label_artists = []
+        self.right_label_artists = []
+        if not (self.left_labels or self.right_labels or
+                self.bottom_labels or self.top_labels):
+            return
+        self._assert_can_draw_ticks()
```
corinnebosley: This assertion is now incorrect here, is it not?

corinnebosley: In the sense that you can now draw labels on projections other than PlateCarree.

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+                        tails = [[(pt.x, pt.y) for pt in intersection[:2]]]
+                        heads = [[(pt.x, pt.y)
+                                  for pt in intersection[-1:-3:-1]]]
+                    elif isinstance(intersection, (sgeom.LineString,
+                                                   sgeom.MultiLineString)):
+                        if isinstance(intersection, sgeom.LineString):
+                            intersection = [intersection]
+                        elif len(intersection) > 4:
+                            # gridline and map boundary are parallel
+                            # and they intersect themselve too musc
```
corinnebosley: should this say: `# and they intersect themselves too much`?

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+                        heads = []
+                        for inter in intersection:
+                            if len(inter.coords) < 2:
+                                continue
+                            tails.append(inter.coords[:2])
+                            heads.append(inter.coords[-1:-3:-1])
+                        if not tails:
+                            continue
+                    else:
+                        # TODO: we should handle GeometryCollection
```
corinnebosley: Can you please put a warning or error in here (maybe a `NotImplementedError`, although a warning may be more user-friendly) explaining to the user that lots of processing isn't being done here and why.

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+    def _segment_to_text_specs(self, pt0, pt1):
+        """Get appropriate kwargs for a label from lon or lat line segment"""
+        x0, y0 = pt0
+        x1, y1 = pt1
+        angle = np.arctan2(y0-y1, x0-x1) * 180 / np.pi
+        kw, loc = self._segment_angle_to_text_specs(angle)
+        return kw, angle, loc
+
+    def _segment_angle_to_text_specs(self, angle):
+        """Get appropriate kwargs for a given direction angle"""
```
corinnebosley: I think that `Get appropriate kwargs for a given text angle` is clearer.

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+        The following rules apply:
+
+        - Labels are plotted and checked by order of priority,
+          with a high priority for longitude labels at the bottom and
+          top of the map, and the reverse for latitude labels.
+        - A label must not overlap another label marked as visible.
+        - A label must not overlap the map boundary.
+        - When a label is about to be hidden, other angles are tried in the
+          absolute given limit of max_delta_angle by increments of delta_angle
+          of difference from the original angle.
```
corinnebosley: This is a beautiful docstring.  Very nice indeed.

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+            return
+        paths = []
+        outline_path = None
+        delta_angle = 22.5
+        max_delta_angle = 45
+        axes_children = self.axes.get_children()
+        for priority, artist in self._labels:
+
+            if artist not in axes_children:
+                warn('The labels if this gridliner do not belongs'
```
corinnebosley: A couple of little typos here; `'The labels if this gridliner do not belongs'` -> `'The labels of this gridliner do not belong'`

lib/cartopy/mpl/gridliner.py
```diff
@@ -327,58 +338,257 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None):
...
+                        break
+
+            # Action
+            artist.set_visible(visible)
+            if not visible:
+                plt.setp(artist, **orig_specs)
+            else:
+                paths.append(this_path)
+                # is this necessary?
+#                artist.update_bbox_position_size(renderer)
```
corinnebosley: I take it this is left in here because you are not sure whether it makes a difference or not.  I am happy to leave it in here unless you have any objections?

lib/cartopy/mpl/ticker.py
```diff
@@ -34,46 +35,138 @@ class _PlateCarreeFormatter(Formatter):
...
-        # Round the transformed value using a given precision for display
-        # purposes. Transforms can introduce minor rounding errors that make
-        # the tick values look bad, these need to be accounted for.
-        f = 1. / self._transform_precision
-        projected_value = round(f * projected_value) / f
+        if self.axis is not None:
+
+            if not isinstance(self.axis.axes, GeoAxes):
+                raise TypeError("This formatter can only be "
+                                "used with cartopy axes.")
```
corinnebosley: Can you please replace `cartopy axes` with `cartopy GeoAxes` for clarity.

lib/cartopy/mpl/ticker.py
```diff
@@ -148,11 +257,24 @@ def __init__(self, degree_symbol=u'\u00B0', number_format='g',
             lat_formatter = LatitudeFormatter(degree_symbol='')
             ax.yaxis.set_major_formatter(lat_formatter)
 
+        When not bounded to an axis::
```
corinnebosley: Can you expand this comment a little please to explain what this does when not bounded to an axis?

Also I'm not sure about the use of the word 'bounded'.  The SciTools Iris package  is very often used with Cartopy, and this has a different meaning of the word 'bounded', so this may be a little confusing.  

I think it might be alright if we just use 'bound' instead, and I can't think of a better word that means the same thing.

lib/cartopy/mpl/ticker.py
```diff
@@ -175,12 +297,18 @@ def __init__(self,
...
+                 seconds_number_format='g',
+                 auto_hide=True,
+                 ):
         """
-        Create a formatter for longitude values.
+        Create a formatter for longitudes.
 
-        The axis must be part of an axes defined on a rectangular
-        projection (e.g. Plate Carree, Mercator).
+        When bounded to an axis, the axis must be part of an axes defined
```
corinnebosley: `bounded` -> `bound`

lib/cartopy/mpl/ticker.py
```diff
@@ -198,13 +326,25 @@ def __init__(self,
             The symbol used to represent degrees. Defaults to u'\u00B0'
             which is the unicode degree symbol.
         number_format: optional
-            Format string to represent the longitude values. Defaults to
-            'g'.
+            Format string to represent the degrees tick values.
```
corinnebosley: I find this wording a bit confusing, from a user's perspective.  I prefer `Format string to represent the longitude values`.

lib/cartopy/mpl/ticker.py
```diff
@@ -198,13 +326,25 @@ def __init__(self,
...
+            degrees-minutes-seconds.
+        minute_symbol: str, optional
+            The character(s) used to represent the minute symbol.
+        second_symbol: str, optional
+            The character(s) used to represent the second symbol.
+        seconds_number_format: optional
+            Format string to represent the seconds tick values.
+            Defaults to 'g'.
+        auto_hide: bool, optional
+            Auto-hide degrees or minutes when redondant.
```
corinnebosley: `redondant` -> `redundant`

lib/cartopy/mpl/ticker.py
```diff
@@ -233,18 +373,58 @@ def __init__(self,
             lon_formatter = LongitudeFormatter()
             ax.xaxis.set_major_formatter(lon_formatter)
 
+
+        When not bounded to an axis::
```
corinnebosley: Same again please, a bit more detail about what this does, and `bounded` -> `bound`.

lib/cartopy/mpl/ticker.py
```diff
@@ -99,12 +192,15 @@ def _hemisphere(self, value, value_source_crs):
 class LatitudeFormatter(_PlateCarreeFormatter):
     """Tick formatter for latitude axes."""
     def __init__(self, degree_symbol=u'\u00B0', number_format='g',
-                 transform_precision=1e-8):
+                 transform_precision=1e-8, decimal=False,
```
corinnebosley: I'm concerned that the kwarg `decimal=False` introduces a breaking change.  In your tests for these Formatters you have added an extra kwarg `decimal=True` to test that the result remains the same as it was previously.

In reality, that means that if we keep the syntax of those tests the same, the result will now be different, which is a breaking change.  That means that we can only release these changes for v1.0.0, which is the next major release of Cartopy.

To address this, you can either flip the default value of the `decimal` kwarg so that the behaviour is the same by default, or you can create a v1.0.x branch and target that instead of master.  Your call.

lib/cartopy/mpl/ticker.py
```diff
@@ -175,12 +297,18 @@ def __init__(self,
                  dateline_direction_label=False,
                  degree_symbol=u'\u00B0',
                  number_format='g',
-                 transform_precision=1e-8):
+                 transform_precision=1e-8,
+                 decimal=False,
```
corinnebosley: Same case as above; this is a breaking change according to the tests, so you must address it by either flipping the default `decimal` value or targetting a v1.0.0 branch.

lib/cartopy/tests/mpl/test_ticker.py
```diff
@@ -84,7 +89,7 @@ def test_LatitudeFormatter_degree_symbol():
 
 
 def test_LatitudeFormatter_number_format():
-    formatter = LatitudeFormatter(number_format='.2f')
+    formatter = LatitudeFormatter(number_format='.2f', decimal=True)
```
corinnebosley: This and similar tests may have to be reverted depending on action taken re: breaking changes to Longitude and Latitude Formatters.

lib/cartopy/mpl/geoaxes.py
corinnebosley: OK, fair enough.  I still don't understand, but I trust you.

lib/cartopy/mpl/geoaxes.py
corinnebosley: That makes sense.  A little comment in the code to that effect would be great.

stefraynaud:
Thanks for the review @corinnebosley 

I'm not sure I will be able finalize it before the end week, and thus before january.

lib/cartopy/mpl/ticker.py
stefraynaud: I think it is important that `decimal=False`  by default for everybody working at small areas. However, it remains less important than the capability of setting it manually.
So, is the good solution to let it at `True` by default until v1.0.0, then switch to `False`?

This highlights another need: to extend the list of keyword accepted by `Gridliner.__init__` for tuning  everything. In this particular case, a keyword like `xformatter_kwargs` would be useful to tune the default formatter at creation time: `xformatter_kwargs:{'decimal': False}`. By extension, all the tuning that is accessible by changing attributes like `xlines`or `xpadding` should be accessible from the keywords. Should we add the keywords and in which order?

And by the way, v1.0.0 must be the occasion to replace `x/y` by `m/p` when appropriate since X/Y no longer necessarily means longitude/latitude but meridian/parallel.

corinnebosley:
@stefraynaud I did think that might be the case (especially looking back on the review).  There's no hurry though, and I will look at it when I can.

stefraynaud:
A question: what will be the strategy for merging with PR #1089? #1117 and #1089 are complementary. However, the latter is now based on a gridliner core that has been fully changed in the current PR.

corinnebosley:
@stefraynaud It might be a good idea to create a feature branch to merge #1117 and # 1089 into together, then merge that branch back into master once everything works nicely together.

dopplershift:
I think it would be nicer to get this one merged and then do the work separately to rework #1089 on top of this one. Unless there's reason to believe that changes in this PR could make implementing the functionality of #1089 harder?

lib/cartopy/mpl/gridliner.py
stefraynaud: I simply updated the required mpl version to 3.1.

lib/cartopy/mpl/ticker.py
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: Fixed

lib/cartopy/mpl/geoaxes.py
stefraynaud: Fixed

lib/cartopy/mpl/ticker.py
```diff
@@ -34,46 +35,138 @@ class _PlateCarreeFormatter(Formatter):
...
+
+        if mn or (not self._auto_hide_minutes and label):
+            label = self._format_minutes(mn) + label
+
+        if not self._auto_hide_degrees or not label:
+            label = self._format_degrees(deg) + hemisphere + label
+
+        return label
+
+    def _get_dms(self, x):
```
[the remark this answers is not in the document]
stefraynaud: Fixed

lib/cartopy/mpl/ticker.py
stefraynaud: I reformulated the description of both  `number_format` and `seconds_number_format`.

lib/cartopy/mpl/ticker.py
stefraynaud: Fixed

lib/cartopy/mpl/ticker.py
stefraynaud: Fixed

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -92,19 +93,25 @@ def test_gridliner():
...
-    assert gl.xlocator.tick_values(None, None).tolist() == xs
-    assert gl.ylocator.tick_values(None, None).tolist() == ys
+    assert gl.xlocator.tick_values(None, None).tolist() == meridians
+    assert gl.ylocator.tick_values(None, None).tolist() == parallels
 
 
 # The tolerance on this test is particularly high because of the high number
 # of text objects. A new testing strategy is needed for this kind of test.
-if MPL_VERSION >= '2.0':
+if MPL_VERSION >= '2':
```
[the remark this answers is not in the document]
stefraynaud: Simpler is better, but ok, it's irrelevent. I remove it.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -113,11 +120,11 @@ def test_gridliner_specified_lines():
 @pytest.mark.natural_earth
 @ImageTesting([grid_label_image])
 def test_grid_labels():
-    fig = plt.figure(figsize=(8, 10))
+
```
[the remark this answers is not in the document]
stefraynaud: Which white space must be removed?

lib/cartopy/mpl/ticker.py
```diff
@@ -115,12 +211,25 @@ def __init__(self, degree_symbol=u'\u00B0', number_format='g',
...
+            degrees-minutes-seconds.
+        minute_symbol: str, optional
+            The character(s) used to represent the minute symbol.
+        second_symbol: str, optional
+            The character(s) used to represent the second symbol.
+        seconds_number_format: optional
+            Format string to represent the seconds tick values.
+            Defaults to 'g'.
+        auto_hide: bool, optional
+            Auto-hide degrees or minutes when redondant.
```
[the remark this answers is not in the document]
stefraynaud: Fixed

lib/cartopy/mpl/gridliner.py
stefraynaud: I handled the GeometryCollection base.
I'm not sure of other cases may occurs, so I added a warning if one of them happens.
However, a warning may not be appropriate if the user cannot do anything, and raising an error is probably too much, since it will simply not plot some labels, which occurs often for other reasons.

lib/cartopy/mpl/ticker.py
stefraynaud: Set to True by default (until v1.0).

stefraynaud:
Same error everywhere that seems independant from commits (here from travis):
```bash
$ python -c "import cartopy; print('Version ', cartopy.__version__)" && python setup.py version
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/cartopy/__init__.py", line 107, in <module>
    import cartopy.crs
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/cartopy/crs.py", line 36, in <module>
    from cartopy._crs import (CRS, Geodetic, Globe, PROJ4_VERSION,
  File "__init__.pxd", line 918, in init cartopy._crs
ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
The command "python -c "import cartopy; print('Version ', cartopy.__version__)" && python setup.py version" failed and exited with 1 during .
```

Where does it come from?

corinnebosley:
@stefraynaud I'm not sure what's going on here regarding your Travis fails, could you please rebase while I review your changes?

lib/cartopy/tests/mpl/test_gridliner.py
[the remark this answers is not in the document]
corinnebosley: Just L.123.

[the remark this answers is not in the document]
corinnebosley: Although if this is the only change left to make before merge then I'll merge anyway because it's non-functional.

corinnebosley:
@stefraynaud Having re-reviewed following recent changes, I have posted a comment regarding the kwarg `decimal`, but everything else looks good to me and I'm happy with it.

Except of course all the CI failing, which I intend to try and recreate now to see if I can find the problem.

I'll check back shortly and report on any progress.

lib/cartopy/mpl/ticker.py
stefraynaud: In the examples you make a reference to, degrees are displayed as fractions of degree, despite values are plain degrees with a fraction equals to zero. This means that for a smaller area, labels would be like ``35.5°W``. If we want to keep this behavior by default, decimal must be set to True in the formatter classes (and locators) since ``decimal=False`` means no minutes/seconds.
In one of the last commits, I added a decimal kwargs also to ``add_gridliner`` and ``Gridliner`` to easily change this important behavior.

corinnebosley: Actually in those examples, the top one shows the default behaviour of gridline labels (as demonstrated by not using the kwarg `decimal` at all) as not decimal.

The lower example has a decimal point but this is spec ified by the kwarg `number_format` (which does make me wonder which would override if both `decimal=False` and `number_format=.1f` were used).

corinnebosley: And now that I think about that a little more, I think it is a bad idea to be able to set 2 kwargs which can contradict each other; this is asking for trouble.

I understand the requirement to be able to use whole numbers only in Latitude and Longitude Formatters, but we can get all the control we need over it with the kwarg `number_format` can't we?

In which case, I would suggest removing `decimal` completely.

Thoughts?

stefraynaud: The ``number_format`` is simply ignored when ``decimal=False``.
By letting the decimal keyword, we allow people working in small area to access the hability to have minutes and seconds in their labels without too much effort.

In think there is no contradiction. However, it may be clearer to use another keyword name that decimal.

corinnebosley: You're right, I misunderstood the output format when `decimal=True`.

To me decimal implies a number divided into 10, whereas you need to describe a number divided into 60.  How about a boolean arg for `minutes`?

stefraynaud: Yep. ``minutes`` is probably less confusing than ``decimal``. So, when ``minutes=True``, the locators will be able to fall on fractions of 60, the formatter will display minutes and seconds when appropriate, and ``number_format`` will be deactivated.
So I rename it everywhere?

corinnebosley: Yes please, and set it to `False` as default so that we don't change the default behaviour. Are you happy with that?

corinnebosley:
@stefraynaud Travis CI errors look like they may be a result of some upstream problem, although I haven't yet identified what.  Certainly not fixable in this PR, but may have to rebase again before merge.  Not sure.

lib/cartopy/mpl/ticker.py
QuLogic: But what is does is produce [Decimal degrees](https://en.wikipedia.org/wiki/Decimal_degrees). `minutes` is not right because it can also produces seconds.

corinnebosley: I quite like `dms`.  I'd be happy with that.

QuLogic:
I think I need to experiment with the `Gridliner` interface a bit; I'm not sure if some of the documentation changes here might be making a few assumptions that aren't true.

lib/cartopy/mpl/ticker.py
```diff
@@ -93,7 +93,7 @@ def __call__(self, value, pos=None):
     def _format_value(self, value, original_value):
         hemisphere = self._hemisphere(value, original_value)
 
-        if self._decimal:
+        if not self._minutes:
```
corinnebosley: Why has this gained a `not`?

lib/cartopy/mpl/ticker.py
stefraynaud: I also agree. So minutes->dms?

corinnebosley: Yes please!

lib/cartopy/mpl/gridliner.py
```diff
@@ -121,45 +173,75 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
             to determine the locations of the gridlines in the y-coordinate of
             the given CRS. Defaults to None, which implies automatic locating
             of the gridlines.
+        xformatter: optional
+            A :class:`matplotlib.ticker.Formatter` instance to format
+            longitude labels. Defaults to
+            :class:`cartopy.mpl.ticker.LongitudeFormatter`.
```
corinnebosley: Surely this defaults to `None`, does it not?

ahuang11:
Hi, was just curious if this is still planned for cartopy 0.18?

corinnebosley:
@ahuang11 If all goes well then we'll be releasing version 0.18 tomorrow, so it is unlikely that @stefraynaud will get the final changes in with time to be reviewed by then.

But if this PR can be merged without any breaking change3s then we will probably put it into 0.19 quite soon.

stefraynaud:
@corinnebosley I have to recheck but I think everything have been commited.
We also have to wait for @QuLogic that want to have a better look at the code and test it.

corinnebosley:
@stefraynaud There was one thing I was still unsure about, which is in my [last code review comment ](https://github.com/SciTools/cartopy/pull/1117#discussion_r247905062), about the wording in the docstring for `gridliner.py`.  

Although it's a very minor detail, could you please clarify for me that the xformatter keyword argument definitely defaults to `cartopy.mpl.ticker.LongitudeFormatter` when not specified.  

Once I am happy that your wording is correct there, we just have to wait for approval from @QuLogic and then we're good to go.

lib/cartopy/mpl/gridliner.py
stefraynaud: Clarified

allenea:
I just tried to recreate both (9/20 & 12/10)of @stefraynaud code above using cartopy 0.17.0. I added one print statement in the first example to show the proj and extent before the savefig.

**Example 1 (9/20/2018) **
<cartopy.crs.PlateCarree object at 0xb24cac1a8> [-10, 3, 49, 59]
<cartopy.crs.Mercator object at 0xb245d00a0> [-10, 3, 49, 59]

  File "/Users/allenea/Documents/Eric_Allen/Python_Programs/GEOL_Spatial_Stats/gridlines.py", line 40, in <module>
    ax.gridlines(color='lightgrey', linestyle='-', draw_labels=True)

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/geoaxes.py", line 1224, in gridlines
    ylocator=ylocs, collection_kwargs=kwargs)

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/gridliner.py", line 185, in __init__
    self._assert_can_draw_ticks()

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/gridliner.py", line 399, in _assert_can_draw_ticks
    'supported.'.format(prj=self.axes.projection))

TypeError: Cannot label gridlines on a RotatedPole plot.  Only PlateCarree and Mercator plots are currently supported.



**Example 2 (12/10/2018) **

  File "/Users/allenea/Documents/Eric_Allen/Python_Programs/GEOL_Spatial_Stats/gridlines.py", line 132, in <module>
    ax.gridlines(pcar, draw_labels=True)

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/geoaxes.py", line 1224, in gridlines
    ylocator=ylocs, collection_kwargs=kwargs)

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/gridliner.py", line 185, in __init__
    self._assert_can_draw_ticks()

  File "/Users/allenea/anaconda3/lib/python3.6/site-packages/cartopy/mpl/gridliner.py", line 399, in _assert_can_draw_ticks
    'supported.'.format(prj=self.axes.projection))

TypeError: Cannot label gridlines on a AlbersEqualArea plot.  Only PlateCarree and Mercator plots are currently supported.

Can you offer some insight as to why this might be happening?

stefraynaud:
@allenea these features are not available in cartopy 0.17.0.
To test them, you must install cartopy from the new_labbeling branch of my fork:
https://github.com/stefraynaud/cartopy/tree/new_labelling

allenea:
> @allenea these features are not available in cartopy 0.17.0.
> To test them, you must install cartopy from the new_labbeling branch of my fork:
> https://github.com/stefraynaud/cartopy/tree/new_labelling

Okay. I'll take a look. If I'm following this correctly.... The feature won't be ready until 0.19?

You mentioned earlier "I'm working on a more universal way to plot labels, not only on plot axes limits, but also everywhere grid lines intersects the map boundary, whatever its shape."... Does this mean something similar to coastlines where regardless of the projection it'll accurately place the gridlines?

Thanks, E

stefraynaud:
@allenea This means that labels  are plotted not only along the four sides of rectangular axes, like on the first figure, but also when map boundary is not rectangular, like on the last figure.

lib/cartopy/mpl/gridliner.py
```diff
@@ -175,12 +175,14 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
...
+            :class:`cartopy.mpl.ticker.LongitudeFormatter` initiated
+            with the ``dms`` argument.
         yformatter: optional
             A :class:`matplotlib.ticker.Formatter` instance to format
-            latitude labels. Defaults to
-            :class:`cartopy.mpl.ticker.LatitudeFormatter`.
+            latitude labels. It defaults to None, which implies the use of a
+            :class:`cartopy.mpl.ticker.LatitudeFormatter` initiated
+            with the ``dms`` argument.
         collection_kwargs: optional
```
corinnebosley: Beautiful @stefraynaud.  Thanks.

pp-mo:
:+1: just awesome !!

crazyapril:
Any plans for a new release? This PR clears the ways for many developers to replace Basemap which will be EOL'ed just several months later. I think this PR alone can make a new release.

stefraynaud:
I didn't know about the [curvilinear grid support](https://matplotlib.org/api/toolkits/index.html#matplotlib-axisartist-toolkit) of matplotlib tick labelling. Maybe it should be appropriately used for the cartopy labelling system. However, some adaptation would be needed to support labels placed around the edges of the map that do not coincide with the axes edges.

lukelbd:
This thread hasn't been referenced in a while. Just curious, since version 0.18 has been delayed for a while and this is merged into master, will it be included in the version 0.18 release instead of version 0.19 like discussed above?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -693,7 +733,7 @@ def _get_extent_geom(self, crs=None):
 
         return geom_in_crs
 
-    def set_extent(self, extents, crs=None):
+    def set_extent(self, extents, crs=None, clip=False):
```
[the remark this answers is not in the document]
stefraynaud: You are right. It allows to adapt the map boundaries to specified lat/lon box, so you can make plots such as this one : https://github.com/SciTools/cartopy/pull/1117#issuecomment-444163733

PR #1117 allows to add lon/lat labels even when map boundaries are not rectangular, like with clip=True if the projection is not rectangular.

Maybe this piece of code should be removed from master, and readded with another PR with a valid documentation.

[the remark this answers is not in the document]
QuLogic: But non-rectangular boundaries can be done with `set_boundary`, right?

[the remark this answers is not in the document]
stefraynaud: Setting `clip=True ` makes two things
* compute de boundary according to the extent (here is the added value)
* apply the boundary set `set_boundary`.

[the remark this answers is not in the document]
QuLogic: Seems valuable, but I think it really should go through its own PR with documentation, examples, etc.

zzheng93:
Hello, I installed the latest version 
`conda install -c conda-forge/label/dev cartopy`

But I still got the error:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-a98763482089> in <module>
     25     ax.set_extent(extent, crs=plain_crs)
     26     ax.coastlines(resolution='50m', color='k')
---> 27     ax.gridlines(color='lightgrey', linestyle='-', draw_labels=True)
     28 
     29 plt.show()

/opt/anaconda3/envs/only_plot/lib/python3.7/site-packages/cartopy/mpl/geoaxes.py in gridlines(self, crs, draw_labels, xlocs, ylocs, **kwargs)
   1222         gl = Gridliner(
   1223             self, crs=crs, draw_labels=draw_labels, xlocator=xlocs,
-> 1224             ylocator=ylocs, collection_kwargs=kwargs)
   1225         self._gridliners.append(gl)
   1226         return gl

/opt/anaconda3/envs/only_plot/lib/python3.7/site-packages/cartopy/mpl/gridliner.py in __init__(self, axes, crs, draw_labels, xlocator, ylocator, collection_kwargs)
    183         # public attributes are changed after instantiation.
    184         if draw_labels:
--> 185             self._assert_can_draw_ticks()
    186 
    187         #: The number of interpolation points which are used to draw the

/opt/anaconda3/envs/only_plot/lib/python3.7/site-packages/cartopy/mpl/gridliner.py in _assert_can_draw_ticks(self)
    397                             '{prj.__class__.__name__} plot.  Only PlateCarree'
    398                             ' and Mercator plots are currently '
--> 399                             'supported.'.format(prj=self.axes.projection))
    400         return True
    401 

TypeError: Cannot label gridlines on a RotatedPole plot.  Only PlateCarree and Mercator plots are currently supported.
```

I am wondering if I did something wrong?

lukelbd:
Unfortunately it appears this feature has not yet been published in a release :/. See the [version 0.18 milestone](https://github.com/SciTools/cartopy/milestone/25).

QuLogic:
There is no 0.18.0b1 on https://anaconda.org/conda-forge/cartopy/labels, so you need to wait until they update.

zzheng93:
Hi @lukelbd  @QuLogic, thank you very much for your kind help!
I downloaded this repo, then run
`python setup.py install`
`pip install Cartopy==0.18.0b1`
Now it works.

philippemiron:
I'm on v0.18.0 and getting:
`set_extent() got an unexpected keyword argument 'clip'`

is it suppose to be supported?

QuLogic:
`clip` was not added, and is not an important part of this PR.

philippemiron:
Do you know of any example how to manually add labels after using `ax.set_boundary()`? Thanks.

e.g.
Screen Shot 2020-10-02 at 16 00 42

PR #1125, opened 2018-09-25
Restrict to not version 3 of Matplotlib.

tomgreen66:
## Rationale

Due to conflicts with API changes in Matplotlib 3, make sure we only use compatible versions of Matplotlib.

## Implications

Fix issues with using some features such as contour plotting that do not work with Matplotlib 3.


QuLogic:
Thanks for the PR; I'm not going to merge this because I'm working on fixing it instead.

PR #1129, opened 2018-09-25
TST: Speed up travis a bit

QuLogic:
## Rationale

Currently, Travis takes approximately 250s (4m10s) to install conda, cartopy, and its dependencies, 153s (2.5m) to run tests, and 5s for other bits.
* `conda update conda` does nothing since we pull the latest miniconda, so it wastes 16s
* running tests in parallel locally gets me 120s-> 40/50s, or 2.5-3 times faster. I'm hoping we can shave a similar amount of time of the builds here.

## Implications

Hopefully, this will get builds a bit closer to 5m.


QuLogic:
Because we re-use filenames for tests, I added a lock using `filelock` (since it's used by Iris already) when doing I/O for the image tests. This prevents one process from overwriting the file a different process is trying to read.

PR #1130, opened 2018-09-25
Fix support for Matplotlib 3.0.0

QuLogic:
This fixes one straightforward issue wrt Matplotlib 3.0.0 compatibility. Two issues still exist in Matplotlib (cf matplotlib/matplotlib#12288), so I don't think we will be able to support 3.0.0, but 3.0.1 should be okay.

Once those issues are fixed, there are only 2 or 3 image failures that I need to go through to see if they're okay.


QuLogic:
This is now okay with the `v3.0.x` branch of Matplotlib, with a slight patch to fix the above issues. So when 3.0.1 is out and added to the conda-forge testing label, this should be good to go.

QuLogic:
Hmm, these new images should probably use the new default style since they will only be used by mpl>=2. I can add that, but we probably want to squash to remove the extra images.

.travis.yml
```diff
@@ -11,10 +11,10 @@ env:
         # New FreeType causes changes in text output.
         - NAME="Latest everything."
           PYTHON_VERSION=3.6
-          PACKAGES="numpy matplotlib freetype<2.8 scipy"
+          PACKAGES="numpy matplotlib scipy"
```
corinnebosley: How come no freetype?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1449,19 +1449,19 @@ def pcolormesh(self, *args, **kwargs):
...
-        A temporary, modified duplicate of
-        :func:`~matplotlib.pyplot.pcolormesh'.
+        A modified duplicate of :func:`~matplotlib.pyplot.pcolormesh'.
 
-        This function contains a workaround for a Matplotlib issue
-        and will be removed once the issue has been resolved.
-        https://github.com/matplotlib/matplotlib/pull/1314
+        This function contains patches for Cartopy-specific behaviour, such as
+        using the transform for limit checks, applying longitude wrapping, etc.
+        See PATCH comments below.
```
corinnebosley: Nice detail.  It is appreciated.

.travis.yml
QuLogic: The Matplotlib from the testing label uses a fixed copy of FreeType bundled into Matplotlib, so it will never change (at least until Matplotlib rebuilds all its images with a new version, which is unlikely unless some new comparison method is used.)

.travis.yml
```diff
@@ -46,6 +46,7 @@ install:
   # Customise the testing environment
   # ---------------------------------
   - conda config --add channels conda-forge
+  - conda config --add channels conda-forge/label/testing
```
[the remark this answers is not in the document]
QuLogic: See https://github.com/conda-forge/matplotlib-feedstock/issues/110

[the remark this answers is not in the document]
corinnebosley: I see.  That all makes a bit more sense now.  Well This looks good to me, I guess we are just waiting for mpl 3.0.1.  

Thanks for this @QuLogic, ping me when this can be merged and I'll press the button.

dopplershift:
@QuLogic So this is waiting on me to get the matplotlib 3.0.1 conda-forge testing packages up?

QuLogic:
Yep; let me know when it's done and I'll push a rebase and some fixes.

dopplershift:
@QuLogic Packages should be up.

QuLogic:
@corinnebosley @dopplershift OK, all set; we're just failing on #1171 now.

dopplershift:
I'll merge over CI and we can fix #1171 elsewhere.

PR #1134, opened 2018-09-27
WIP: Buffer exterior rings when creating polygons.

QuLogic:
## Rationale

This is already done for interior rings, and prevents invalid geometries from crashing things later.

## Implications

WIP, since I need to write a test and I might try some benchmarking.


QuLogic:
I've added a test, pared down from the oceans feature to the limit of what's needed to trigger the error.

I tried run the tests a few times on `master` vs the branch and don't see any appreciable change in runtimes.

dopplershift:
Test failure looks real, but I can't readily tell if it's because of your change or upstream.

QuLogic:
Well, it didn't fail for me in testing, but I didn't try the older versions...

QuLogic:
OK, it does fail with 2.7 locally, so I will try and investigate.

QuLogic:
Oops, apparently I missed the obvious that I was not running the same versions of Proj!

QuLogic:
IIRC, this is difficult to get working for both versions of Proj, since they return different values. I think this can't move forward until we bump minimum Proj requirements.

QuLogic:
TBH, I'm not really sure about this; as you can see tests fail and I'm not sure whether this can be corrected.

greglucas:
This looks like it passes all tests, and I do not see a speed regression. Seems good to go after a rebase?

This commit
```bash
[ 50.00%] ··· ============================ ============= ============
              --                                   resolution        
              ---------------------------- --------------------------
                       projection               110m         50m     
              ============================ ============= ============
                      PlateCarree            24.5±0.3ms   2.96±0.02s 
                    NorthPolarStereo        24.6±0.08ms    3.12±0s   
                        Robinson             10.9±0.1ms   3.43±0.02s 
               InterruptedGoodeHomolosine   11.1±0.07ms   3.72±0.01s 
              ============================ ============= ============
```

Current master
```bash
[100.00%] ··· ============================ ============= ============
              --                                   resolution        
              ---------------------------- --------------------------
                       projection               110m         50m     
              ============================ ============= ============
                      PlateCarree            25.8±0.6ms   2.90±0.01s 
                    NorthPolarStereo         25.4±0.4ms   3.07±0.01s 
                        Robinson             11.1±0.2ms   3.34±0.01s 
               InterruptedGoodeHomolosine   11.1±0.06ms   3.61±0.01s 
              ============================ ============= ============
```

PR #1136, opened 2018-09-30
Deprecate clip_path.

QuLogic:
It is just a thin wrapper around a Matplotlib method at this point, and the old compatibility function doesn't even work properly (it uses xrange). We also don't support a version of Matplotlib old enough to require use of the compatibility function now.

### DIFF
--- a/lib/cartopy/mpl/clip_path.py
+++ b/lib/cartopy/mpl/clip_path.py
@@ -17,69 +17,10 @@
 
 from __future__ import (absolute_import, division, print_function)
 
+import warnings
+
 import matplotlib.path as mpath
 import numpy as np
-
-
-def clip_path_python(subject, clip, point_inside_clip_path):
-    """
-    Clip the subject path with the given clip path using the
-    Sutherland-Hodgman polygon clipping algorithm.
-
-    Parameters
-    ----------
-    subject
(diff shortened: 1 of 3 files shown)


lib/cartopy/mpl/clip_path.py
```diff
@@ -113,25 +54,14 @@ def intersection_point(p0, p1, p2, p3):
...
-        bbox_center = ((clip_bbox.x0 + clip_bbox.x1) / 2,
-                       (clip_bbox.y0 + clip_bbox.y1) / 2)
-        return clip_path_python(subject, bbox_patch, bbox_center)
+def clip_path(subject, clip_bbox):
+    """
+    Clip the given path to the given bounding box.
+
+    """
+    warnings.warn("This method has been deprecated. "
+                  "See the \"What's new\" section for v0.17.")
```
ajdawson: I'd prefer the warning to contain the core information from the what's new so cross referencing is not necessary:

```
You can replace ``clip_path(subject, clip_bbox)`` by ``subject.clip_to_bbox(clip_bbox)``)
```

QuLogic: Done (though for some reason GitHub's UI still shows the old warning.)

PR #1137, opened 2018-10-01
Fix some minor typos and style issues

QuLogic:
Just a few things found with codespell.

### DIFF
--- a/docs/make_projection.py
+++ b/docs/make_projection.py
@@ -43,9 +43,9 @@ def plate_carree_plot():
 
     for i in range(0, nplots):
         central_longitude = 0 if i == 0 else 180
-        ax = fig.add_subplot(nplots, 1, i+1,
-                             projection=ccrs.PlateCarree(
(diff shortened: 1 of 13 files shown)


QuLogic:
Also fixed a few flake8 things and some invalid escape sequences.

ajdawson:
Should the changes to `versioneer.py` and `_version.py` be made in versioneer itself and a new version deployed for cartopy? The changes you made are small so I don't mind changing the generated files in this case, but it might not be a sustainable path.

QuLogic:
I tried re-generating from the latest versioneer, but the version we have is already a patched one from @pelson, so it makes for a much larger change, and doesn't support the version scheme we currently have set.

ajdawson:
>I tried re-generating from the latest versioneer, but the version we have is already a patched one from @pelson, so it makes for a much larger change, and doesn't support the version scheme we currently have set.

Something for another time then.

Issue #1080, opened 2018-06-19
SRTM path in USGS server changed


odermatt:
My SRTMDownloader can no longer find SRTM data in the usual location. Please change 

```
_SRTM_BASE_URL = ('https://e4ftl01.cr.usgs.gov/SRTM/SRTMGL{resolution}.'
                      '003/2000.02.11/')
```
to
```
_SRTM_BASE_URL = ('https://e4ftl01.cr.usgs.gov/MEASURES/SRTMGL{resolution}.'
                      '003/2000.02.11/')
```

PR #1145, opened 2018-10-12
TST: Add a fixture enabling SRTM login using envvars.

QuLogic:
## Rationale

There is currently no way to run SRTM tests, so they tend to fall into disrepair (cf #1080).

## Implications

If you set `SRTM_USERNAME` and `SRTM_PASSWORD` in the environment, you can run pytest and have the SRTM tests work. In the future, we may want to [define an encrypted variable on Travis](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml) to enable running these on CI. I don't want to do this myself as I'd rather someone from the Met Office handle the credentials.

### DIFF
--- a/lib/cartopy/io/srtm.py
+++ b/lib/cartopy/io/srtm.py
@@ -413,11 +413,12 @@ class SRTMDownloader(Downloader):
         Downloader.__init__(self, None,
                             target_path_template,
                             pre_downloaded_path_template)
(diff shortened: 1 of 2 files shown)


QuLogic:
Note, you will need to try this out locally since we don't have the envvars set on any CI.

PR #1150, opened 2018-10-20
Use pykdtree when available

QuLogic:
## Rationale

The pykdtree project by @storpipfugl is super fast compared to scipy, so this tries to use it if it's available.

## Implications

There is no effect on tests, but a great effect on the examples:
```
                   example   function     master   pykdtree     ratio
0          aurora_forecast    imshow0   1.614568   0.697115  0.431765
1          aurora_forecast    imshow1   1.620853   0.704454  0.434619
2          aurora_forecast       draw   0.916317   1.119499  1.221738
3        eccentric_ellipse     imshow   2.400456   1.269291  0.528771
4        eccentric_ellipse       draw   0.341589   0.677258  1.982666
5   effects_of_the_ellipse  add_image   0.000008   0.000008  1.031250
6   effects_of_the_ellipse       draw   1.835607   1.980497  1.078933
7             eyja_volcano  add_image   0.000003   0.000003  0.928571
8             eyja_volcano       draw   3.526339  10.100620  2.864336
9            geostationary     imshow  96.402943   1.546087  0.016038
10           geostationary       draw   0.643935   0.896134  1.391653
11                     wms    add_wms   0.514387   0.521981  1.014763
12                     wms       draw   2.045766   2.763029  1.350609
```
In almost all cases, the `imshow` is approximately half the time while `draw` is slightly slower. For some reason, the draw in `eyja_volcano` is much slower which I need to investigate a bit, but the real win here is `geostationary` which is two orders of magnitude faster.

This speedup should hopefully improve doc builds a bit (at least where it was added.)

### DIFF
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -23,7 +23,12 @@ transformations.
 from __future__ import (absolute_import, division, print_function)
 
 import numpy as np
-import scipy.spatial
+try:
+    import pykdtree.kdtree
+    _is_pykdtree = True
+except ImportError:
+    import scipy.spatial
+    _is_pykdtree = False
 
 import cartopy.crs as ccrs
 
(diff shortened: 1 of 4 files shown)


lib/cartopy/img_transform.py
```diff
@@ -270,15 +275,20 @@ def regrid(array, source_x_coords, source_y_coords, source_cs, target_proj,
...
+        _, indices = kdtree.query(target_xyz, k=1, sqr_dists=True)
+    else:
+        # Versions of scipy >= v0.16 added the balanced_tree argument,
+        # which caused the KDTree to hang with this input.
+        try:
+            kdtree = scipy.spatial.cKDTree(xyz, balanced_tree=False)
+        except TypeError:
+            kdtree = scipy.spatial.cKDTree(xyz)
+        _, indices = kdtree.query(target_xyz, k=1)
+    mask = indices >= len(xyz)
```
QuLogic: This change is there because we don't really care about the distance and pykdtree returns differing results for out-of-domain vs. nan https://github.com/storpipfugl/pykdtree/issues/35

QuLogic:
Rebased; I enabled pykdtree on both doc builds because a) it's so much faster, and b) there's already checks for plain scipy on Travis.

ajdawson:
I don't see any reason why we wouldn't want to do this, so 👍

PR #1167, opened 2018-11-03
Normalize and (slightly) speedup boundary and limit calculations

QuLogic:
## Rationale

Projections are currently inconsistent in how they calculate the boundary and projection limits. Some projections calculate the boundary a point at a time and this is slow (though only on the order of tenths of milliseconds). Private variables are named differently, etc.

Try to normalize the calculations by:
* Using `transform_points` instead of `transform_point`.
* Calculating the boundary in the correct orientation from the beginning.
* Using NumPy to calculate limits from boundary array instead of routing through Shapely.
* Use `_ellipse_boundary` for `Stereographic` even for circular projections. It's twice as fast and higher resolution by default, too.
* Always produce `LinearRing` boundaries instead of sometimes `LineString`s.

## Implications

The `Stereographic` boundary is slightly higher resolution in the default case, and all boundaries are rings; this causes a few minor differences in antialiased lines in the test figures. Times for creating the affected projections are approximately 100 microseconds less on my machine (a small benefit, but it's about 30-45% less than the original time.)

I'm also trying to write a 'How to add a projection' doc, so the more consistency the better.


QuLogic:
I made a small tweak of reverting the change to the `multiple_projections5.png` test image and instead increased the tolerance, because I want to add some new projections which will mean regenerating the image again.

Issue #1156, opened 2018-10-26
Projections that are ① equal-area, ② pseudo-cylindrical, ③ project poles as lines not points


gerritholl:
### Description

There exist a family of map projections that share three desirable properties:

* Pseudocylindrical, lines of latitude are parallel and straight
* Poles are lines, not points, so that lines of longitude can be labelled outside of the map area
* Equal area

Some of the map projections that support those characteristics:

* [Eckert IV](https://en.wikipedia.org/wiki/Eckert_IV_projection)
* [Wagner IV](http://www.manifold.net/doc/radian/wagner_iv_projection.htm)
* [Putniņš P4'](https://proj4.org/operations/projections/putp4p.html)
* [Eckert VI](https://en.wikipedia.org/wiki/Eckert_VI_projection)
* [Equal Earth](https://en.wikipedia.org/wiki/Equal_Earth_projection)

![image](https://user-images.githubusercontent.com/500246/47562074-32864100-d915-11e8-8c46-eaaa0a35c358.png)  
*Source: [Wikimedia Commons](https://commons.wikimedia.org/wiki/File:The-Equal-Earth-compared-to-similar-equal-area-pseudocylindrical-projections.png)*

Some others can be viewed via https://map-projections.net/singleview.php by selecting the pseudocylindrical equal-area properties, although it doesn't support to select the "poles are lines" property.  Currently, the only pseudocylindrical equal-area map projection supported by cartopy are the Mollweide and the Sinusoidal.  I'm currently using the Mollweide, but Šavrič et al. (2018) put a convincing argument for poles as lines rather than points.  The Sinusoidal map projection is ugly.  The Robinson has poles as points, but is not equal area.

For details on the Equal Earth projection as well as a convincing argument for the three aforementioned properties, see: Šavrič, Bojan; Patterson, Tom; Jenny, Bernhard (2018-08-07). "The Equal Earth map projection". International Journal of Geographical Information Science: 1–12. doi:10.1080/13658816.2018.1504949.

gerritholl:
Similar but slightly different projection request: https://github.com/SciTools/cartopy/issues/878

gerritholl:
All mentioned projections are [available in proj4](https://proj4.org/operations/projections/index.html).

dopplershift:
I’d be willing to merge a Pull Request implementing any of these. I’d wager they’d be pretty straightforward to implement based on the fact that they are supported by PROJ.

QuLogic:
[...]

PR #1168, opened 2018-11-05
Add Eckert family of projections

QuLogic:
## Rationale

Requested in #1156.

## Implications

Note, this is based on #1167, because I want to avoid some result changes that it _might_ create (though I didn't really check.)


QuLogic:
Here is [the Eckert family](https://732-5282596-gh.circle-artifacts.com/0/root/project/docs/build/html/crs/projections.html#eckerti) in the docs.

QuLogic:
Oops, just realized that `Eckert*` don't support elliptical globes, so I've dropped some extra tests and added the relevant warning for incompatible globes.

QuLogic:
Wish GitHub could figure that out automatically, but rebased.

dopplershift:
Test failure is #1171, merging over that.

PR #1170, opened 2018-11-06
Update mapbox tiles interface

dkillick:
The API for accessing mapbox tiles has updated. This PR brings the cartopy mapbox tiles interface in line with the updated mapbox API, and adds a new mapbox interface for retrieving user-defined map tiles (known as "styles" in the mapbox lexicon – see https://www.mapbox.com/studio-manual/overview/map-styling/#what-is-a-style).


lib/cartopy/io/img_tiles.py
```diff
@@ -285,7 +285,7 @@ class MapboxTiles(GoogleTiles):
     For terms of service, see https://www.mapbox.com/tos/.
 
     """
-    def __init__(self, access_token, map_id):
+    def __init__(self, access_token, map_name):
```
dopplershift: Technically, this is an API break, as it would break code calling by kwarg. But we are 0.x for a reason, so how do the rest of the devs feel about this? @pelson ?

dkillick: Ooh, good spot - thanks @dopplershift. I only changed it to be more descriptive of the situation and I'm not at all wedded to the change, so happy to change back for the sake of API consistency.

PR #1173, opened 2018-11-06
Add links to the Gitter chatroom

dopplershift:
Now that we have Gitter for chat, we should advertise it. (attn: @QuLogic @ocefpaf )


QuLogic:
Add to `docs/source/index.rst` as well?

dopplershift:
Good idea, done.

PR #1185, opened 2018-11-14
TST: Install all doc deps at the same time.

QuLogic:
This prevents any upgrade/downgrade dance.


QuLogic:
Well, it's better than the [upgrade/downgrade](https://circleci.com/workflow-run/d5e33b51-673b-4bc2-9aae-ac55fe0b285f) from before that produced 4.9.3, as we now get 5.0.1. I don't know what's pinning us below the latest yet.

PR #1189, opened 2018-11-15
Fix up the satellite projections (Fixes #1144)

### LINKED ISSUE #1144: Clipping issue with Geostationary/NearsidePerspective
I received this report from a user who was concerned by the points outside the circle of the globe:
```python
import numpy as np
import cartopy.crs as ccrs

lons, lats = np.meshgrid(np.linspace(0, 360, 100), np.linspace(-90, 90, 100))

data_projection = ccrs.Geodetic()
map_projection1 = ccrs.NearsidePerspective(central_longitude=0.0, central_latitude=0.0,     
                                           satellite_height=700000)

map_projection2 = ccrs.Geostationary(central_longitude=0.0, satellite_height=700000)

map_projections = [map_projection1, map_projection2]

for map_projection in map_projections:
    fig = plt.figure(figsize=(12, 8))
    ax = fig.add_subplot(1, 1, 1, projection=map_projection)

    # add some common geographic features
    ax.coastlines(resolution='10m', color='black')
    ax.add_feature(cfeature.STATES, edgecolor='black')
    ax.add_feature(cfeature.BORDERS)

    # add some lat/lon gridlines
    ax.gridlines()
    ax.set_global()
    ax.scatter(lons, lats, transform=data_projection)
```
produces these images for me.
NearsidePerspective:
cartopy-projection-bug-image2

Geostationary:
cartopy-projection-bug-image

I'm a little confused at the moment as to why these are so different as well.

dopplershift:
This refactors the `Geostationary` and `NearsidePerspective` projections so that they can individually set up their boundary, limits, etc. Doing so allows fixing the incorrect boundary being set for `NearsidePerspective`. The two projections can't share the math for the boundary because they are fundamentally different coordinate systems.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -2003,20 +2003,14 @@ class _Satellite(Projection):
             proj4_params.append(('sweep', sweep_axis))
         super(_Satellite, self).__init__(proj4_params, globe=globe)
 
-        # TODO: Let the globe return the semimajor axis always.
-        a = np.float(self.globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS)
-        b = np.float(self.globe.semiminor_axis or a)
-        h = np.float(satellite_height)
-        max_x = h * np.arcsin(a / (a + h))
-        max_y = h * np.arcsin(b / (a + h))
-
+    def _set_bounds(self, max_x, max_y):
+        false_easting = self.proj4_params['x_0']
+        false_northing = self.proj4_params['y_0']
         coords = _ellipse_boundary(max_x, max_y,
                                    false_easting, false_northing, 61)
         self._boundary = sgeom.LinearRing(coords.T)
-        mins = np.min(coords, axis=1)
-        maxs = np.max(coords, axis=1)
-        self._x_limits = mins[0], maxs[0]
-        self._y_limits = mins[1], maxs[1]
+        self._x_limits = -max_x + false_easting, max_x + false_easting
+        self._y_limits = -max_y + false_northing, max_y + false_northing
         self._threshold = np.diff(self._x_limits)[0] * 0.02
 
     @property
@@ -2038,12 +2032,38 @@ class _Satellite(Projection):
 
 class Geostationary(_Satellite):
     """
+    A view appropriate for satellites in Geostationary Earth orbit.
+
     Perspective view looking directly down from above a point on the equator.
+
+    In this projection, the projected coordinates are scanning angles measured
+    from the satellite looking directly downward, multiplied by the height of
+    the satellite.
 
     """
     def __init__(self, central_longitude=0.0, satellite_height=35785831,
                  false_easting=0, false_northing=0, globe=None,
                  sweep_axis='y'):
+        """
+        Parameters
+        ----------
+        central_longitude: float, optional
+            The central longitude. Defaults to 0.
+        satellite_height: float, optional
+            The height of the satellite. Defaults to 35785831 meters
+            (true geostationary orbit).
+        false_easting:
+            X offset from planar origin in metres. Defaults to 0.
+        false_northing:
+            Y offset from planar origin in metres. Defaults to 0.
+        globe: :class:`cartopy.crs.Globe`, optional
+            If omitted, a default globe is created.
+        sweep_axis: 'x' or 'y', optional. Defaults to 'y'.
+            Controls which axis is scanned first, and thus which angle is
+            applied first. The default is appropriate for Meteosat, while
+            'x' should be used for GOES.
+        """
+
         super(Geostationary, self).__init__(
             projection='geos',
             satellite_height=satellite_height,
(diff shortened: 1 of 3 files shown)


dopplershift:
Before this PR:

before

After:

after

(See #1144 for sample code.)

lib/cartopy/crs.py
```diff
@@ -2038,12 +2032,38 @@ def y_limits(self):
...
+    the satellite.
+
     """
     def __init__(self, central_longitude=0.0, satellite_height=35785831,
                  false_easting=0, false_northing=0, globe=None,
                  sweep_axis='y'):
+        """
+        Parameters
+        ----------
+            central_longitude: float, optional
```
QuLogic: Unnecessary indent. It looks like there's no effect on the output, but I think that's because napoleon is a bit lenient and/or the theme settings.

lib/cartopy/crs.py
```diff
@@ -2054,15 +2074,59 @@ def __init__(self, central_longitude=0.0, satellite_height=35785831,
...
+    point (e.g. a satellite).
+
     """
     def __init__(self, central_longitude=0.0, central_latitude=0.0,
                  satellite_height=35785831,
                  false_easting=0, false_northing=0, globe=None):
+        """
+        Parameters
+        ----------
+            central_longitude: float, optional
```
QuLogic: Unnecessary indent too.

lib/cartopy/crs.py
```diff
@@ -2054,15 +2074,59 @@ def __init__(self, central_longitude=0.0, satellite_height=35785831,
...
+        if globe is None:
+            globe = Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS, ellipse=None)
+
+        # TODO: Let the globe return the semimajor axis always.
+        a = globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS
+        b = globe.semiminor_axis or a
+
+        if b != a or globe.ellipse is not None:
+            warnings.warn('The proj "nsper" projection does not handle '
+                          'elliptical globes.')
```
QuLogic: So I've been thinking, maybe I should've moved this to a base class. Maybe change the default `Globe` creation based on an `__init__` parameter, or private class attribute. Something for another PR, though.

lib/cartopy/tests/crs/test_nearside_perspective.py
```diff
@@ -23,28 +23,48 @@
...
+import cartopy.crs as ccrs
 
 
-class TestEquatorialDefault(GeostationaryTestsMixin, object):
-    # Check that it behaves just like Geostationary, in the absence of a
-    # central_latitude parameter.
-    test_class = NearsidePerspective
-    expected_proj_name = 'nsper'
+class TestNearsidePerspective(object):
+    def test_default(self):
```
QuLogic: Since there is no longer any subclassing, might as well write these at the top-level.

lib/cartopy/tests/crs/test_nearside_perspective.py
```diff
@@ -23,28 +23,48 @@
...
+                      'x_0=0', 'y_0=0'}
+
+        check_proj4_params('nsper', geos, other_args)
+
+        assert_almost_equal(geos.boundary.bounds,
+                            (-5476336.098, -5476336.098,
+                             5476336.098, 5476336.098),
+                            decimal=3)
+
+    def test_eastings(self):
```
QuLogic: Rename to `test_offset` to match the others.

lib/cartopy/crs.py
dopplershift: 🐑 fixed

lib/cartopy/crs.py
dopplershift: fixed

lib/cartopy/crs.py
dopplershift: yeah, I think there's a few more things we could hoist out of the leaf nodes in the tree of projections.

lib/cartopy/tests/crs/test_nearside_perspective.py
dopplershift: Done.

lib/cartopy/tests/crs/test_nearside_perspective.py
dopplershift: Done.

lib/cartopy/crs.py
pelson: Or more specifically - into a factory of Globe rather than the ``__init__`` of a baseclass.

lib/cartopy/tests/crs/test_nearside_perspective.py
dopplershift: OH, I understand better now.

dopplershift: Thought you meant the weird subclassing still going on for Geostationary. Apparently reading comprehension is not a strong suit for me.

dopplershift: Now actually fixed.

lib/cartopy/crs.py
```diff
@@ -2038,12 +2032,38 @@ def y_limits(self):
...
     def __init__(self, central_longitude=0.0, satellite_height=35785831,
                  false_easting=0, false_northing=0, globe=None,
                  sweep_axis='y'):
+        """
+        Parameters
+        ----------
+        central_longitude: float, optional
+            The central longitude. Defaults to 0.
+        satellite_height: float, optional
+            The height of the satellite. Defaults to 35785831 meters
```
QuLogic: This is very minor, but `false_easting/northing` use metres (which I guess you copied from the others.)

PR #1198, opened 2018-11-16
Refactor ellipse warnings and add tests for spherical-only projections

QuLogic:
## Rationale

The warnings and fallback to a spherical globe occur in several projections. This moves that check down to the base class so it does need to be repeated so many times. Some projections did not correctly handle said fallback or were missing a warning. Also, move the `check_proj_params` into one place so it doesn't have to be copied to every projection's tests.

## Implications
* Warnings about ellipses are consistent.
* `Orthographic` boundaries are now correct for eccentric globes (even though the projection doesn't support it.)
* `Gnomonic` now also warns that it doesn't support ellipses.
* Projections that don't support ellipses are now tested for not supporting ellipses (using a non-default sphere, an ellipse, and an eccentric ellipse, the latter two of which should warn.)

### DIFF
--- a/lib/cartopy/_crs.pyx
+++ b/lib/cartopy/_crs.pyx
@@ -60,6 +60,9 @@ if _match is not None:
 else:
     PROJ4_VERSION = ()
 
+WGS84_SEMIMAJOR_AXIS = 6378137.0
+WGS84_SEMIMINOR_AXIS = 6356752.3142
+
 
 class Proj4Error(Exception):
     """
@@ -134,6 +137,10 @@ cdef class CRS:
     Define a Coordinate Reference System using proj.
 
     """
+
+    #: Whether this projection can handle ellipses.
+    _handles_ellipses = True
+
     def __cinit__(self):
         self.proj4 = NULL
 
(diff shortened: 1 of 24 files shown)


lib/cartopy/tests/crs/__init__.py
```diff
@@ -20,3 +20,8 @@
 """
 
 from __future__ import (absolute_import, division, print_function)
+
+import pytest
+
+
+pytest.register_assert_rewrite('cartopy.tests.crs.helpers')
```
dopplershift: This was a new one for me. Interesting.

dopplershift:
Of course, this conflicts with the changes I made to `test_geostationary.py`. Can you rebase?

lib/cartopy/tests/crs/__init__.py
QuLogic: Yea, I would've put everything in `__init__.py`, but not having assertion rewriting is annoying.

dopplershift:
So...why is the shape file test failing now?

QuLogic:
I'm not sure, but it seems a rebase fixed it.

Issue #1207, opened 2018-11-18
tight_layout and Matplotlib 3


deeplycloudy:
### Description

Recently, I was working on a multi panel map figure. The figure used `tight_layout`. The figures produced in Matplotlib 2 have smaller margins than those in Matplotlib 3. 

As @tacaswell [pointed out on Twitter](https://twitter.com/tacaswell/status/1057633159141982211) in a thread with @dopplershift , “More things now participate it [sic] tight_layout.” A warning message received while running suggests that perhaps additional Cartopy “axes decorations” participate in tight_layout in Matplotlib 3.

Below, I include four plots (both MPL versions; with and without Caropy axes) that demonstrate the problem. The upstream changes in Matplotlib 3 apparently result in more than cosmetic layout differences on Cartopy axes for these relatively simple plots.

![2 2 3](https://user-images.githubusercontent.com/1325771/48678076-19536780-eb44-11e8-9c69-2e735121d47f.png)
![3 0 2](https://user-images.githubusercontent.com/1325771/48678077-240dfc80-eb44-11e8-999b-5bb8cd508b22.png)
![2 2 3_no_map](https://user-images.githubusercontent.com/1325771/48678078-27a18380-eb44-11e8-918a-d2faae8dc6c5.png)
![3 0 2_no_map](https://user-images.githubusercontent.com/1325771/48678080-2a03dd80-eb44-11e8-8318-224c290eab8f.png)

#### Code to reproduce

I created two minimal environments:

```bash
conda create -n mpl2 -c conda-forge matplotlib=2 python=3.7 cartopy

conda create -n mpl3 -c conda-forge matplotlib=3 python=3.7 cartopy
```

Then using the plotting script, I received the following output.

```bash
ebruning$ source activate mpl3
(mpl3) ebruning$ python plot.py 
Plotting with map
/Users/ebruning/anaconda/envs/mpl3/lib/python3.7/site-packages/matplotlib/tight_layout.py:176: UserWarning: Tight layout not applied. The left and right margins cannot be made large enough to accommodate all axes decorations. 
  warnings.warn('Tight layout not applied. The left and right margins '
Plotting with no map
(mpl3) ebruning$ source deactivate
ebruning$ source activate mpl2
(mpl2) ebruning$ python plot.py 
Plotting with map
Plotting with no map
```

The plotting script
```python
import numpy as np
import cartopy.crs as ccrs

import matplotlib
matplotlib.use('Agg')
ver = matplotlib.__version__

[...]

PR #1213, opened 2018-11-20
Turn GeoAxes.outline_patch into a GeoSpine.

### LINKED ISSUE #829: frameon=False does not remove map frame
With cartopy, using ``plt.axes(..., frameon=False)`` does not remove the map frame. One has to use the seldom-used ``plt.gca().outline_patch.set_visible(False)`` command. This should probably be fixed.

### LINKED ISSUE #1077: Cannot change the border width of the cartopy axes
### Description

I just started to use cartopy.  I cannot change the (frame) border width of the cartopy axes by using mpl.rcParams['axes.linewidth']. But mpl.rcParams always works for other axes or basemap.

#### Code to reproduce
``` python
import cartopy as cpy
import cartopy.feature as cpf
import matplotlib as mpl

mpl.rcParams['axes.linewidth'] = 3

ax = plt.axes(projection=cpy.crs.PlateCarree())
ax.add_feature(cpf.LAND, facecolor='0.9', linewidth=0.5)
ax.add_feature(cpf.OCEAN, linewidth=0.5)
ax.coastlines(resolution='50m', linewidth=0.5)

plt.savefig('cpy.test.png', dpi=600, bbox_inches='tight')
```

#### Traceback 

```

```

<details>

### Operating system
Windows 10
Python 2.7.14

### Cartopy version
cartopy-0.16.0
shapely-1.6.4
owslib-0.16.0
pyepsg-0.3.2

Matplotlib 2.2.2

### conda list

```
```

### pip list

```
```

</details>

QuLogic:
## Rationale

This patch deprecates `GeoAxes.outline_patch` as well, since the spine machinery works fine for what we want to do. The dictionary key is 'geo' to match the one used by the Matplotlib `GeoAxes`.

The `GeoSpine` is a minimal implementation to have it work as it did before, but more things may be implemented in the future if necessary.

Since `PathPatch` has default line width of 1.0, and spines have default line width of 0.8, some boundary lines have changed slightly in the test images.

## Implications

Fixes #1077.
Fixes #829.
First step towards fixing up #1207.


QuLogic:
Small amendment; I decided to override the two public methods of `Spine` we needed, instead of just the private method, as I wanted to avoid another `_hold` situation.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1978,11 +2006,11 @@ def add_wms(self, wms, layers, wms_kwargs=None, **kwargs):
...
-    Define an event callback for a GeoAxes which forces the outline and
-    background patches to be re-clipped next time they are drawn.
+    Define an event callback for a GeoAxes which forces the background patch to
+    be re-clipped next time it is drawn.
 
     """
     axes = event.axes
     # trigger the outline and background patches to be re-clipped
-    axes.outline_patch.reclip = True
+    axes.spines['geo'].stale = True
```
dopplershift: The docstring only talks about forcing the background patch to be reclipped. Do we really need to set stale here?

QuLogic: Originally, no matter which had `reclip=True`, [both would be updated](https://github.com/SciTools/cartopy/pull/1213/files/fbc8de276a37679e7b14657a0322b0faab76cdb9#diff-b599a1ae2427ad1c63351afff7e4e9c5R411). Setting `stale` now is equivalent to setting what `reclip` did except each thing happens independently.

It doesn't seem like this is necessary now (at least from tests) and I don't remember exactly why I left it, but I believe this will be necessary when `backgroud_patch` is removed. I will try again when I get to working on that once more.

PR #1222, opened 2018-11-29
Test against oldest Cython.

QuLogic:
We technically require Cython >= 0.15.1, but don't really check that. I'm not really sure if this will work, but this PR tries to build it that way, if it turns out to be possible.


QuLogic:
Had to bumpy to 0.17, since that's the oldest on conda now.

PR #1234, opened 2018-12-05
Ensure boundary paths are always closed.

### LINKED ISSUE #1233: OSNI Projection Grid Lines seem to fail on version 0.17.0
Directly from pip
image

cartopy==0.17.0
image

cartopy==0.16.0
image

QuLogic:
## Rationale

Boundaries are currently inconsistent about whether they are `LinearRing`s or `LineString`s. The input to `Path.clip_to_bbox` is [expected to be closed](https://matplotlib.org/api/path_api.html#matplotlib.path.Path.clip_to_bbox).

## Implications

This makes boundaries consistently `LinearRing`s, and handles them in `geos_to_path` so that closed `Path`s are created.
Fixes #1233

### DIFF
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -45,6 +45,7 @@ def geos_to_path(shape):
         A list, tuple or single instance of any of the following
         types: :class:`shapely.geometry.point.Point`,
         :class:`shapely.geometry.linestring.LineString`,
+        :class:`shapely.geometry.linestring.LinearRing`,
         :class:`shapely.geometry.polygon.Polygon`,
         :class:`shapely.geometry.multipoint.MultiPoint`,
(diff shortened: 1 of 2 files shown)


lib/cartopy/mpl/patch.py
```diff
@@ -64,7 +65,9 @@ def geos_to_path(shape):
             paths.extend(geos_to_path(shp))
         return paths
 
-    if isinstance(shape, (sgeom.LineString, sgeom.Point)):
+    if isinstance(shape, sgeom.LinearRing):
+        return [Path(np.column_stack(shape.xy), closed=True)]
+    elif isinstance(shape, (sgeom.LineString, sgeom.Point)):
```
QuLogic: Note: a `LinearRing` is a `LineString`, so that's why this conversion worked before.

PR #1249, opened 2018-12-27
Add Matplotlib geo projections

QuLogic:
## Rationale

In Matplotlib, there are 4 builtin geo projections: Aitoff, Hammer, Mollweide, and Lambert. They are not quite as flexible as Cartopy, but work for maps at least. We have the latter two, but not the former.

## Implications

This adds Aitoff and Hammer. In proj, [Hammer support two more parameters](https://proj4.org/operations/projections/hammer.html): `+W` and `+M`. I do not know what these stand for, so I have not added them, but would like to if anyone has any understanding of what they mean.


QuLogic:
Any thoughts on Hammer parameters?

It seems like 4.9.3 might be buggy with one of these.

pelson:
> Any thoughts on Hammer parameters?

As in ``+W`` and ``+M``? I'm quite content to not provide these until they are needed (in the form of a PR from the person that does). 

The test is failing with a lot of warnings, but they may be not an issue? The image that now has all of the projections is (copied from Travis):

full

greglucas:
Interest in coming back to these projections?

QuLogic:
Rebased, but I did not add the other parameters. I'm also not sure if the threshold needs setting nowadays.

lib/cartopy/crs.py
```diff
@@ -2112,6 +2112,45 @@ def y_limits(self):
...
+        globe: :class:`cartopy.crs.Globe`, optional
+            If omitted, a default globe is created.
+
+            .. note::
+                This projection does not handle elliptical globes.
+
+        """
+        proj_params = [('proj', 'aitoff'),
+                       ('lon_0', central_longitude)]
+        super(Aitoff, self).__init__(proj_params, central_longitude,
```
greglucas, line 2147:
```suggestion
        super().__init__(proj_params, central_longitude,
```
You must have had this sitting around for a while needing to rebase from Python 2 supers :)

lib/cartopy/tests/crs/test_aitoff.py
```diff
@@ -0,0 +1,97 @@
+# (C) British Crown Copyright 2021, Met Office
```
greglucas, line 1: Should change this to the new license file header.

lib/cartopy/crs.py
QuLogic: Yea, I did catch the `__future__` imports at least.

QuLogic:
Oops, created the test images with the oldest supported Matplotlib instead of the latest.

lib/cartopy/crs.py
```diff
@@ -2117,6 +2117,45 @@ def y_limits(self):
...
+            The central longitude. Defaults to 0.
+        false_easting: float, optional
+            X offset from planar origin in metres. Defaults to 0.
+        false_northing: float, optional
+            Y offset from planar origin in metres. Defaults to 0.
+        globe: :class:`cartopy.crs.Globe`, optional
+            If omitted, a default globe is created.
+
+            .. note::
+                This projection does not handle elliptical globes.
```
greglucas, line 2147: It doesn't look like this note appears anywhere in the docs? It seems like all of the __init__ docstrings are left out currently.

greglucas (approved):
I just had one minor comment, which should probably be a follow-up PR. Looks good to me otherwise.

lib/cartopy/crs.py
QuLogic: Hmm, maybe it was with the change in doc themes? I don't remember what the older one looked like now.

greglucas:
Sorry, I actually meant to merge this in before the release and just forgot.

lib/cartopy/crs.py
greglucas: We can open up another issue if we want to update the docs later.

Issue #1140, opened 2018-10-07
Convert to proj v5 API


QuLogic:
>The proj_api.h header and the functions related to it is considered deprecated from version 5.0.0 and onwards. The header will be removed from PROJ in version 7.0.0 scheduled for release February 1st 2020.

This applies to `trace.pyx`.

_Originally posted by @dopplershift in https://github.com/SciTools/cartopy/issues/1045#issuecomment-390043613_

QuLogic:
xref to jswhit/pyproj#152

snowman2:
I would be happy to take a stab at it if no one else is planning to do it anytime soon.

aeroevan:
Now that proj 6 is out, it might be nice to define `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` for the proper proj_version until the API update is complete.

adamjstewart:
Just a heads up, PROJ 7 is now out...

dopplershift:
Looks like removing `proj_api.h` has been postponed to 8.0 according to OSGeo/PROJ#837. Which explains how cartopy was built against PROJ 7.1 on conda-forge. I'd love to solve this by taking a stab at #1477.

snowman2:
The `proj_api.h` has been removed from the PROJ src code. Change will be released in PROJ 8.

sebastic:
PROJ 8.0.0-rc1 has been released with which it's impossible to build cartopy.

bjkindia:
> The `proj_api.h` has been removed from the PROJ src code. Change will be released in PROJ 8.

In this case how can we install Cartopy ? 
I am getting error 
`lib/cartopy/trace.cpp:634:10: fatal error: 'proj_api.h' file not found
  #include "proj_api.h"
           ^~~~~~~~~~~~`

dopplershift:
@bjkindia For now, the workaround is to install PROJ<8.

bjkindia:
Well finally I installed with Conda. I can't solve with manually. 

Conda is great

Issue #1164, opened 2018-11-01
Add LatLong "projection"


dopplershift:
Go ahead and wrap Proj.4's `latlong` [pseudo-projection](https://proj4.org/usage/projections.html#units). This should provide an alternative to `PlateCarree` to fix #709.

dopplershift:
Looks non-trivial to make `latlong` accept something other than radians. Possible with [pipelines](https://proj4.org/operations/conversions/unitconvert.html), but not happening for 0.17.

deeplycloudy:
Unless there is some other configuration of pyproj being used, it has always been my understanding that `latlong` takes degrees.

```
import pyproj as proj4
lon = np.array([-101., -101.])
lat = np.array([33., 34.]) # Should be about 100 km distance
alt= np.array([0.,0.])
ellipse=datum='WGS84'
ERSlla = proj4.Proj(proj='latlong', ellps=ellipse, datum=datum)

# ECEF cartesian, meters
ERSxyz = proj4.Proj(proj='geocent', ellps=ellipse, datum=datum)
x,y,z = proj4.transform(ERSlla, ERSxyz, lon, lat, alt)
dx = np.diff(x)
dy = np.diff(y)
dz = np.diff(z)
print(np.sqrt(dx*dx + dy*dy + dz*dz)/1000.0, ' km')
```
Which results in 111 km.

QuLogic:
@dopplershift Do you mean https://proj4.org/operations/conversions/latlon.html? Is that not `Geodetic`?
```
>>> ccrs.Geodetic().proj4_init
'+datum=WGS84 +ellps=WGS84 +proj=lonlat +no_defs'
```

dopplershift:
I want a `Projection` so that we can hand it to matplotlib. `Geodetic` does not work for this.

deeplycloudy:
Well, I’m the one that spoke without realizing that cartopy has its own proj.4 wrapper, and I was using the pyproj wrapper in my example above, instead of the cartopy `_crs.pyx` and its class `CRS`. 

So I may be responsible for some noise. But here is some analysis of what pyproj does. It seems related to why I’ve never worried about radians when using it.

[...]

PR #1252, opened 2018-12-31
WIP: updating to proj.4 6.0.0

snowman2:
This likely has bugs and whatnot at this stage, but it gets the concepts of changes needed out there for discussion.

## Rationale

#1140


snowman2:
Current issues in CRS tests is the error:
```
proj_create_operations: Source and target ellipsoid do not belong to the same celestial body
```

snowman2:
Also, this will mean that python 2.7 won't work on windows due to c++11 requirement for Proj 6.0.0

dopplershift:
I think we were already close to dropping 2.7, so this is probably as good a reason as any. Might want to have one more release before we merge this...but then we'd have to be careful about what proj C library is used to build. Ick. @pelson ?

snowman2:
I wouldn't recommend merging this just yet. It was created before 6.0.0 was released and the API has changed since then.

snowman2:
Well, at least the errors I mentioned earlier are showing in the build now :)

snowman2:
I see a lot of failures due to the `Source and target ellipsoid do not belong to the same celestial body` error.

More information about this issue:
- https://lists.osgeo.org/pipermail/proj/2019-March/008390.html
- https://lists.osgeo.org/pipermail/proj/2019-May/008586.html

Essentially, the tests are using invalid transformations in multiple places if I understand correctly.

snowman2:
Question for @pelson @dopplershift @QuLogic 
Do y'all know the reasoning behind the logic for the default globe for `PlateCaree`?

https://github.com/SciTools/cartopy/blob/1942c7bd0a6d28c78da3d7f0e4c646833287f44a/lib/cartopy/crs.py#L705

```python
>>> from cartopy.crs import PlateCarree
>>> pc = PlateCarree()
>>> pc.proj4_init
'+proj=eqc +ellps=WGS84 +a=57.29577951308232 +lon_0=0.0 +no_defs'
``` 
The semimajor_axis varies quite a bit from any other value I have seen.

Also, from what I can tell, it seems that meters are the units expected for the semi major axis:
```python
>>> from pyproj import CRS
>>> crs = CRS('+proj=eqc +ellps=WGS84 +a=57.29577951308232 +lon_0=0.0 +no_defs')
>>> crs.ellipsoid
ELLIPSOID["unknown",57.2957795130823,298.257223563,
    LENGTHUNIT["metre",1,
        ID["EPSG",9001]]]
>>> crs.ellipsoid.semi_major_metre
57.29577951308232
```

snowman2:
Another data point on the plate carree ellipsoid base on https://proj.org/operations/projections/eqc.html:

```python
>>> from pyproj import CRS
>>> cc = CRS.from_epsg(32622)
>>> cc.ellipsoid
ELLIPSOID["WGS 84",6378137,298.257223563,
    LENGTHUNIT["metre",1],
    ID["EPSG",7030]]
>>> cc.ellipsoid.semi_major_metre
6378137.0
>>> ceqc = CRS("+proj=eqc +lat_ts=0")
>>> ceqc.ellipsoid
ELLIPSOID["WGS 84",6378137,298.257223563,
    LENGTHUNIT["metre",1],
    ID["EPSG",7030]]
>>> ceqc.ellipsoid.semi_major_metre
6378137.0
```
It seems `6378137.0` could be a good default?

dopplershift:
@snowman2 That particular choice of ellipsoid as the default for `PlateCarree` is a ~~hack~~ clever solution to allow passing lats and lons as Cartesian coordinates to `PlateCarree`, which is really just an equirectangular projection. That choice of `Globe` also means you get unexpected behavior when changing it (i.e. it no longer takes lons/lats). See #709.

snowman2:
Thanks for the explanation - that make sense. Well, it seems this is no longer allowed in PROJ 6+, so it will need to be updated before upgrading.

Based on #1164, I think it would make sense to use latlon. In pyproj we default to assume all input is degrees unless you toggle the `radians` flag and do conversions accordingly as needed by PROJ:
https://github.com/pyproj4/pyproj/blob/cecbdf4a516bf031c6e8fab0cf7188148dafc9f4/pyproj/_transformer.pyx#L164-L176

Is this something you would be interested in?

dopplershift:
@snowman2 I think what we need is:
1. Make a new lat/lon projection that takes degrees
2. Add an EquidistantCylindrical projection that works like everything else with regards to globes
3. Deprecate `PlateCarree`, or at least its current behavior

snowman2:
@dopplershift, I think something just clicked in my brain. Is the goal to create something that converts from the "WGS84" or "EPGS:4326" to a global projected coordinate system? Or to convert from the geodetic CRS to the projected CRS. Either way, I think a new base class is needed. This class would be similar to the [pyproj.Transformer](http://pyproj4.github.io/pyproj/stable/api/transformer.html#pyproj-transformer), however it would default to have the `from_crs` always be either "WGS84" or the geodetic CRS based on the projected CRS depending on what you want (See: http://pyproj4.github.io/pyproj/stable/gotchas.html#proj-not-a-generic-latitude-longitude-to-projection-converter). This would enable the `globe` parameter to be supported as well open up the option for users to use additional `Transformer` classes that convert from latlon to a projected coordinate system.

snowman2:
An idea of different global projected coordinate systems could include:
EPSG:3857 (Pseudo web mercator)
EPSG:3395 "WGS 84 / World Mercator"

Ideas from: https://lists.osgeo.org/pipermail/proj/2019-June/008653.html

snowman2:
Currently upgrading to use the `proj.h` API in PROJ 6+ (an soon 7) has many changes/challenges. @dopplershift has expressed interest in potentially using `pyproj 2` itself within `cartopy` to replace the direct need for building against the `proj.h` API. Is this something that the `cartopy` project is interested in?

greglucas:
I'm :+1: on using pyproj2, since you've already built in all of the new Transformer hooks. Let's see how others feel about this too though. Would this remove the need for Cartopy to use pyepsg too? There was a bit of a hiccup recently when epsg.io went down and not depending on an external website would be nice.

Cartopy is currently still supporting some pretty old Python versions, so we would have to drop some older Python versions to make this work it looks like.

I think that all of this should actually be thought about from a broader perspective of what packages should handle what operations in a future big release. It looks like pygeos and Shapely will be [merging their GEOS library work together](https://github.com/Toblerity/Shapely/issues/782) in the (hopefully not too distant) future. There was a discussion about [moving the GEOS optimizations out of Cartopy](https://github.com/SciTools/cartopy/issues/805). Maybe this goes in parallel with that update to start moving some of the Cartopy Cython optimization code into other upstream libraries where it makes sense. (i.e. code in `trace.pyx` to shapely/pygeos, and `_crs.pyx` into pyproj2)

snowman2:
> Let's see how others feel about this too though.

Should I open up a separate issue?

> Would this remove the need for Cartopy to use pyepsg too?

Yes, it will remove that.

dopplershift:
@snowman2 A separate issue might make sense, since we're veering off the topic of this PR. I'll go ahead and open it with my thoughts.

dopplershift:
See #1477.

snowman2:
Going to close in favor of #1477.

PR #1273, opened 2019-02-06
ensure the setup.py directory is in sys.path because of PEP 517

ngoldbaum:
## Rationale

Currently cartopy is failing to build on e.g. Travis with the following traceback:

```
    File "/home/travis/build/ngoldbaum/yt/venv/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
      main()
    File "/home/travis/build/ngoldbaum/yt/venv/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/travis/build/ngoldbaum/yt/venv/lib/python2.7/site-packages/pip/_vendor/pep517/_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-dD1vz4/overlay/lib/python2.7/site-packages/setuptools/build_meta.py", line 130, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-dD1vz4/overlay/lib/python2.7/site-packages/setuptools/build_meta.py", line 112, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-dD1vz4/overlay/lib/python2.7/site-packages/setuptools/build_meta.py", line 126, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 36, in <module>
      import versioneer
  ImportError: No module named versioneer
```

The issue is that in the latest version of pip, they enabled PEP 517 and PEP 518 support, which means that the directory of setup.py is no longer included in `sys.path`. Since cartopy uses versioneer and we *want* that behavior, the fix is to explicitly add the path to the `setup.py` directory to `sys.path`.

This was suggested to me by @pganssle on IRC.

## Implications

I think this is harmless for older setuptools and pip verisons, because `.` is already in sys.path.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,12 +33,17 @@ import subprocess
 import sys
 import warnings
 
+# ensure the current directory is in sys.path so versioneer can be imported
+# this is needed because PEP 517 makes the directory of setup.py not included
+# in sys.path
+HERE = os.path.dirname(__file__)
+sys.path.append(HERE)
+
 import versioneer
 
 
 # The existence of a PKG-INFO directory is enough to tell us whether this is a
 # source installation or not (sdist).
-HERE = os.path.dirname(__file__)
 IS_SDIST = os.path.exists(os.path.join(HERE, 'PKG-INFO'))
 
 if not IS_SDIST:


ngoldbaum:
I signed the CLA.

pganssle:
One thing to note is that even though, after the next `pip` release, this won't be *necessary*, you will need to deal with this when you eventually explicitly specify `setuptools.build_meta` as your PEP 517 backend. As long as your build system works with PEP 517, you can also explicitly specify `setuptools.build_meta` as your backend in your `pyproject.toml`, since you won't be relying on the legacy behavior.

```toml
[build-system]
requires = [
    "setuptools >= 0.40.8",
    "wheel",
    "Cython >= 0.28",
    "numpy >= 1.10",
]
build-backend="setuptools.build_meta"
```

I recommend doing this in a separate PR, though. Your build is complicated enough that you should probably test the transition better than you need to test this.

ngoldbaum:
And I guess this isn't strictly necessary to build cartopy because `pyproject.toml` was removed in https://github.com/SciTools/cartopy/pull/1264. But it would be necessary if you ever re-added `pyproject.toml`.

ocefpaf:
I was looking into this recently and maybe the best course of action is to drop `versioneer` instead of append the current path to `sys.path`. At @ngoldbaum and @pganssle do you know any nice `versioneer` alternatives out there?

ngoldbaum:
I think `setuptools_scm` plays more nicely with pep517.

ocefpaf:
> I think `setuptools_scm` plays more nicely with pep517.

Thanks! I'll play with it in other projects I have more control and, once I feel comfortable with it, I'll send a PR here. (Unless someone beats me to it first.)

QuLogic:
I do like `setuptools_scm`, especially since we have some patched copy of `versioneer`, which also seems dead upstream. If it fixes things, I would be happy to start using it.

dopplershift:
I've switched MetPy over to setuptools_scm (if for no other reason than it's alive and under the PyPA umbrella). So far, no complaints. The only gotcha is that it's actual main purpose in life is to include *everything* under source control in release tarballs; for MetPy this took some adjusting of things, but in the end I think the product was better.

IIRC, the versioneer here is tweaked, so we'd have to look at either changing the versioning behavior or adjusting to something new.

QuLogic:
I looked at `setuptools_scm`, and I think it requires a new-ish version of setuptools. I don't know if we can rely on that for Python 2, but as soon as 0.18 is out and we drop support for Py2, then I'll look at it again.

QuLogic:
I've opened #1545 to switch to setuptools-scm.

QuLogic:
So I think we don't need this any more with the above PR merged.

PR #1299, opened 2019-04-18
FIX: remove unused default_params

jklymak:
## Rationale

The copy of the default_params causes an error on matplotlib master.  
```
~/Cartopy/lib/cartopy/mpl/ticker.py in LongitudeLocator()
    469     """
    470 
--> 471     default_params = MaxNLocator.default_params.copy()
    472     default_params.update(nbins=8, dms=False)
    473 

AttributeError: '_deprecated_property' object has no attribute 'copy'
```

This change avoids using `default_params` at all.

### DIFF
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -467,12 +467,14 @@ class LongitudeLocator(MaxNLocator):
     dms: bool
         Allow the locator to stop on on minutes and seconds (False by default)
     """
+    def __init__(self, *args, **kwargs):
+        nbins = kwargs.pop('nbins', 8)
+        super().__init__(*args, nbins=nbins, **kwargs)
 
     def set_params(self, **kwargs):
         """Set parameters within this locator."""
-        nbins = kwargs.pop('nbins', 8)
         self._dms = kwargs.pop('dms', False)
-        MaxNLocator.set_params(self, nbins=nbins, **kwargs)
+        MaxNLocator.set_params(self, **kwargs)
 
     def _guess_steps(self, vmin, vmax):
 
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -469,7 +469,7 @@ class LongitudeLocator(MaxNLocator):
     """
     def __init__(self, *args, **kwargs):
         nbins = kwargs.pop('nbins', 8)
-        MaxNLocator(*args, nbins=nbins, **kwargs)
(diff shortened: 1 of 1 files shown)


QuLogic:
`default_params` is used by `MaxNLocator` and this derives from it, so it's not unused. Matplotlib 3.1 is making it private though, so that's what the error is.

jklymak:
Really?  
```python
old = {'a': 1, 'b':2}
new = old.copy()
new.update(a=3)
print('new', new)
print('old', old)
```

```
new {'a': 3, 'b': 2}
old {'a': 1, 'b': 2}
```

QuLogic:
Yes, that's the point:
```ipython
In [11]: class A:
    ...:     defaults = {'a': 1, 'b': 2, 'c': 3}
    ...:     def __init__(self):
    ...:         print(self.defaults)
    ...:         

In [12]: A()
{'a': 1, 'b': 2, 'c': 3}
Out[12]: <__main__.A at 0x7f176002ae10>

In [13]: class B(A):
    ...:     defaults = A.defaults.copy()
    ...:     defaults.update({'c': 5, 'b': 6, 'a': 7})
    ...:     

In [14]: B()
{'a': 7, 'b': 6, 'c': 5}
Out[14]: <__main__.B at 0x7f1760032198>

In [15]: A()
{'a': 1, 'b': 2, 'c': 3}
Out[15]: <__main__.A at 0x7f17600326d8>
```

jklymak:
OMG, thats pretty obscure.  Thanks, and sorry for the noise.

jklymak:
OK, I *think* this does the same thing w/o the copy

lib/cartopy/mpl/ticker.py
```diff
@@ -468,14 +468,11 @@ class LongitudeLocator(MaxNLocator):
...
 
-    default_params = MaxNLocator.default_params.copy()
-    default_params.update(nbins=8, dms=False)
-
     def set_params(self, **kwargs):
         """Set parameters within this locator."""
-        MaxNLocator.set_params(self, **kwargs)
-        if 'dms' in kwargs:
-            self._dms = kwargs['dms']
+        nbins = kwargs.pop('nbins', 8)
```
QuLogic: This won't work, because `__init__` will use `nbins` from `default_params` always.

jklymak: OK, that is a bonkers way to set default params, but I think the new change works.  Not quite sure if the documentation gets inherited the right way...

jklymak:
I assume this has somehow fixed itself some other way....

PR #1308, opened 2019-05-18
Fix calls to pytest.raises.

QuLogic:
## Rationale

The `message` argument is the output if the check fails, not the message to check. Also, it's a regex, so escape special characters.

## Implications

Fixes deprecation warnings;
```
lib/cartopy/tests/mpl/test_ticker.py:41
  /home/elliott/code/cartopy/lib/cartopy/tests/mpl/test_ticker.py:41: PytestDeprecationWarning: The 'message' parameter is deprecated.
  (did you mean to use `match='some regex'` to check the exception message?)
  Please comment on https://github.com/pytest-dev/pytest/issues/3974 if you have concerns about removal of this parameter.
    with pytest.raises(TypeError, message=message):
```

### DIFF
--- a/lib/cartopy/tests/io/test_ogc_clients.py
+++ b/lib/cartopy/tests/io/test_ogc_clients.py
@@ -74,8 +74,8 @@ class TestWMSRasterSource(object):
         assert source.layers == self.layers
 
     def test_no_layers(self):
-        msg = 'One or more layers must be defined.'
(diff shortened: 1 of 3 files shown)


dopplershift:
Ugh. NOW why are failing?

QuLogic:
At least one env passed, and the ones that didn't are for unrelated reasons, so I'm going to merge this.

PR #1309, opened 2019-05-19
Use HTTPS URIs in code and documentation

jynik:
(*Updated*)

## Rationale

In general, the use of HTTP instead of HTTPS can yield security risks for end users.  *(Even with the 30x redirects to associated HTTPS URIs, an adequately positioned attacker could theoretically tamper with the redirect itself, which also overs HTTP.)*

However, as outlined in the patch comments, I don't believe there is really much practical risk here; the threat scenario is pretty unlikely and any practical exploitation would be predicated on the existence of high-impact vulnerabilities in the projects' dependencies (e.g. Python, Shapely).

Therefore, this patch set is largely just intended to address this as a matter of security best practice, and encourage the explicit use of HTTPS moving forward. 

The initial reason for this patchset was that I happened to note the NaturalEarth zip file download being initiated using an `http://` URI.  However, the use of a file whitelist when extracting individual files mitigates the initial concern I had regarding any potential path traversal issues, prior to reading the code.

Since I was already grepping for additional occurrences of resource downloads via `http://` URIs, I applied similar changes throughout the code and documentation, where possible, and have attempted  to break up on a per-module or per-directory basis.

Unfortunately, a few hosts appear to have TLS certificate issues.  I've noted these in the patch comments, and have already reached out to osgeo.org regarding the certificate CN issue with vmap0.tiles.osgeo.org.

For some of links in the documentation, I could see a valid security-based argument for making the `https://` change and letting users complain to the associated website/domain owners to get their TLS  certificates updated according.  However, as a new Cartopy user, I don't think I'm familiar enough with the community to know whether or not this would just cause documentation usability issues/concerns.

## Implications

The documentation fails to build when attempting to access objects.inv for the Shapely documentation (now at `https://shapely.readthedocs.io/`).  However, prior to the `docs/source/conf.py` change, I see this already is failing due to a 404 on `http://toblerity.org/shapely/objects.inv`.

Beyond this, there are no adverse impacts with these changes, to the best of my knowledge.  I've attempted to confirm that all URI change is valid, and run tests/examples where I could.

Most, if not all, URIs changes within this patchset were already redirecting to an HTTPS URI, so I'd argue there's no "performance" argument to be made for not applying these changes.

## Additional Comments

I don't believe any of these patches constitute anything materially significant enough to involve "intellectual property" that needs to be assigned.

However, I'd be happy to get a signed CLA your way if there's any concerns.

**Update**: Submitted CLA .


jynik:
Pull request closed due to Sphinx documentation build failure. Reviewing this and will re-submit PR.

jynik:
Reopening due to the documentation build failure already being an issue (objects.inv no longer hosted on toblerity.org), regardless of the change made in fdb1c300880a1f53550ea0e5c0fe63e6963671f5.

Let me know if I should drop or further modify that patch.

lib/cartopy/_crs.pyx
```diff
@@ -546,7 +546,7 @@ class Geodetic(CRS):
         super(Geodetic, self).__init__(proj4_params, globe)
 
     # XXX Implement fwd such as Basemap's Geod. Would be used in the tissot example.
-    # Could come from http://geographiclib.sourceforge.net
+    # Could come from https://geographiclib.sourceforge.net
```
QuLogic: https links to SourceForge project pages should use .io.

PR #1314, opened 2019-05-31
Fix bug causing interior polygons to be ignored

stephenworsley:
When path_to_geos read a path with the codes ...,1,2,79,... it would create a LineString which would be appended to collection. This would cause further path_codes to be ignored. This was causing a bug with contourf where occasionally these codes will be passed. This fix would cause these LineStrings to be added to other_result_geoms instead. I have not yet explored the possible consequences of passing LineStrings to other_result_geoms.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -543,10 +543,17 @@ class Projection(six.with_metaclass(ABCMeta, CRS)):
                                            're-added. Please raise an issue.')
 
         # filter out any non-valid linear rings
+        def ring_validity(linear_ring):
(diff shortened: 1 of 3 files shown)


stephenworsley:
I messed the commit history up so I've made a new pull request here #1326

stephenworsley:
@dopplershift I've rebased in #1326 (the cpy of this one I had to open because I messed up with Git/GitHub) and now the checks are all passing. I'm having trouble getting GitHub to reopen this pull request now that I've tried force pushing the fixed, rebased branch on to here. Maybe it would make sense to transfer the milestones to that pull request?

PR #1317, opened 2019-06-05
Avoid Shapely bug causing segfaults

stephenworsley:
There is a bug in Shapely which causes a segfault when a `LineString` of length 3, with the first point equal to the last, is passed to `LinearRing`. For example, the following causes a segfault when the `ring` object is accessed:
``` python
from shapely.geometry.polygon import LinearRing, LineString
string = LineString([(1,0), (0,0), (1,0)])
ring = LinearRing(string)
print(ring)
````
When filtering out non-valid linear rings in `Projection._attach_lines_to_boundary()`, such cases will now also be filtered out.
It's worth noting that this bug appears to be fixed in the latest version of shapely:
https://github.com/Toblerity/Shapely/pull/638

### CODE REVIEW

**Comment by stephenworsley** [comment]
For reference, #1047 and #1090 were concerned with the same problem. It seems that #1047 was considered to broad but the fix in #1090 was slightly too narrow. This fix is somewhere in between, catching more bugs but not discarding any potential non-zero area polygons.

File: lib/cartopy/tests/test_polygon.py [hunk 1]
```diff
@@ -259,6 +259,21 @@ def test_same_points_on_boundary_2(self):
...
+        # Geometry comes from a matplotlib contourf.
+        mstring = shapely.wkt.loads(
+            'MULTILINESTRING ('
+            '(-179.9999982118607 71.87500000000001,'
+            '-179.0625 71.87500000000001,'
+            '-179.9999982118607 71.87500000000001))')
+        multi_line_strings = [mstring]
+
+        src = ccrs.PlateCarree()
+        polygons = src._attach_lines_to_boundary(multi_line_strings, True)
```
dopplershift: Is there any way to do this without invoking an internal API? Leaving the test as is would hinder future attempts at refactoring and changing the non-public API.

stephenworsley: So here's my thinking behind writing the test as it is:
If I were to call the nearest public method (project_geometry) then on failure, the problematic shapely object will be called and this will cause a segfault. As it is written, the problematic shapely object is never called and the test fails entirely within python.
Also, this file already calls this method at line [137](https://github.com/SciTools/cartopy/pull/1317/files#diff-67bdbbb09e12e0d9d12703599512761cL137).

That said, I ought to be able to write a test (whose failure state is a segfault) which calls the public project_geometry method if that is preferable.

dopplershift: Can you add the segfault test in addition? That way if we refactor, we can just ditch the old test but ensure that we're still testing the original use case and not breaking it.

stephenworsley: Okay, I've added the second test.

trexfeathers:
Now that the tests are passing, is anyone with the expertise available to review this PR? Thanks 🙂

lib/cartopy/crs.py
```diff
@@ -529,10 +529,21 @@ def filter_last(t):
                                            're-added. Please raise an issue.')
 
         # filter out any non-valid linear rings
+        def ring_validity(linear_ring):
```
QuLogic, line 532: The input here is a `LineString` before converting to `LinearRing`, so I find this naming confusing (and yes, the original list comprehension was confusingly named too.)

stephenworsley: I've changed the name to `line_string`, which ought to better reflect what's going on.

dopplershift: I think the comment also about the name of the function being `ring_validity` when it's working with a line string.

PR #1354, opened 2019-08-16
Tight bounding box gridliner bugfix

lukelbd:
## Rationale

The `matplotlib.axes.Axes.get_tightbbox` algorithm fails to include cartopy gridliner labels. This is because `cartopy.mpl.gridliner.Gridliner._draw_gridliner` is not called until `Axes.draw` is invoked, which comes **after** `Figure.draw` invokes `Figure.get_tight_layout`. 

This change adds a `get_tightbbox` override to the `GeoAxes` class, which calls `_draw_gridliner` on all the `GridLiner` objects, so the tight layout algorithm will work. This mimics the approach used in the matplotlib API: for example, `Axes.draw` and `Axes.get_tightbbox` contain many identical calls to various post-processing steps, like `_update_title_position`.

## Implications

You can now use `fig.tight_layout()` on figures with cartopy axes containing gridliner labels.

## Examples

I wasn't able to import my clone of cartopy (got `ModuleNotFoundError: No module named 'cartopy._crs'` after running `import sys; sys.path.insert(0, '/Users/username/cartopy/lib`). But I implemented a virtually identical change in my [proplot package](https://github.com/lukelbd/proplot), which automatically applies `fig.tight_layout` to figures. The first image is without the `get_tightbbox` override -- there are labels present, they are just off the edge of the figure. The second example is with the `get_tightbbox` override.

```python
import proplot as plot
f, ax = plot.subplots(proj='cyl', axwidth=3)
ax.format(labels=True, lonlines=60, latlines=30, title='Rectilinear projection')
```

A similar `pyplot` example could be generated by just calling `fig.tight_layout()` on any cartopy plot with gridlines.

example1
example2


lukelbd:
I accidentally included other edits in this PR. Submitting a new one.

PR #1355, opened 2019-08-16
Include gridliner labels in tight bbox calculation

lukelbd:
## Rationale

`matplotlib.axes.Axes.get_tightbbox` fails to include cartopy gridliner labels. This is because `cartopy.mpl.gridliner.Gridliner._draw_gridliner` is not called until `Axes.draw` is invoked, which comes *after* `Figure.draw` invokes `Figure.tight_layout`. 

This change adds a `get_tightbbox` override to the `GeoAxes` class, which calls `_draw_gridliner` on all the `GridLiner` objects, so the tight layout algorithm will work. This mimics the approach used in the matplotlib API: `Axes.draw` and `Axes.get_tightbbox` contain many identical calls to various post-processing steps, like `_update_title_position`.

## Implications

You can now use `fig.tight_layout()` on figures with cartopy axes containing gridliner labels.

## Examples

I wasn't able to install a development version of cartopy, but I implemented a virtually identical change in a `GeoAxes` subclass in my [proplot](https://github.com/lukelbd/proplot) package (proplot calls `fig.tight_layout()` on all figures by default).

Below, the first example is without the `get_tightbbox` override (there are labels present, they are just off the edge of the figure). The second example is with the `get_tightbbox` override.

```python
import proplot as plot
fig, ax = plot.subplots(proj='pcarree', axwidth=3)
ax.format(labels=True, lonlines=60, latlines=30, title='Plate Carrée')
fig.save('example.png')
```

A similar `pyplot` example could be generated by just calling `fig.tight_layout()` on any cartopy plot with gridlines.

example1
example2


lib/cartopy/mpl/geoaxes.py
```diff
@@ -392,6 +392,16 @@ def hold_limits(self, hold=True):
...
                     self._autoscaleXon, self._autoscaleYon) = other
 
+    def get_tightbbox(self, *args, **kwargs):
+        """Draw gridliner objects to tight bounding box algorithm will
+        incorporate gridliner labels."""
+        # Matplotlib approaches this problem in the same way, by duplicating
+        # post-processing steps in both Axes.draw and Axes.get_tightbbox
+        for gl in self._gridliners:
+            gl._draw_gridliner(background_patch=self.background_patch)
+        self._gridliners = []
```
dopplershift: I may not understand the rest of CartoPy well enough, but why are you resetting `self._gridliners`?

lukelbd: This is because `draw` and `get_tight_bbox` can be called more than once (for example, if the user calls `Figure.show` more than once, or if an interactive backend automatically updates figure changes). Without `self._gridliners = []` they would be drawn multiple times. Note this code block is copied directly from `GeoAxes.draw`.

greglucas:
@lukelbd did you sign the CLA? Could you rebase if so to rerun CI on this.

lukelbd:
@greglucas Sorry for the delay; I've signed the [CLA](https://docs.google.com/forms/d/e/1FAIpQLSfd0tdE-DcJOXh8ej_7T93IizwJFYBFyRWYQOi2A8QRaKwykA/viewform), and can try to add a test by next weekend.

greglucas:
Sounds good, @lukelbd! @dopplershift was there anything else you were requesting in addition to the test?

dopplershift:
@greglucas I think at this point it just needs a test.

greglucas:
ping @lukelbd.

greglucas:
@lukelbd we are going to try and push another beta release out this weekend. Do you think you can add the test by then, or should we look to push this back to the next 0.19 release milestone?

QuLogic:
Tagging the rc today, so I'm pushing this back to 0.19.

greglucas:
I just ran into the case of saving a pdf and the side grid labels were partially cut off. I think this may solve that issue...? It would be good to get it into the code since it seems logical enough.

lukelbd:
Sorry forgot about this PR, was really busy when I was pinged. Will take a look this weekend

lukelbd:
I added a test for the "latest everything" build only. Turns out matplotlib 1.5.1 does not use `Figure.get_tightbbox()` to calculate the bounding box -- it uses a `get_figure_tight_layout()` function in `tight_layout.py` instead. So it is impossible to override the tight layout algorithm:

https://github.com/matplotlib/matplotlib/blob/be91fac9fa2b7250080557e723af75124659da4e/lib/matplotlib/figure.py#L1720-L1756

My test is similar to `test_grid_labels`, but it calls `Figure.tight_layout()` at the end and ensures that every gridliner was drawn by `get_tightbbox()` by ensuring `Gridliner._plotted == True`.

Here is the baseline result:

image

And here is the result without the `Figure.tight_layout()` call:

result-gridliner_labels_tight

lukelbd:
Not sure why it's failing... I built the test image by running `pytest` as follows:


```bash
conda config --set always_yes yes --set changeps1 no --set show_channel_urls yes
conda config --add channels conda-forge
conda config --add channels conda-forge/label/testing
conda config --set restore_free_channel true

export CYTHON_COVERAGE=1
echo "backend: agg" > ~/.matplotlib/matplotlibrc  # macOS matplotlibrc

conda create -n cartopy-dev python=3.6 cython numpy matplotlib-base proj4 pykdtree scipy fiona pillow pytest pytest-xdist filelock pep8 pyshp shapely six requests pyepsg owslib pytest-cov coveralls
conda activate cartopy-dev

pip install --no-deps -e .
python -c "import cartopy; print('Version ', cartopy.__version__)" && python setup.py version
purest
```

lukelbd:
Nevermind -- looks like the latest push fixed it! I might have accidentally generated the baseline image without the `backend: agg` line or in the matplotlib 1.5.1 environment.

Anything else I need to do?

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -234,6 +238,50 @@ def test_grid_labels():
     plt.subplots_adjust(wspace=0.25, hspace=0.25)
 
 
+# Matplotlib 1.5.1 uses tight_layout.py/get_tight_layout_figure rather
+# than figure.py/Figure.get_tightbbox. Impossible to override algorithm.
+if MPL_VERSION >= '2.0':
```
greglucas, line 243: You should use `@pytest.mark.skipif()` here.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -129,6 +129,7 @@ def test_gridliner_specified_lines():
 grid_label_tol = grid_label_inline_tol = grid_label_inline_usa_tol = 0.5
 if MPL_VERSION >= '2.0':
     grid_label_image = 'gridliner_labels'
+    grid_label_tight_image = 'gridliner_labels_tight'
```
greglucas, line 132: This could just be put in the image comparison directly I think since there isn't another version at all.

lukelbd:
@greglucas Done!

lukelbd:
@dopplershift I think the remaining requested change is out of date

PR #1361, opened 2019-09-04
fix issue #1341 by adding a user_agent when requesting tiles from a provider.

htonchia:
## Rationale

It was no more possible to get tiles from OSM with the current code, as OSM was expecting a user agent in the tile request.
Furthermore, when the server could not be reached for network issue for example, the error would stop the program.
## Implications

There is a variable for the user agent and the possibility to set it at init, so there is a new parameter.

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -51,10 +51,14 @@ class GoogleWTS(six.with_metaclass(ABCMeta, object)):
     """
     _MAX_THREADS = 24
 
-    def __init__(self, desired_tile_form='RGB'):
+    def __init__(self, desired_tile_form='RGB', user_agent='cartopybot/1.0'):
         self.imgs = []
         self.crs = ccrs.Mercator.GOOGLE
         self.desired_tile_form = desired_tile_form
+        self.user_agent = user_agent
+        # some providers like osm need a user_agent in the request issue #1341
+        # osm may reject requests if there are too many of them, in which case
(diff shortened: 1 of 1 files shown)


htonchia:
Hi
From what I can tell, the errors are not related to the part I have modified. So what can i do ?

QuLogic:
Don't worry about the test failures; as long as `test_web_tiles` is working again, this is fine.

QuLogic:
Can you rebase to fix conflicts, and preferably squash this into one commit?

lib/cartopy/io/img_tiles.py
```diff
@@ -175,18 +179,23 @@ def _image_url(self, tile):
...
+            request = Request(url, headers={"user-agent": self.user_agent})
+            fh = urlopen(request)
+            im_data = six.BytesIO(fh.read())
+            fh.close()
+            img = Image.open(im_data)
+
+            img = img.convert(self.desired_tile_form)
+        except (HTTPError, URLError) as err:
+            print(err)
+            return (np.full((256, 256, 3), (252, 248, 248)),
```
QuLogic: This causes a new warning:
```
numpy/core/numeric.py:294: FutureWarning: in the future, full(..., (252, 248, 248)) will return an array of dtype('int64')
    format(fill_value, array(fill_value).dtype), FutureWarning)
```
I think you need to specify an explicit dtype to silence this.

Also, if `desired_tile_form` is `RGBA` (or `ARGB`?), then it should return a different shape.

PR #1380, opened 2019-10-08
wip: no quick transform, when there is no vertices

guziy:
## Rationale

The tricontourf fails with the dataset I have, because the `vertices` parameter in `quick_vertices_transform` is empty.

## Checklist

 * [ ] If you have not already done so, ensure you've read and signed the Contributor Licence Agreement (CLA).
   (See the [governance page](http://scitools.org.uk/governance.html) for the CLA and what to do with it).

 
 * [ ] Ensure there is a suitable item in the cartopy test suite for the change you are proposing. 

      I am not able to run the tests.

     ```
         ImportError while importing test module 
         '../Python/cartopy/lib/cartopy/tests/mpl/test_web_services.py'.
         Hint: make sure your test modules/packages have valid Python names.
        Traceback:
          __init__.py:95: in <module>
         import cartopy.crs
         crs.py:36: in <module>
        from cartopy._crs import (CRS, Geodetic, Globe, PROJ4_VERSION,
      E   ModuleNotFoundError: No module named 'cartopy._crs'
     ```

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1250,7 +1250,7 @@ class Projection(CRS, metaclass=ABCMeta):
         if vertices.size == 0:
             return vertices
 
-        if self == src_crs:
+        if self == src_crs and len(vertices) > 0:
             x = vertices[:, 0]
             y = vertices[:, 1]
             # Extend the limits a tiny amount to allow for precision mistakes
@@ -1371,6 +1371,10 @@ class PlateCarree(_CylindricalProjection):
 
     def quick_vertices_transform(self, vertices, src_crs):
         return_value = super().quick_vertices_transform(vertices, src_crs)
+
+        # if there is no vertices
+        if len(vertices) == 0:
+            return return_value
 
         # Optimise the PlateCarree -> PlateCarree case where no
         # wrapping or interpolation needs to take place.


QuLogic:
What is the error that this is fixing?

You will need an editable install of Cartopy in order to run tests.

guziy:
> What is the error that this is fixing?
> 

It fixes the IndexError and allows to plot my data. I need to make a test yet for this.

> You will need an editable install of Cartopy in order to run tests.

At some point, I was able to run the tests, but now I have to remember how I did it.

rcomer:
These changes look very similar to part of #2213.  @guziy are you able to reproduce the problem with Cartopy v0.22?  I appreciate that some time has passed, so it’s fair enough if you don’t have the means to check any more!

guziy:
Sorry @rcomer, I can't find the test code... I should have posted it....((

lib/cartopy/crs.py
```diff
@@ -1250,7 +1250,7 @@ def quick_vertices_transform(self, vertices, src_crs):
         if vertices.size == 0:
             return vertices
 
-        if self == src_crs:
+        if self == src_crs and len(vertices) > 0:
```
rcomer, line 1253: Because of the lines 1250-1251, which were added at #2213, I don’t think this check will ever be exercised.

rcomer (commented):
Thanks for looking @guziy.  Given how similar this change is to part of #2213, I strongly suspect that whatever the problem was is now fixed in the most recent release.  So I’m going to close this for now.  Please do get back in touch if you have further problems though.

PR #1396, opened 2019-11-14
Update installation instructions

MarkWieczorek:
This pull request attempts to clean up and simplify the Cartopy installation instructions. In addition, detailed instructions are given for macOS using brew and pip.

Though I think that this is better than before, it is not perfect by any means. I won't be offended if this is not accepted... In particular:
* I question whether we need the "other pre-built binaries" section.
* I think that the detailed list of required and optional dependencies is too long. In my opinion, it would be preferable to instead describe how to install these on Linux.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -4,8 +4,9 @@ Installing Cartopy
 Conda pre-built binaries
 ------------------------
 
-The easiest way to install Cartopy is through 
-`Conda <https://conda.io/miniconda.html>`_. If conda is already installed, installation is as easy as::
+The easiest way to install Cartopy is by using
+`Conda <https://conda.io/miniconda.html>`_. If conda is already installed,
+installation is as easy as::
 
     conda install -c conda-forge cartopy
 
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -20,98 +24,99 @@ Additional options include:
...
-    Cartopy requires Python 2.7 or later.
 
 **Cython** 0.28 or later (https://pypi.python.org/pypi/Cython/)
 
 **NumPy** 1.10 or later (http://www.numpy.org/)
-    Python package for scientific computing including a powerful N-dimensional
-    array object.
 
 **GEOS** 3.3.3 or later (https://trac.osgeo.org/geos/)
-    GEOS is an API of spatial predicates and functions for processing geometry
```
dopplershift: Why remove all of these descriptions?

dopplershift:
Overall, this looks like a good improvement. Just the one question.

MarkWieczorek:
Feel free to put them back or modify this in any way you want. This was just a suggestion.

In any case, the reason for me is: I want to install Cartopy, and I don't care what the dependencies are or do. In an ideal world, the dependencies would be installed automatically without me needing to know about it.

INSTALL
```diff
@@ -20,98 +24,99 @@ Additional options include:
...
 
 **SciPy** 0.10 or later (https://www.scipy.org/)
-    Python package for scientific computing.
+    A Python package for scientific computing.
 
 **OWSLib** 0.8.7 (https://pypi.python.org/pypi/OWSLib)
-     Python package for client programming with Open Geospatial Consortium
-     (OGC) web service.  Gives access to cartopy ogc clients.
+     A Python package for client programming with the Open Geospatial
+     Consortium (OGC) web service, and which gives access to cartopy ogc
```
greglucas: capitalize Cartopy here since it is elsewhere in the file too.

MarkWieczorek:
I've added all the package descriptions back to INSTALL and made the requested minor changes. If there is anything else, please let me know.

INSTALL
```diff
@@ -1,16 +1,20 @@
...
 
     conda install -c conda-forge cartopy
 
-Additional options include:
+
+Other pre-built binaries
+------------------------
+
+Additional pre-built binaries can be found at a variety of sources, including:
  * `Enthought Canopy <https://www.enthought.com/product/canopy/>`_.
```
gepcel: I couldn't find a pre-built binaries here:    `Enthought Canopy <https://www.enthought.com/product/canopy/>`

dopplershift: Looks like Enthought re-branded or something. They now have the [Enthought Deployment Manager](http://docs.enthought.com/edm/), but I can't find anything on the binaries available.

MarkWieczorek: I just removed the link to Enthought.

greglucas:
@MarkWieczorek have you signed the CLA now? If so, could you rebase this to rerun the CI and get the tests passing?

MarkWieczorek:
Rebase and CLA signed in https://github.com/SciTools/cartopy/pull/1480

QuLogic:
It wasn't really necessary to open a new PR, but that's fine.

PR #1408, opened 2019-12-05
Create wms_ecmwf.py

htonchia:
Example on how to manage the geographical boundary extent and color and "token" for wms server.

## Rationale

I have added an example using data from a wms server.
It shows how to send extra information needed to access a server, in the example, a 'token'.

It also show how to manage the boundary of the data, and how to control its extent and how to change its color or remove it.
For the color, this is a workaround as there is no function yet to do it. 

## Implications

There have been some changes in how the boundary is managed from an outline_patch to a spines['geo']. The example handles both.

### DIFF
--- a/lib/cartopy/examples/wms_ecmwf.py
+++ b/lib/cartopy/examples/wms_ecmwf.py
@@ -1,0 +1,123 @@
+"""
+WMS (Web Map Service): managing the boundary and token to the wms server
+------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


QuLogic:
I fear this example may be doing a bit too many things at the same time. I don't really understand the working of `full_geo_extent`, and it may be easier to do now that tight layout works better. The docs should be for the current version, so the edge colour stuff can be simplified.

Issue #1386, opened 2019-10-17
Trouble rendering style color palettes in various projections using wms in cartopy


hevgyrt:
### Description

I try to plot data from wms using cartopy. In particular, I would like to both change color style and projection for some applications. I am able to change style using PlateCaree as projection. However, when I try other projections in ccrs like Mercator() the color palette is always gray scale. Am I doing something wrong?

#### Code to reproduce

```
import cartopy.crs as ccrs
import cartopy
import matplotlib.pyplot as plt
from owslib.wms import WebMapService

url = "http://thredds.met.no/thredds/wms/sea/arctic20km/1h/aggregate_be?service=WMS&version=1.3.0&request=GetCapabilities"
wms = WebMapService(url,timeout=80)

layer='temperature'
style = ['boxfill/redblue']

projections = [ccrs.Mercator(), ccrs.PlateCarree()]
for projection in projections:
    subplot_kw = dict(projection=projection)
    fig, ax = plt.subplots(subplot_kw=subplot_kw)
    
    ax.add_wms(wms, layer,
	       wms_kwargs={'transparent': False,
               'styles':style})
    print(wms.contents[layer].styles[style[0]])
    plt.show()


```

#### Traceback 

```

```

<details>
 <summary>Full environment definition</summary>

### Operating system
Linux ubuntu 18.04

### Cartopy version
'0.17.0'

### conda list

```
```

### pip list

```
```

</details>

htonchia:
You did nothing wrong.

Well, it seems that when the image sent by the server is not an RGB(A) image, the color palette is lost in the regrid process to the target projection.

If one convert the image from the server to RGBA, one gets:
![MercatorBlueRed](https://user-images.githubusercontent.com/54799451/70455711-0b54e980-1aad-11ea-8136-92db1c4188ba.png)
I guess it is what you expected.

it is fixed by:         
```
# convert to RGBA to keep the color palette
        img, extent = warp_array(np.asanyarray(image.convert('RGBA'))[::-1]
```
in io/ogc_clients.py

greglucas:
Closed by #1473

PR #1413, opened 2019-12-09
Update ogc_clients.py issue #1386 color palette

htonchia:
Fix issue #1386 when the color palette of the wms file is lost during the regrid process.

## Rationale

When the image fetched from the wms server is not in the expected projection, it is regridded.
When the source image is not RGB or RGBA, it looses its color palette in the regrid process.
To fix this, I have converted the source image to RGBA before the regrid process.

## Implications

 I am not aware of any implication.
There would be some implication if in some place cartopy regrids non RGBA image and need those to stay non RGBA.

In the following example, when it works, both images have a "blue/red" palette, otherwise the reprojected image, the Mercator one, is a viridis colored image.
As those are images live from a server, they change with time.
```
import cartopy.crs as ccrs

URL = "http://thredds.met.no/thredds/wms/sea/arctic20km/1h/aggregate_be?\
service=WMS&version=1.3.0&request=GetCapabilities"
WMS = WebMapService(URL, timeout=80)
LAYER = 'temperature'
WMS_KWARGS = {'transparent': False,
              'styles': ['boxfill/redblue']}
PROJECTIONS = [ccrs.Mercator(),
               ccrs.PlateCarree()]

for projection in PROJECTIONS:
    subplot_kw = dict(projection=projection)
    ax.add_wms(WMS, LAYER, WMS_KWARGS)
    plt.title(projection.proj4_params['proj'])
```
MercatorBlueRed

### DIFF
--- a/lib/cartopy/io/ogc_clients.py
+++ b/lib/cartopy/io/ogc_clients.py
@@ -137,19 +137,10 @@ def _warped_located_image(image, source_projection, source_extent,
         # This avoids unsightly grey boundaries appearing when the
         # extent is limited (i.e. not global).
         if np.ma.is_masked(img):
-            if img.shape[2:3] == (3,):
-                # RGB
-                old_img = img
-                img = np.zeros(img.shape[:2] + (4,), dtype=img.dtype)
-                img[:, :, 0:3] = old_img
-                img[:, :, 3] = ~ np.any(old_img.mask, axis=2)
-                if img.dtype.kind == 'u':
-                    img[:, :, 3] *= 255
-            elif img.shape[2:3] == (4,):
-                # RGBA
-                img[:, :, 3] = np.where(np.any(img.mask, axis=2), 0,
-                                        img[:, :, 3])
-                img = img.data
+            # RGBA
+            img[:, :, 3] = np.where(np.any(img.mask, axis=2), 0,
+                                    img[:, :, 3])
+            img = img.data
 
         # Convert warped image array back to an Image, undoing the
         # earlier flip.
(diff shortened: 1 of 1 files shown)


lib/cartopy/io/ogc_clients.py
```diff
@@ -122,7 +122,8 @@ def _warped_located_image(image, source_projection, source_extent,
     else:
         # Convert Image to numpy array (flipping so that origin
         # is 'lower').
-        img, extent = warp_array(np.asanyarray(image)[::-1],
+        # convert to RGBA to keep the color palette
+        img, extent = warp_array(np.asanyarray(image.convert('RGBA'))[::-1],
```
greglucas: Does adding this change how RGB (without alpha) is read/used?

Below this code (Line 140) there are tests for the shape of img if there is an alpha layer or not. I assume with this change it would always return an RGBA img and not go through the other branch?

QuLogic: Do you mean 'test' as in unit test, or as in the conditional below, i.e., is this PR complete now?

QuLogic:
Can you rebase to get fixed CI? Also, perhaps squash those commits and give them a better description.

greglucas:
ping @htonchia.

lib/cartopy/io/ogc_clients.py
htonchia: I meant the conditional.
The PR is complete.

htonchia: I just added some comments.

htonchia:
Replaced by #1473  to take into account comments and fix CI so that it passes the checks.

PR #1417, opened 2019-12-13
Fix most build issues

QuLogic:
## Rationale

This fixes CI and most of the tests against all versions of Proj, and latest micro release of every Matplotlib release. The only thing left is the OWSLib failures, but upstream does not seem to have figured that one out.

## Implications

Mostly bug fixes and increased tolerances on image tests. There are several workarounds for buggy versions, but they should be limited to only the versions of deps that are broken.

There's one bit of new functionality in 7c7488e to allow working with both TransversMercator and ExtendedTransverseMercator (since Proj switched implementations.)

See commit messages for fixed issue numbers.


QuLogic:
I fixed a couple other issues with docs now, too.

@SciTools/cartopy-admins what do we want to do about the WMTS failures? It doesn't seem like OWSLib has made any progress.

ajdawson:
I think set them to XFAIL or SKIP so we can continue without them.

QuLogic:
OK, I've had to tweak the doc build a bit, but now I've added the necessary xfail's to the examples and tests so CI is passing now.

dopplershift:
I agree with @ajdawson .

lib/cartopy/crs.py
```diff
@@ -1496,7 +1528,7 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
             if (5, 0, 0) <= PROJ4_VERSION < (5, 1, 0):
                 warnings.warn(
                     'The Orthographic projection in Proj between 5.0.0 and '
-                    '5.1.0 incorrectly transforms points. Use this projection '
+                    '5.0.1 incorrectly transforms points. Use this projection '
```
dopplershift: This message doesn't agree with the version check above.

QuLogic: I've changed the message to be closer to what the Robinson one is, so it should be clearer.

QuLogic: I've changed the message to be closer to what the Robinson one is, so it should be clearer.

QuLogic:
No, I don't think we should do that, because Proj6 makes `tmerc` into `etmerc`, so it will be more confusing to call it `etmerc` in the future. The meaning is [described in the docstring](https://github.com/SciTools/cartopy/pull/1417/files#diff-6dfcada343269e6678a777c4f9a5223dR826-R829), though I could add a reference to [Proj's parameter](https://proj.org/operations/projections/tmerc.html#parameters) if it helps.

greglucas:
I agree with your reasoning, especially considering `+approx` is what is sent on to PROJ.

Another thought is whether this should even be exposed to the Cartopy user as a keyword option at all? i.e. What if you just added this option to your current PROJ version checks in `TransverseMercator` and say approx if False if PROJ < 6, and approx is True for PROJ > 6? To me, it just seems like a lot of extra weight to carry around for not much additional benefit other than being able to use a less-accurate, but faster transform (which likely is not the bottle-neck in Python plotting is my guess).

I'm 👍 as is though and I would suggest merging this in before beginning on the other PR's so that CI will be working again.

ajdawson:
OK I think we have consensus that this is good to go. Thanks a lot @QuLogic.

Issue #1416, opened 2019-12-12
pcolormesh wrapping not working: tiny bug


htonchia:
### Description

There is a patch in pcolormesh to avoid overlapping cells crossing the plots when they cross a boundary.
in this patch the overlapping cells are masked and plotted with a zorder not low enough to mask them.

The data in the examples below are ice concentrations on a rectangular polar stereographic grid.

The figure below shows that a lower zorder (minus missing) masks the overlapping cells while the one in the current code doesn't.
The current code correspond to the plots on the right and the modified code to the plots on the left.
There is a second issue, the cells are masked is they span over half the limit of the projection, but overlapping cells may span a smaller area, when the projection is not rectangular, like Mollweide.
We can see that there are some overlapping cells left on the left plot in Mollweide projection.

![pcolormesh_wrap_nowrap_zorder](https://user-images.githubusercontent.com/54799451/70753954-4cabfa00-1d36-11ea-95aa-519453875cdb.png)

This relates to other issues, #1151 and probably #323 and #976 

**There are two ways to fix this:**

- changing the zorder when plotting the overlapping cells so that they do no show
- not plotting altogether those overlapping cells and comment the lines that plot them

in the second case, the idea of keeping the code in comments is that someone may want to split those cells and rewrap them to plot them the right way.
They both work.

The process already implemented in pcolormesh in GeoAxes could be implemented to fix the same issue with pcolor, contour and contourf.
On the left, if we mask the overlapping cells, on the right the current code.

![Mollweide_0_wrap_nowrap](https://user-images.githubusercontent.com/54799451/70756158-140f1f00-1d3c-11ea-8679-de73f26e627e.png)

#### Code to reproduce

```
import numpy as np
import numpy.ma as ma
import matplotlib.pyplot as plt
from netCDF4 import Dataset
import cartopy.crs as ccrs

# file available here:
# https://github.com/htonchia/cartopy-plotting-examples/blob/master\
#/ice_conc_nh_polstere-100_multi_201912091200.nc
FILE = "ice_conc_nh_polstere-100_multi_201912091200.nc"
VARNAME = 'ice_conc'
STEP = 1  # chose 20 or 40 to fasten plots

def projection4test():
    """
    define the lists of projection for the tests
    """

    proj2s = {}
    proj2s['PlateCarre_90'] = ccrs.PlateCarree(central_longitude=90)
    proj2s['Robinson_0'] = ccrs.Robinson(central_longitude=0)
[...]

Issue #1333, opened 2019-06-25
PROJ Mollweide projection singularity at the poles


stephenworsley:
### Description

The [PROJ Mollweide projection](https://github.com/OSGeo/PROJ/blob/master/src/projections/moll.cpp) will map all points above a certain latitude (somewhere between ±89.144 and ±89.145 degrees) directly to the pole. This seems to be due to Mollweide using an iterative algorithm which is slow to converge at the poles.
As a result, this causes certain geometries to be incorrectly projected. In particular, this will affect contourf plots which use data above 89 degrees, with some polygons incorrectly being projected onto almost the entire map, blotting out previously plotted polygons.

Below is a polygon taken from a contourf. When projected, it gets 'turned inside out'.

#### Code to reproduce

```
import cartopy
import shapely.wkt
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

polygon = shapely.wkt.loads('POLYGON ((-124.6875 88.51576898014021, -124.6875 89.375, -126.5625 89.375, -128.4375 89.375, -130.3125 89.375, -132.1875 89.375, -134.0625 89.375, -135.9375 89.375, -137.8125 89.375, -139.6875 89.375, -141.5625 89.375, -143.4375 89.375, -145.3125 89.375, -147.1875 89.375, -149.0625 89.375, -150.9375 89.375, -152.8125 89.375, -154.6875 89.375, -156.5625 89.375, -158.4375 89.375, -160.3125 89.375, -162.1875 89.375, -164.0625 89.375, -165.9375 89.375, -167.8125 89.375, -169.6482699764407 89.375, -167.8125 88.48909844616483, -165.9375 88.57715321661044, -164.0625 88.66842665037734, -162.1875 88.56738315818514, -160.3125 88.75973810567709, -158.4375 88.61867874791304, -156.5625 88.48770835812856, -154.6875 88.73331726616871, -152.8125 88.6420385333048, -150.9375 89.20007552152518, -149.0625 88.46018604660821, -147.1875 88.42282488713046, -145.3125 88.39871713455433, -143.4375 88.9979823936324, -141.5625 88.60989689925717, -139.6875 88.86032315648676, -137.8125 88.58963849370639, -135.9375 88.63502159672632, -134.0625 88.53263139067644, -132.1875 88.75896594337623, -130.3125 88.76036263098899, -128.4375 88.79108697623622, -126.5625 88.52046382385734, -124.6875 88.51576898014021))')

[...]

PR #1418, opened 2019-12-13
Update geoaxes.py pcolormesh overlapping cells plot issue #1416

htonchia:
Commented out the part plotting the overlapping cells (despite a smaller zorder).
Now the overlapping cells larger than half the projection span stay masked.

## Rationale

There is already a patch in pcolormesh that masks overlapping cells which are the cells crossing the longitude boundary and therefore overlapping the plot. 
In that patch, the overlapping cells, those who size is more than half the projection span in x, are masked and plot then again with a zorder supposedly small enough so that they don't appear.
Unfortunately they still appeared on the plot.
The idea in plotting the bad cells is that someone may want to use them, by splitting them at the boundary and plot the split part. But they shouldn't be plotted as there are.

So the fix is to comment out the part plotting the bad cells.

This patch works only for pcolormesh, there should be one of the same kind to mask the data generating overlapping cells or contours in  contourf, pcolor and contour.

This patch removes the cell greater than half the projection span in x, which works well for Plate Carree but not for the narrow part of a projection like Mollweide. This part is not fixed.

The image below displays ice concentration provided in a polar stereographic grid.
It had been plotted after the proposed fix.
The left plots are pcolormesh which includes the patch for overlapping cells, the right plots are contourf and displays many overlapping cells.
The pcolormesh plot in Mollweide projection still displays some small overlapping cells (less than half the projection span) on the narrow part of the plot.

There are no overlapping cells in the stereographic projection.
pcolormesh_contourf

The data can be found [here](https://github.com/htonchia/cartopy-plotting-examples/blob/master/ice_conc_nh_polstere-100_multi_201912091200.nc).

```
from netCDF4 import Dataset
import cartopy.crs as ccrs

# file available here:
# https://github.com/htonchia/cartopy-plotting-examples/blob/master\
#/ice_conc_nh_polstere-100_multi_201912091200.nc

FILE = "ice_conc_nh_polstere-100_multi_201912091200.nc"
VARNAME = 'ice_conc'
STEP = 1  

def projection4test():
    """
    define the list of projection for the tests
    """
#    proj_data = ccrs.Stereographic(
#        central_longitude=-45, central_latitude=90,
#        true_scale_latitude=70)

    proj2s = {}
    proj2s['PlateCarre_90'] = ccrs.PlateCarree(central_longitude=90)
    proj2s['Robinson_0'] = ccrs.Robinson(central_longitude=0)
    proj2s['Mollweide_90'] = ccrs.Mollweide(central_longitude=90)
    proj2s['polar 45'] = ccrs.Stereographic(
        central_longitude=45, central_latitude=90,
        true_scale_latitude=70)
    
    return proj2s

def main(data, varname, step=1):
    """
    Run plotting tests on different set of projections
    """

    lats = data.variables['lat'][::step, ::step]
    lons = data.variables['lon'][::step, ::step]

    projs = projection4test()

    dplotkwargs = {}
    dplotkwargs['pcolor'] = {'cmap' : 'bwr'}
    dplotkwargs['pcolormesh'] = {'cmap' : 'bwr'}
    dplotkwargs['contourf'] = {'cmap' : 'bwr'}
    dplotkwargs['contour'] = {'colors' : 'black', 'linewidths' : 1}

    opes = ['pcolormesh', 'contourf']

    plt.figure(figsize=(12, 6.4), dpi=100)
    plt.subplots_adjust(wspace=0.01, left=0.01, right=0.99, top=0.95,
                        hspace=0.01, bottom=0.01)

    iii = 0
    
    plt.suptitle("{} {}".format(*opes))
    for proj in projs:
        for ope in opes:
            iii += 1
            mdata = data.variables[varname][0, ::step, ::step]
            axe = plt.subplot(4, 2, iii,
                              projection=projs[proj])
            axe.set_title("{} {}".format(proj, ope))
            if (hasattr(projs[proj], 'proj4_params') and
                    projs[proj].proj4_params['proj'] in
                    ['geos', 'nsper', 'gnom']
                    and ope == 'pcolormesh'):
                # will crash if we try to plot
                axe.coastlines()
                continue
            xy2 = axe.projection.transform_points(
                    ccrs.Geodetic(),
                    lons, lats)
            cx2, cy2 = xy2[..., 0], xy2[..., 1]
            getattr(axe,ope)(cx2, cy2, mdata, **dplotkwargs[ope])
            axe.coastlines()

    plt.savefig("img/{}_{}.png".format(*opes),
                    dpi=200, pad_inches=0)


if __name__ == '__main__':
    main(Dataset(FILE), varname=VARNAME, step=STEP)


```
## Implications

I don't see any implication.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1720,36 +1720,47 @@ class GeoAxes(matplotlib.axes.Axes):
 
                     collection.set_array(pcolormesh_data.ravel())
 
-                    # now that the pcolormesh has masked the bad values,
-                    # create a pcolor with just those values that were masked
-                    if C_mask is not None:
-                        # remember to re-apply the original data mask
-                        pcolor_data = np.ma.array(C, mask=~mask | C_mask)
-                    else:
-                        pcolor_data = np.ma.array(C, mask=~mask)
-
-                    pts = coords.reshape((Ny, Nx, 2))
-                    if np.any(~pcolor_data.mask):
-                        # plot with slightly lower zorder to avoid odd issue
-                        # where the main plot is obscured
-                        zorder = collection.zorder - .1
-                        kwargs.pop('zorder', None)
-                        kwargs.setdefault('snap', False)
-                        pcolor_col = self.pcolor(pts[..., 0], pts[..., 1],
-                                                 pcolor_data, zorder=zorder,
-                                                 **kwargs)
-
-                        pcolor_col.set_cmap(cmap)
-                        pcolor_col.set_norm(norm)
-                        pcolor_col.set_clim(vmin, vmax)
-                        # scale the data according to the *original* data
-                        pcolor_col.norm.autoscale_None(C)
-
-                        # put the pcolor_col on the pcolormesh collection so
-                        # that if really necessary, users can do things post
-                        # this method
-                        collection._wrapped_collection_fix = pcolor_col
-
+                    # the part below in comment retrieves the masked 
+                    # overlapping cells and plot them with pcolor and a 
+                    # smaller zorder to hide them
+                    # but there are not always hidden 
+                    # if one wants to retrieve those cells, split them at the 
+                    # boundary and plot them, the code below could be 
+                    # useful.
+                    
+#                    #############
+#
+#                    # now that the pcolormesh has masked the bad values,
+#                    # create a pcolor with just those values that were masked
+#                    if C_mask is not None:
+#                        # remember to re-apply the original data mask
+#                        pcolor_data = np.ma.array(C, mask=~mask | C_mask)
+#                    else:
+#                        pcolor_data = np.ma.array(C, mask=~mask)
+#
+#                    pts = coords.reshape((Ny, Nx, 2))
+#                    
+#                    if np.any(~pcolor_data.mask):
+#                        # plot with slightly lower zorder to avoid odd issue
+#                        # where the main plot is obscured
+#                        zorder = collection.zorder - .1
+#                        kwargs.pop('zorder', None)
+#                        kwargs.setdefault('snap', False)
+#                        pcolor_col = self.pcolor(pts[..., 0], pts[..., 1],
+#                                                 pcolor_data, zorder=zorder,
+#                                                 **kwargs)
+#
+#                        pcolor_col.set_cmap(cmap)
+#                        pcolor_col.set_norm(norm)
+#                        pcolor_col.set_clim(vmin, vmax)
+#                        # scale the data according to the *original* data
+#                        pcolor_col.norm.autoscale_None(C)
+#
+#                        # put the pcolor_col on the pcolormesh collection so
+#                        # that if really necessary, users can do things post
+#                        # this method
+#                        collection._wrapped_collection_fix = pcolor_col
+#                    #############
             # Clip the QuadMesh to the projection boundary, which is required
             # to keep the shading inside the projection bounds.
             collection.set_clip_path(self.background_patch)
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1720,14 +1720,14 @@ class GeoAxes(matplotlib.axes.Axes):
 
                     collection.set_array(pcolormesh_data.ravel())
 
-                    # the part below in comment retrieves the masked 
-                    # overlapping cells and plot them with pcolor and a 
+                    # the part below in comment retrieves the masked
+                    # overlapping cells and plot them with pcolor and a
                     # smaller zorder to hide them
-                    # but there are not always hidden 
-                    # if one wants to retrieve those cells, split them at the 
-                    # boundary and plot them, the code below could be 
+                    # but there are not always hidden
+                    # if one wants to retrieve those cells, split them at the
+                    # boundary and plot them, the code below could be
                     # useful.
-                    
+
 #                    #############
 #
 #                    # now that the pcolormesh has masked the bad values,
@@ -1739,7 +1739,7 @@ class GeoAxes(matplotlib.axes.Axes):
 #                        pcolor_data = np.ma.array(C, mask=~mask)
 #
 #                    pts = coords.reshape((Ny, Nx, 2))
-#                    
+#
 #                    if np.any(~pcolor_data.mask):
 #                        # plot with slightly lower zorder to avoid odd issue
 #                        # where the main plot is obscured


dopplershift:
We shouldn't be commenting out code, but just remove it if it's wrong. Before doing so, I'd really want to see some before and after comparisons, because that original code was added for a reason (that I haven't dug into yet to figure out).

htonchia:
The part, I have commented out, plots with pcolor the overlapping cells that have been masked earlier. It plots them with a lower zorder. So if there was no land_mask or it the data were atmospheric data, they would not show. With oceanographic data, like ice concentration, those cells still appear on the plot despite their lower zorder.
It seems to me that its is a very common issue when one designs plot functions for atmospheric data and uses them for oceanographic data. 

My understanding is that by keeping those cells, one can later process them to split them at the boundary and plot them correctly, that's why I kept the code commented out. Removing it would be cleaner and if one wants to process those cells, then one would have to change the code.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1720,36 +1720,47 @@ def _pcolormesh_patched(self, *args, **kwargs):
...
+                    # the part below in comment retrieves the masked
+                    # overlapping cells and plot them with pcolor and a
+                    # smaller zorder to hide them
+                    # but there are not always hidden
+                    # if one wants to retrieve those cells, split them at the
+                    # boundary and plot them, the code below could be
+                    # useful.
+
+#                    #############
+#
```
greglucas: What about making this a kwarg option, rather than commenting it out?

```
if kwargs['keep_bad_cells']:
    ...
```

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1720,36 +1720,47 @@ def _pcolormesh_patched(self, *args, **kwargs):
...
+#                        pcolor_data = np.ma.array(C, mask=~mask | C_mask)
+#                    else:
+#                        pcolor_data = np.ma.array(C, mask=~mask)
+#
+#                    pts = coords.reshape((Ny, Nx, 2))
+#
+#                    if np.any(~pcolor_data.mask):
+#                        # plot with slightly lower zorder to avoid odd issue
+#                        # where the main plot is obscured
+#                        zorder = collection.zorder - .1
```
greglucas: Did you try making this a much lower zorder like -1, so it is below all other artists?

htonchia: I tried a much lower order like -1.
In that case, it doesn't always work so I tried an ever larger number, and it worked.
Now I think there is no guarantee that it will always hid the cell, not plotting them is a better guarantee.
I think the issue comes with masked data, like oceanographic data with a land mask. I would say that when this was written, pushing the "hidden cell" behind the good one, was good enough, but when there is no data on the land mask, then the cells behind still appear on the masked area, not obscuring the data but the mask, and still making a mess of the plot.

Making it a kwargs is a possible option.

The idea was that those cell had to be hidden, so let's hide them efficiently. I admit that I spend quite some time thinking about it and I am sure of one thing is that cartopy needs to get rid of the cells...

greglucas:
I just took a little closer look at this, and I think it is important to keep those lines in there for boundary wrapping purposes. For your cases could you use `mesh._wrapped_collection_fix.set_visible(False)`?

In general, though, I think those cells need to be replotted to appear on the boundaries as demonstrated below.
Figure_1


```python
import numpy as np
import cartopy.crs as ccrs

fig, (ax1, ax2) = plt.subplots(nrows=2, subplot_kw={'projection': ccrs.Mercator()})

x = np.linspace(0, 360)
y = np.linspace(-45, 45)
X, Y = np.meshgrid(x, y)
Z = np.ma.masked_greater(X**2 + Y**2, 200**2)

mesh1 = ax1.pcolormesh(X, Y, Z, transform=ccrs.PlateCarree())
mesh2 = ax2.pcolormesh(X, Y, Z, transform=ccrs.PlateCarree())
mesh2._wrapped_collection_fix.set_visible(False)

ax1.set_title('Before')
ax2.set_title('After')

plt.show()
```

htonchia:
Hi,


I am not sure what you mean.

Plotting the overlapping cells doesn’t improve the boundary, because they are plotted across the plot, not through the boundary.
On this example (which shows the overlapping cells on the bottoms plots), we can see that the boundaries are white, because the overlapping cells are plotted across the plot.
So I don’t see the what plotting those overlapping cells could bring.
Of course, if we want them to cross the plot, this is an other issue.

htonchia:
I tried the example but the _wrapped_collection_fix doesn't work on my version.
Where does it come from ? Cartopy or a new version of matplotlib?

greglucas:
I apologize, I wasn't very clear in the previous message. I think the "Before" in the image above is the correct way of doing things. In the second "After" image, the cells that cross the boundary are masked and removed from the plot (same as commenting out the lines of code).

Note that I haven't looked into the full transformation path, but I think what is happening is:

* Make a quadmesh and mask out any cells spanning half the plot.
* Replot the masked quadmesh cells from above with **pcolor**, which returns a `PolyCollection`
* Now, here is the speculation on my part, I think that the Polygons in the `PolyCollection` go through the `_project_geometry` code path which detects boundary intersections and makes two separate Polygons, one on each side of the boundary.

That is why I think this code path is important to leave in, we want those two extra Polygons to go right up to the border on each side of the map, not to just ignore them.

Perhaps this means that the boundary needs to be defined at a higher resolution for the actual clip paths being used? I'm not sure if #1422 will help here with the clipping or not.
One of the worst offenders of your plot demonstrations is the Mollweide, which it looks like has some issues at the poles. See #1333 and the fix of #1334

I think that your other PR #1420 will help with identifying which cells to mask/wrap better. But, I think there is likely a better way to deal with the masked/wrapped cells than just dropping them altogether.

htonchia:
Thanks for your comment.

I have been studying the issue and will issue soon (I hope) a detailed comment.
I have learned a little bit more on the subject.
I didn't understand why your example didn't work as the data used that highlighted the issue to fix, and now I think I have part of the answer.

The issue of overlapping cells not managed properly by pcolor comes from data provided in lat lon for which we don't know the projection used to create the grid. 
So I am working on a fix that would check this case and manage it.

Your comment and your example have been very useful.

htonchia:
This PR is closed because taken into account by #1467 

After further analysis and using both my example (ice data in stereographic projection) and the example above, I came to the conclusion that pcolor was failing to plot the overlapping cells when the data coordinates were given in the axis projection.

Here is the code to used to show that point.
```
import numpy as np
import cartopy.crs as ccrs


PROJ_DATA = ccrs.Stereographic(
        central_longitude=-45, central_latitude=90,
        true_scale_latitude=70)

def sample_data(proj=PROJ_DATA):
    """
    simulated data in a mesh in projection PROJ_DATA
    with coordinates in latitudes and longitudes
    return (lons, lats) in  projection proj, simulated data, proj
    """
    xcc = np.linspace(-3845, 3745, 760//10) * 1000
    ycc = np.linspace(5845, -5345, 1120//10) * 1000
    xcm, ycm = np.meshgrid(xcc, ycc)
    lls = ccrs.Geodetic().transform_points(proj, xcm, ycm)
    lats = np.deg2rad(lls[..., 1])
    lons = np.deg2rad(lls[..., 0])
    mlats = np.min(lats)
    lats = lats - mlats
    data = wave + mean
    circle = (xcm+2000000)**2 + (ycm-200000)**2
    data = np.ma.masked_where(circle < 1000000**2, data)
    return (lls[..., 0], lls[..., 1]), data, proj

def main():
    """
    Test geoaxe pcolormesh with cells with not removed by
    cell border length but by the diagonals length
    """
    coords, data2plot, proj_data = sample_data(proj=PROJ_DATA)
    proj = ccrs.PlateCarree(central_longitude=182)
    plt.figure(figsize=(12, 6.4), dpi=100)

    ### here we consider that we don't know the data projection
    ### and we have the coordinates in lat lon
    xy2 = proj.transform_points(
         ccrs.Geodetic(),
         coords[0], coords[1])
    cx2, cy2 = xy2[..., 0], xy2[..., 1]

    axe0 = plt.subplot(4, 2, 1, projection=proj)
    axe0.pcolormesh(cx2, cy2, data2plot)
    axe0.set_title('pcolormesh _wrapped_collection_fix: auto, ax proj')
    axe1 = plt.subplot(4, 2, 3, projection=proj)
    axe1.set_title('pcolormesh _wrapped_collection_fix visible ax proj')
    mesh1 = axe1.pcolormesh(cx2, cy2, data2plot)
    axe2 = plt.subplot(4, 2, 5, projection=proj)
    mesh2 = axe2.pcolormesh(cx2, cy2, data2plot)
    axe2.set_title('pcolormesh _wrapped_collection_fix not visible ax proj')
    axe3 = plt.subplot(4, 2, 7, projection=proj)
    axe3.pcolor(cx2, cy2, data2plot)
    if hasattr(mesh2, '_wrapped_collection_fix'):
        mesh2._wrapped_collection_fix.set_visible(False)
        mesh1._wrapped_collection_fix.set_visible(True)
    axe3.set_title('pcolor coords ax proj')

    ### here we consider that we know the data projection
    ### and we have the coordinates in data projection
    xy2 = proj_data.transform_points(
         ccrs.Geodetic(),
         coords[0], coords[1])
    cx2, cy2 = xy2[..., 0], xy2[..., 1]

    axe10 = plt.subplot(4, 2, 2, projection=proj)
    axe10.pcolormesh(cx2, cy2, data2plot, transform=proj_data)
    axe10.set_title('pcolormesh _wrapped_collection_fix auto projdata')
    axe4 = plt.subplot(4, 2, 4, projection=proj)
    mesh4 = axe4.pcolormesh(cx2, cy2, data2plot, transform=proj_data)
    axe4.set_title('pcolormesh _wrapped_collection_fix visible projdata')
    axe5 = plt.subplot(4, 2, 6, projection=proj)
    mesh5 = axe5.pcolormesh(cx2, cy2, data2plot, transform=proj_data)
    axe5.set_title('pcolormesh _wrapped_collection_fix not visible  projdata')
    axe6 = plt.subplot(4, 2, 8, projection=proj)
    axe6.pcolor(cx2, cy2, data2plot, transform=proj_data)
    if hasattr(mesh5, '_wrapped_collection_fix'):
        mesh5._wrapped_collection_fix.set_visible(False)
        mesh4._wrapped_collection_fix.set_visible(True)
    axe6.set_title('pcolor coords in projdata')


    plt.show()
    plt.savefig("img/PR1419_test_wrapped_collection.png",
                    dpi=100, pad_inches=0)



if __name__ == '__main__':
    main()

```
Here is the resulting plot:
PR1419_test_wrapped_collection

Here is a very simple example with big cells showing how it works.
This example uses the geoaxes of #1467.
In axes projection pcolor plot the cells as they come, not managing wrapping, pcolormesh hides the large cells considering it is an overlap (they can be set to visible though as showed in the previous code).
```
import numpy as np
import cartopy.crs as ccrs

axproj = ccrs.Mercator()

fig, ((ax1, ax2), (ax3, ax4)) = plt.subplots(nrows=2, ncols=2, 
         figsize=(12, 6.4), dpi=100,
         subplot_kw={'projection': axproj})

lon = np.linspace(120, 240, 4)
lat = np.linspace(0, 75, 4)
lons, lats = np.meshgrid(lon, lat)
Z = lons**2 + lats**2 * 10

xyas = axproj.transform_points(ccrs.Geodetic(), lons, lats)

ax1.pcolormesh(lon, lat, Z, transform=ccrs.PlateCarree())
ax2.pcolormesh(xyas[..., 0], xyas[..., 1],  Z)
ax3.pcolor(lon, lat, Z, transform=ccrs.PlateCarree())
ax4.pcolor(xyas[..., 0], xyas[..., 1], Z)

ax1.set_title('pcolormesh coords in data projection')
ax2.set_title('pcolormesh coords in axis projection')
ax3.set_title('pcolor coords in data projection')
ax4.set_title('pcolor coords in axis projection')
for axe in [ax1, ax2, ax3, ax4]:
    axe.set_global()
    axe.coastlines()

plt.show()
plt.savefig("img/I1416_test_dataproj.png",
                    dpi=100, pad_inches=0)
```
I1416_test_dataproj

PR #1424, opened 2019-12-18
Followup to #1117

### LINKED ISSUE #1338: Documentation errors and redundant code in geoaxes
Some valid points were raised in 
[@QuLogic  review on #1117](https://github.com/SciTools/cartopy/pull/1117#pullrequestreview-192568559)
These were not attended to at the time, because it was previously merged.

QuLogic:
## Rationale

This takes care of the documentation and minor cleanups I noted in #1117. Also fixes a few minor bugs.

## Implications

Fixes #1338.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -25,7 +25,6 @@ import matplotlib.collections as mcollections
 import matplotlib.ticker as mticker
 import matplotlib.transforms as mtrans
 import matplotlib.path as mpath
-import matplotlib.pyplot as plt
 import numpy as np
 import shapely.geometry as sgeom
 
@@ -710,7 +709,7 @@ class Gridliner(object):
 
                 if angle is not None:
                     specs, _ = self._segment_angle_to_text_specs(angle, lonlat)
-                    plt.setp(artist, **specs)
+                    artist.update(specs)
 
                 artist.update_bbox_position_size(renderer)
                 this_patch = artist.get_bbox_patch()
@@ -750,7 +749,7 @@ class Gridliner(object):
             # Action
             artist.set_visible(visible)
             if not visible:
-                plt.setp(artist, **orig_specs)
+                artist.update(orig_specs)
             else:
                 paths.append(this_path)
 
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -145,14 +145,14 @@ class Gridliner(object):
             the given CRS. Defaults to None, which implies automatic locating
             of the gridlines.
         xformatter: optional
-            A :class:`matplotlib.ticker.Formatter` instance to format
-            longitude labels. It defaults to None, which implies the use of a
-            :class:`cartopy.mpl.ticker.LongitudeFormatter` initiated
+            A :class:`matplotlib.ticker.Formatter` instance to format labels
+            for x-coordinate gridlines. It defaults to None, which implies the
+            use of a :class:`cartopy.mpl.ticker.LongitudeFormatter` initiated
             with the ``dms`` argument.
         yformatter: optional
-            A :class:`matplotlib.ticker.Formatter` instance to format
-            latitude labels. It defaults to None, which implies the use of a
-            :class:`cartopy.mpl.ticker.LatitudeFormatter` initiated
+            A :class:`matplotlib.ticker.Formatter` instance to format labels
+            for y-coordinate gridlines. It defaults to None, which implies the
+            use of a :class:`cartopy.mpl.ticker.LatitudeFormatter` initiated
             with the ``dms`` argument.
         collection_kwargs: optional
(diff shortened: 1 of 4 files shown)


lib/cartopy/mpl/gridliner.py
```diff
@@ -511,23 +502,34 @@ def _draw_gridliner(self, nx=None, ny=None, background_patch=None,
...
-                        tails = []
-                        heads = []
-                        for ht, slicer in [(tails, slice(0, 2)),
-                                           (heads, slice(-1, -3, -1))]:
-                            x = array('d', [])
-                            y = array('d', [])
-                            for geom in list(intersection.geoms)[slicer]:
-                                x += geom.xy[0]
-                                y += geom.xy[1]
-                            ht.append(list(zip(x[slicer], y[slicer])))
```
QuLogic: @stefraynaud I think there used to be a bug here, but can you confirm?

When producing `heads`, it looks at the last geom, grabs its coordinates (in order), then appends the coordinates (in order) of the second-last geom. Finally, it takes the last two points from that list of coordinates in reverse. If the last two geoms are `Point`s, then you get the second-last and then the last coordinate.

But if the second-last geom is a `LineString`, then you get the last two coordinates of the second-last geom, when it should be the last coordinate of the second-last geom and the coordinate of the last geom.

Also, for the new code, I'm not sure if it `heads` should be in order, or reversed.

QuLogic: Here is an example:
```python
In [6]: def new(intersection): 
   ...:     xy = [] 
   ...:     for geom in intersection.geoms: 
   ...:         for coord in geom.coords: 
   ...:             xy.append(coord) 
   ...:             if len(xy) == 2: 
   ...:                 break 
   ...:         if len(xy) == 2: 
   ...:             break 
   ...:     tails = [xy] 
   ...:     xy = [] 
   ...:     for geom in reversed(intersection.geoms): 
   ...:         for coord in reversed(geom.coords): 
   ...:             xy.append(coord) 
   ...:             if len(xy) == 2: 
   ...:                 break 
   ...:         if len(xy) == 2: 
   ...:             break 
   ...:     heads = [xy] 
   ...:     return tails, heads 
   ...:                                                                                                                                               

In [7]: def orig(intersection): 
   ...:     tails = [] 
   ...:     heads = [] 
   ...:     for ht, slicer in [(tails, slice(0, 2)), 
   ...:                        (heads, slice(-1, -3, -1))]: 
   ...:         x = array('d', []) 
   ...:         y = array('d', []) 
   ...:         for geom in list(intersection.geoms)[slicer]: 
   ...:             x += geom.xy[0] 
   ...:             y += geom.xy[1] 
   ...:         ht.append(list(zip(x[slicer], y[slicer]))) 
   ...:     return tails, heads 
   ...:                                                                                                                                               
```

If it's just `Point`s, then the result is:
```python
In [8]: g = sgeom.GeometryCollection([sgeom.Point(i, i) for i in range(10)])                                                                          

In [9]: orig(g)                                                                                                                                       
Out[9]: ([[(0.0, 0.0), (1.0, 1.0)]], [[(8.0, 8.0), (9.0, 9.0)]])

In [10]: new(g)                                                                                                                                       
Out[10]: ([[(0.0, 0.0), (1.0, 1.0)]], [[(9.0, 9.0), (8.0, 8.0)]])
```
So I should flip the `heads` to be in order. I'm not sure if it's semantically correct, but at least it will match.

But if you use `LineString`s (`x` is the geometry, and `y` is the point in the geometry):
```python
In [11]: g = sgeom.GeometryCollection([sgeom.LineString([(x, y) for y in range(10)]) for x in range(10)])                                             

In [12]: g                                                                                                                                            
Out[12]: <shapely.geometry.collection.GeometryCollection at 0x7f68e544b550>

In [13]: orig(g)                                                                                                                                      
Out[13]: ([[(0.0, 0.0), (0.0, 1.0)]], [[(8.0, 9.0), (8.0, 8.0)]])

In [14]: new(g)                                                                                                                                       
Out[14]: ([[(0.0, 0.0), (0.0, 1.0)]], [[(9.0, 9.0), (9.0, 8.0)]])
```
So the original code picks the second-last geometry and returns the last two elements from it. Not only that, but this time it returns the elements in reverse.

QuLogic: This code path is only triggered 5 times in `test_grid_labels_inline`, and in all cases, it's 10% `Point`, followed by `LineString` (except the one time it's 1 of each.) So `tails` is always consistent between old and new, but `heads` is always reversed in the code as it stands. Thus I will leave it reversed for the `Point`s case even though the original didn't do that.

QuLogic:
Since you basically approved, minus the question, and I think I've managed to answer it without any further changes, I'm going to merge this.

lib/cartopy/mpl/gridliner.py
stefraynaud: It seems you're right

PR #1425, opened 2019-12-18
Prepare for Matplotlib 3.2 and 3.3

QuLogic:
## Rationale

Matplotlib deprecated some things and moved around some internals. Streamplot was also changed to correct a bug (matplotlib/matplotlib#11947).

## Implications

Tests should continue to work in new Matplotlib and generate no warnings.

Matplotlib 3.3 support patch from @anntzer [here](https://github.com/matplotlib/matplotlib/issues/15806#issuecomment-564055197).


dopplershift:
So are we allowed to override the CLA checker...?

ajdawson:
I guess strictly all commit authors should have agreed to the CLA which hasn't happened in this case. It would probably be better (but I agree, much less convenient) to ask the author @anntzer to do so via https://scitools.org.uk/cla/v4/form

anntzer:
I would rather not have my name attached to the patch (which I can release e.g. in the public domain(?)), in that case.  Would that work for you?

ajdawson:
>I would rather not have my name attached to the patch (which I can release e.g. in the public domain(?)), in that case. Would that work for you?

That works for me (caveat: neither a lawyer or a Met Office employee)

QuLogic:
If you call it Public Domain, or CC0 where not possible, then I think the lawyers _shouldn't_ have anything to complain about.

anntzer:
I hereby release the patch in question in the public domain.

ajdawson:
Thanks all, sorry for the hassle!

Issue #1345, opened 2019-07-22
tricontour and tricontourf produce wrong bounds


MiCurry:
### Description

Similar to issue #811 and its fix #814, tricontour and tricountourf incorrectly
Change the bounds of the plot to be extremely zoomed in. i.e.:

![ortho-zoomed](https://user-images.githubusercontent.com/2590700/61646936-740c0880-ac69-11e9-9a7f-d31316fae626.png)

The extent can be reset using:
```
ax.relim()
ax.autoscale_view()
```

![ortho-normal](https://user-images.githubusercontent.com/2590700/61646740-7e79d280-ac68-11e9-8c9b-ed20e62e1383.png)

#### Code to reproduce
```
import numpy as np
import matplotlib.pyplot as plt
import matplotlib.cm as cm
import cartopy.crs as ccrs

npts = 5000
latCell = np.random.uniform(-90.0, 90.0, npts)
lonCell = np.random.uniform(-180.0, 180.0, npts)
data = np.random.random_sample((latCell.shape[0]))

bmap = ccrs.AlbersEqualArea()

ax = plt.axes(projection=bmap)

ax.coastlines()
ax.gridlines()

plt.tricontourf(lonCell, 
                latCell, 
                data,
                75,
                vmin=data.min(),
                vmax=data.max(),
                transform=ccrs.PlateCarree())

#ax.relim()
#ax.autoscale_view()
```

Similar open issues include #1301, #1302 and maybe #1293(?). We definitely have some interest in getting tricontour and tricontourf working correctly for plotting MPAS field. I definitely have an interesting in helping make this contribution, but I'll need to get permission from NCAR and I would most likely need some help making the change as well.

<details>
 <summary>Full environment definition</summary>

### Operating system
SUSE Linux Enterprise Server 12 SP4

### Cartopy version
0.17.0

### conda list
Not using conda

### pip list

```
 Package                       Version    Location                        
----------------------------- ---------- --------------------------------
alabaster                     0.7.12     
ASAPTools                     0.6.2      
asn1crypto                    0.24.0     
aspy.yaml                     1.3.0      
astroid                       2.2.5      
atomicwrites                  1.3.0      
attrs                         19.1.0     
Babel                         2.7.0      
backcall                      0.1.0      
basemap                       1.2.0      
bash-kernel                   0.7.1      
beautifulsoup4                4.7.1      
bitstring                     3.1.5      
bleach                        3.1.0      
boto3                         1.9.179    
[...]

PR #1430, opened 2019-12-31
Changing autoscale to work with collections during draw.

greglucas:
I have added a quick test case to show that adding collections doesn't properly update the extents, and with this fix the test should pass. The example was taken from https://github.com/SciTools/cartopy/issues/1345#issuecomment-513864270

Note that it doesn't fix the issue with `.tri*()` functions in that issue. I think those will need to be wrapped in Cartopy-specific functions to get back proper data extents returned still.

## Rationale

Update the plot extents when a collection is added to the axes.

## Implications

Adding a collection currently modifies `self.ignore_existing_data_limits`, removing the ability to update the limits of the plot. This removes that check in the draw command. I'm not sure why it was there in the first place, so there could be a potential reason not to do this.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -413,7 +413,7 @@ class GeoAxes(matplotlib.axes.Axes):
         # If data has been added (i.e. autoscale hasn't been turned off)
         # then we should autoscale the view.
 
-        if self.get_autoscale_on() and self.ignore_existing_data_limits:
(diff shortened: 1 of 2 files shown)


greglucas:
This is actually an issue in upstream Matplotlib that should be fixed there instead of the hack here. I'm going to close in hopes that it will continue to be addressed upstream.

See the discussion here https://github.com/matplotlib/matplotlib/issues/7413 and linked issues/PRs within that.

PR #1431, opened 2019-12-31
Changing the transform checks to a decorator

greglucas:
## Rationale

There was a lot of repetitive transform checks/changes inside each function, so I changed it to a single function that wraps each of those calls. This will be more consistent and maintainable going forward.

## Implications

I am not sure how far back in matplotlib versions this will work due to removing all of the `_to_mpl` calls because this is now just passing the Cartopy projections directly into the axes plot calls.


lib/cartopy/mpl/geoaxes.py
```diff
@@ -280,6 +281,31 @@ def set_position(self, position):
             'GeoSpine does not support changing its position.')
 
 
+def _add_transform(func):
+    """A wrapper that adds and validates the transform kwarg."""
```
QuLogic:
```suggestion
    """A decorator that adds and validates the transform keyword argument."""
```

lib/cartopy/mpl/geoaxes.py
```diff
@@ -30,6 +30,7 @@
 import contextlib
 import warnings
 import weakref
+import functools
```
QuLogic: Keep sorted; belongs between `contextlib` and `warnings`.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -280,6 +281,31 @@ def set_position(self, position):
...
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        transform = kwargs.get('transform', None)
+        if transform is None:
+            transform = self.projection
+        if (isinstance(transform, ccrs.CRS) and
+                not isinstance(transform, ccrs.Projection)):
+            raise ValueError('Invalid transform:'
+                             ' Spherical transforms are not supported - '
+                             ' consider using PlateCarree/RotatedPole.')
```
QuLogic: This error seems a bit weird now; there are many `Projection`s other than these two which appear spherical-ish.

Also, you should be able to insert the name of the method by inspecting `func`.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1542,26 +1540,11 @@ def scatter(self, *args, **kwargs):
...
-        # Keep this bit - even at mpl v1.2
-        if t is None:
-            t = self.projection
-        if hasattr(t, '_as_mpl_transform'):
-            kwargs['transform'] = t._as_mpl_transform(self)
-
-        # exclude Geodetic as a valid source CS
-        if (isinstance(kwargs.get('transform', None),
-                       InterProjectionTransform) and
-                kwargs['transform'].source_projection.is_geodetic()):
```
QuLogic: This looks like a different check, no?

lib/cartopy/mpl/geoaxes.py
greglucas: I agree. I was really just trying to maintain consistency with everything else. This could all be tidied up in the future. And it will hopefully be easier with a single source that needs updating rather than each function.

lib/cartopy/mpl/geoaxes.py
greglucas: Yes, I was not even sure it was actually possible to enter this statement because above you're setting kwargs['transform'] to a mpl transform...

Also note that below it actually mentions `contouring` but this is in `scatter`. There are a lot of inconsistencies here that could be addressed later.

I've added it back in for now and changed contouring to scatter.

greglucas:
Thanks for the review @QuLogic. I was quick and loose with the wording here as you pointed out. I've made all the changes you suggested.

I tried to rebase to get a cleaner history and now it thinks I want to merge your last commit from master in again. Can that be fixed on merge or do I somehow need to rebase it on my end differently?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1497,25 +1513,6 @@ def contourf(self, *args, **kwargs):
...
-            raise ValueError('invalid transform:'
-                             ' Spherical contouring is not supported - '
-                             ' consider using PlateCarree/RotatedPole.')
-        if isinstance(t, ccrs.Projection):
-            kwargs['transform'] = t = t._as_mpl_transform(self)
-        else:
-            kwargs['transform'] = t
-
-        # Set flag to indicate correcting orientation of paths if not ccw
-        if isinstance(t, mtransforms.Transform):
```
QuLogic: `ccrs.Projection` is not a `mtransforms.Transform`, but the removed `_as_mpl_transform` would have converted it to a `InterProjectionTransform + axes.transData` which _is_. So this code path would not be triggered in the new version.

Strangely, the code that reads this (somewhere else in this file only), uses `getattr` with a default of `False`. I don't know which one is right. Since this is only trying to affect `InterProjectionTransform`, I wonder why that class doesn't just set a default `force_path_ccw` itself.

This is the only one that checks this though, right, so maybe you can just revert this?

lib/cartopy/mpl/geoaxes.py
QuLogic: True, I'm not sure how you might end up with an `InterProjectionTransform`, but I guess someone could pass it in directly.

lib/cartopy/mpl/geoaxes.py
greglucas: I agree that it is actually cleaner if I put it back in here since this is the only place it is used, so I removed the force_ccw from the _add_transform decorator and put it back here. I get a bit lost reading through all of the Transform pipelines here, and all of the various branches/possibilities with special casing.

QuLogic: Thanks for putting it back, but you missed the first part of my comment, namely that the `_as_mpl_transform` would make `Projection` into a `Transform` that would match this condition.

greglucas: Ahh, I see that now. So, I have now added that back in here as well.

However, another thing to consider is that the `_as_mpl_transform()` was actually used in a lot of the other functions as well. But, seemingly has no effect for those ones and only this code path as you pointed out. I'm not entirely clear whether the call should be put in the decorator or just left here for this one function. Thoughts?

QuLogic: Well, tests seem to be okay with it. I think that was needed for old Matplotlib versions which didn't handle it itself. Eventually, the conversion was upstreamed, so this is no longer needed here.

PR #1432, opened 2020-01-02
Add lcc

jeliashi:
## Rationale

This PR allows users to create Lambert Conformal Conic projections. The available default closely resembles the "conus" region used in NCAR's HRRR model.

## Implications

This will add an available projection to the projection list that allows for Lambert Conformal Conic often used in WRF grids.

Example usage:
```
import cartopy.crs as ccrs

plt.figure(figsize=(4, 3))
ax = plt.axes(projection=ccrs.LambertConformalConic())
ax.coastlines()
ax.gridlines()
```

Another example:
```
import cartopy.crs as ccrs
import cartopy.feature as cfeat

plt.figure(figsize=(4, 3))
Australia = LambertConformalConic(central_longitude=134.489, central_latitude=-23.993,
                 x_limits=(-2.25e6, 2.25e6),
                 y_limits=(-1.9e6, 1.9e6)
                 )
ax = plt.axes(projection=Australia)
ax.add_feature(cfeat.LAND)
ax.add_feature(cfeat.OCEAN)
ax.add_feature(cfeat.COASTLINE)
ax.gridlines()
```

### DIFF
--- a/docs/source/contributors.rst
+++ b/docs/source/contributors.rst
@@ -39,6 +39,7 @@ the package wouldn't be as rich or diverse as it is today:
  * Robert Redl
  * Greg Lucas
  * Sadie Bartholomew
+ * Jonathan Eliashiv
 
 Thank you!
 
--- a/docs/source/crs/projections.rst
+++ b/docs/source/crs/projections.rst
@@ -96,6 +96,20 @@ LambertConformal
     ax.coastlines(resolution='110m')
     ax.gridlines()
 
+LambertConformalConic
+---------------------
+
+.. autoclass:: cartopy.crs.LambertConformalConic
+
+.. plot::
+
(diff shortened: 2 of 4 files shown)


jeliashi:
> It looks like there is already a Lambert Conformal Conic projection.
> https://scitools.org.uk/cartopy/docs/latest/crs/projections.html#lambertconformal
> 
> Perhaps you could update that one with what you specifically need added/changed?
> https://github.com/SciTools/cartopy/blob/c7cb50f3e9235bf80775e4ba62ac8074eb96f36d/lib/cartopy/crs.py#L1132-L1141

Good point!. I'll go ahead and close this PR and make a new when that extends the `LambertConformal` projections to allow for regional projections.

PR #1440, opened 2020-01-11
Adding a bounds check to imshow fast path.

### LINKED ISSUE #1439: central_longitude=0.0 and central_longtitude=0 give different results
I would like to project some simple data using a PlateCarree projection. When I set `central_longitude=0` everything works as expected, but if I specify `central_longitude=0.0` their are major problems, as shown in the image below.

```
import cartopy.crs as ccrs
interval = 30

# set the longitudes and latitudes to span 0-360 and 90 to -90.
lat = np.arange(90, -90-interval, -interval)
lon = np.arange(0, 360+interval, interval)
longrid, latgrid = np.meshgrid(lon, lat)

# make a grid of data that representats the function sin(lon)*cos(lat). Note that the value
# represents the center of a pixel.
data = np.sin(np.deg2rad(longrid)) * np.cos(np.deg2rad(latgrid))

# set the extent of the data coordinates: the boundaries are extended by half a pixel in
# all directions since the function uses gridline coordinates.
extent=(lon[0]-interval/2, lon[-1]+interval/2, lat[-1]-interval/2, lat[0]+interval/2)

# plot using a PlateCarree projection
fig = plt.figure()
ax1 = fig.add_subplot(121, projection=ccrs.PlateCarree(central_longitude=0))
ax1.imshow(data, transform=ccrs.PlateCarree(), extent=extent, origin='upper')
ax1.gridlines()
ax1.set_title('central_longitude=0')

ax2 = fig.add_subplot(122, projection=ccrs.PlateCarree(central_longitude=0.0))
ax2.imshow(data, transform=ccrs.PlateCarree(), extent=extent, origin='upper')
ax2.gridlines()
ax2.set_title('central_longitude=0.0')
```
test

greglucas:
## Rationale

The current imshow call does not wrap the image if the image extents are beyond the valid projection. This adds an additional bounds check to make sure that the fast path is only called when fully within the projection domain.

Fixes #1439

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1223,20 +1223,30 @@ class GeoAxes(matplotlib.axes.Axes):
             Default is ``'lower'``.
 
         """
-        transform = kwargs.pop('transform', None)
         if 'update_datalim' in kwargs:
             raise ValueError('The update_datalim keyword has been removed in '
                              'imshow. To hold the data and view limits see '
                              'GeoAxes.hold_limits.')
 
+        transform = kwargs.pop('transform', None)
+        extent = kwargs.get('extent', None)
         kwargs.setdefault('origin', 'lower')
 
         same_projection = (isinstance(transform, ccrs.Projection) and
                            self.projection == transform)
+        # Only take the shortcut path if the image is within the current
+        # bounds (+/- threshold) of the projection
+        x0, x1 = self.projection.x_limits
+        y0, y1 = self.projection.y_limits
+        eps = self.projection.threshold
+        inside_bounds = (extent is None or
+                         (x0 - eps <= extent[0] <= x1 + eps and
+                          x0 - eps <= extent[1] <= x1 + eps and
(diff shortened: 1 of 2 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -1223,20 +1223,30 @@ def imshow(self, img, *args, **kwargs):
...
+        x0, x1 = self.projection.x_limits
+        y0, y1 = self.projection.y_limits
+        eps = self.projection.threshold
+        inside_bounds = (extent is None or
+                         (x0 - eps <= extent[0] <= x1 + eps and
+                          x0 - eps <= extent[1] <= x1 + eps and
+                          y0 - eps <= extent[2] <= y1 + eps and
+                          y0 - eps <= extent[3] <= y1 + eps))
+
+        if ((transform is None) or (transform == self.transData) or
```
QuLogic: These inner parentheses are redundant.

greglucas:
Good suggestion on the test. I added a quick test and restarted CI.

PR #1452, opened 2020-02-07
Fix file paths for coverage on coveralls.

QuLogic:
## Rationale

Currently, coverage is recorded for the file path as installed in the conda environment on Travis. This causes coveralls to be unable to map coverage to files in the repo, though it still produces final statistics.

## Implications

We need to flatten those paths back to the source directory, so that coverage works properly.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -12,6 +12,11 @@ omit =
     *.pxd
 plugins = Cython.Coverage
 
+[paths]
+source =
+    lib/cartopy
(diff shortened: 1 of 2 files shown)


QuLogic:
OK, now it doesn't post?

QuLogic:
The decrease is somewhat expected since the wrong paths are no longer reported. It doesn't seem to have affected the 2.7 build, and what I tried didn't work. I'm not sure how to get it to work, but I'm not so sure there's much point.

PR #1453, opened 2020-02-07
Fix Travis build for Python 2 latest.

QuLogic:
## Rationale

The latest `pytest-forked` uses `get_latest_marker`, which is only in a new version of `pytest`, that appears unavailable for Conda (or maybe Python 2).

## Implications

Pin `pytest-forked` to an older version so that Travis works again.


QuLogic:
Merging to fix CI, and I think this should probably be not to controversial (plus Py2 is going away soon).

PR #1454, opened 2020-02-08
Make using tox a bit easier

QuLogic:
## Rationale

There are two things here, that help with using tox:
* When running with tox, every install to the environment uses a single sdist. This sdist will not contain the Cython-processed files (unless you built them in your source tree previously), plus we want to rebuild them anyway using the Cython in the environment.
* For installed copies when not running pytest in the source directory, then pytest cannot find the `setup.cfg`, thus doesn't know about the markers, and complains about them.

## Implications

* Add a build envvar to force using Cython.
  This option, when added to tox configuration, allows the sdist to be installed, even without pre-created Cython sources.
* Add a conftest to setup pytest markers.

### DIFF
--- a/lib/cartopy/tests/conftest.py
+++ b/lib/cartopy/tests/conftest.py
@@ -1,0 +1,27 @@
+# (C) British Crown Copyright 2020, Met Office
+#
+# This file is part of cartopy.
+#
+# cartopy is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the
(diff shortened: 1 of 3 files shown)


ajdawson:
Do we still need to define the markers in setup.cfg then?

QuLogic:
Good point, I always had this as an extra file, so I never tried that. I just did right now, and it works correctly with `conftest.py` only, so I've removed the lines from `setup.cfg`.

PR #1461, opened 2020-02-11
Mark additional tests as requiring network.

sebastic:
The Debian package build for 0.18.0b1 failed due to test failures.

`test_repr_html` failed because it the is no network in the package build environment.


sebastic:
I don't sign CLAs, I reject such barriers to contribution.

The bug is reported in #1462.

QuLogic:
The CLA is more of a DCO, but it's not my call.

PR #1463, opened 2020-02-11
Fix proj lib name

ocefpaf:
## Rationale

I believe the sue of `proj>=6` in the `setup.py` logic was mislead due to the use of the old `proj4` in `.appveyor.yml`. I updated the latter and fixed the former. Hopefully this is a simple change and things will pass.

## Implications

This change will allow us to build `cartopy` on Windows with latest and old `proj` with either library names (`proj` and `proj4`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -242,7 +242,7 @@ def get_proj_libraries():
     This function gets the PROJ libraries to cythonize with
     """
     proj_libraries = ["proj"]
-    if os.name == "nt" and proj_version >= (6, 0, 0):
+    if os.name == "nt" and proj_version < (6, 0, 0):
         proj_libraries = [
             "proj_{}_{}".format(proj_version[0], proj_version[1])
         ]
--- a/setup.py
+++ b/setup.py
@@ -242,11 +242,7 @@ def get_proj_libraries():
     This function gets the PROJ libraries to cythonize with
     """
     proj_libraries = ["proj"]
-    if (
-        os.name == "nt"
-        and proj_version >= (6, 0, 0)
-        and proj_version < (6, 3, 0)
-    ):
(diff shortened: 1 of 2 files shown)


ocefpaf:
The failing test is for `Python 3.5` I recommend to remove that test b/c conda-forge no longer supports it. Or, as an alternative, run it with `defaults` only packages.

setup.py
```diff
@@ -242,7 +242,11 @@ def get_proj_libraries():
     This function gets the PROJ libraries to cythonize with
     """
     proj_libraries = ["proj"]
-    if os.name == "nt" and proj_version >= (6, 0, 0):
+    if (
+        os.name == "nt"
+        and proj_version >= (6, 0, 0)
+        and proj_version < (6, 3, 0)
+    ):
```
QuLogic:
```suggestion
    if os.name == "nt" and (6, 0, 0) <= proj_version < (6, 3, 0):
```
should fit.

ocefpaf: Thanks!

QuLogic:
Upstream issue appears to be here: https://github.com/OSGeo/PROJ/issues/1479 but it's milestoned 7.0.0. Is the milestone incorrect, or is is conda-forge doing a downstream rename?

ocefpaf:
> Upstream issue appears to be here: [OSGeo/PROJ#1479](https://github.com/OSGeo/PROJ/issues/1479) but it's milestoned 7.0.0. Is the milestone incorrect, or is conda-forge doing a downstream rename?

I'm pretty sure that is already in `proj 6.3.0`. @snowman2 probably knows better.

snowman2:
Yep, it is in 6.3. I remember it broke the windows back on the build: https://github.com/conda-forge/proj.4-feedstock/pull/65

PR #1473, opened 2020-02-23
Color palette issue #1386 version202002

htonchia:
## Rationale

Fix issue #1386 when the color palette of the wms file is lost during the regrid process.
When the image fetched from the wms server is not in the expected projection, it is regridded.
When the source image is not RGB or RGBA, it looses its color palette in the regrid process.
To fix this, I have converted the source image to RGBA before the regrid process.

Note: this is a new pull request to replace #1413 after some clean up and updated comments.

In the following example, when it works, both images have a "blue/red" palette, otherwise the reprojected image, the Mercator one, is a viridis colored image.
As those are images live from a server, they change with time.

```
import cartopy.crs as ccrs

URL = "https://thredds.met.no/thredds/wms/sea/arctic20km/1h/aggregate_be?\
service=WMS&version=1.3.0&request=GetCapabilities"
WMS = WebMapService(URL, timeout=80)
LAYER = 'temperature'
WMS_KWARGS = {'transparent': False,
              'styles': ['boxfill/redblue']}
PROJECTIONS = [ccrs.Mercator(),
               ccrs.PlateCarree()]

for projection in PROJECTIONS:
    subplot_kw = dict(projection=projection)
    ax.add_wms(WMS, LAYER, WMS_KWARGS)
    plt.title(projection.proj4_params['proj'])
```
Capture d’écran 2020-02-23 à 17 14 55

## Implications
I am not aware of any implication.
There would be some implication if in some place cartopy regrids non RGBA image and need those to stay non RGBA.


greglucas:
Thanks, @htonchia!

PR #1412, opened 2019-12-08
Update img_transform.py: regrid to fix issue #1398

htonchia:
Regrid of image was done in geocentric coordinates. It transformed straight lines in great circle which was not the expected behaviour.
In the new version, regrid is done in the source projection and it preserves the straight lines.

## Rationale

Regrid for image in img_transform is supposed to transform an image from a source projection, the coordinate system in which the given image is rectangular, to a target projection.
The regrid was done in geocentric coordinates, so straight lines were transformed in great circles and the image was not kept rectangular in the source projection (with different longitude).
So I have modified regrid to do it in the source projection.

## Implications

I have tested it with imshow on all projections and it worked well, with the exception of UTM, but probably due to the zonage of UTM.
I don't know of an other usage than imshow.

I did't know where to put the example so here it is
```
import numpy as np
import matplotlib.pyplot as plt
import matplotlib.path as mpath
import cartopy.crs as ccrs

INTERVAL = 30
MINLAT = -75
MAXLAT = 75
MINLON = -165
MAXLON = 75

PROJ_DATA = {}
PROJ_DATA[0] = ccrs.PlateCarree(central_longitude=0)
PROJS = {}
PROJS[0] = {}
PROJS[0]['Plate Carre 180'] = ccrs.PlateCarree(central_longitude=180)
PROJS[0]['Plate Carre 0'] = ccrs.PlateCarree(central_longitude=0)
PROJS[0]['Plate Carre 225'] = ccrs.PlateCarree(central_longitude=225)
PROJS[0]['Mollweide 180'] = ccrs.Mollweide(central_longitude=180)
PROJS[0]['Polar -45'] = ccrs.NorthPolarStereo(central_longitude=-45)
PROJS[0]['AlbersEqualArea'] = ccrs.AlbersEqualArea()
PROJS[0]['AzimuthalEquidistant'] = ccrs.AzimuthalEquidistant()
PROJS[0]['EckertI'] = ccrs.EckertI()
PROJS[0]['EckertII'] = ccrs.EckertII()
PROJS[0]['EckertIII'] = ccrs.EckertIII()
PROJS[0]['EckertIV'] = ccrs.EckertIV()
PROJS[0]['EckertV'] = ccrs.EckertV()
PROJS[0]['EckertVI'] = ccrs.EckertVI()
PROJS[0]['EqualEarth'] = ccrs.EqualEarth()
PROJS[0]['EquidistantConic'] = ccrs.EquidistantConic()
[...]

PR #1474, opened 2020-02-23
Update v202002 img transform.py regrid to fix issue #1398

### LINKED ISSUE #1398: Pixels not plotted with straight lines
I have a global grid of data in a numpy array that I would like to plot in a map projection. The data span longitudes from 0 to 360, and latitudes from -90 to 90. When plotting the grid using `imshow`, however, the upper and lower boundaries of the pixels are not plotted as straight lines, and are somewhat curved. This is all the more bizarre given that latitudes should be projected as straight lines in the PlateCarree projection I am using (other projections yield similar results).

test

Here is the code used to generate the plot
```
import cartopy.crs as ccrs
interval = 30

# set the longitudes and latitudes to span 0-360 and 90 to -90.
lat = np.arange(90, -90-interval, -interval)
lon = np.arange(0, 360+interval, interval)
longrid, latgrid = np.meshgrid(lon, lat)

# make a grid of data that representats the function sin(lon)*cos(lat)
data = np.sin(np.deg2rad(longrid)) * np.cos(np.deg2rad(latgrid))

# set the extent of the data coordinates: the boundaries are extended by half a pixel in
# all directions since the function uses gridline coordinates.
extent=(lon[0]-interval/2, lon[-1]+interval/2, lat[-1]-interval/2, lat[0]+interval/2)

# plot using a PlateCarree projection
fig = plt.figure()
ax = fig.add_subplot(111, projection=ccrs.PlateCarree(central_longitude=180))
ax.imshow(data, transform=ccrs.PlateCarree(), extent=extent, origin='upper')
```
I note that when the interval between grid points is smaller, the images look *much* better. I was only using this large grid interval here because I was trying to figure out how Cartopy plots pixels and treats the image "extent" (there isn't any documentation on this but it appears to use the same conventions as in matplotlib: https://matplotlib.org/3.1.1/tutorials/intermediate/imshow_extent.html)

If it is of importance, I am using macOS with the most recent version of Cartopy on master, and all dependencies are up to date.

htonchia:
## Rationale

Regrid of image was done in geocentric coordinates. It transformed straight lines in great circle which was not the expected behaviour.
In the new version, regrid is done in the source projection and it preserves the straight lines.
This fixes #1398 

Regrid for image in img_transform is supposed to transform an image from a source projection, the coordinate system in which the given image is rectangular, to a target projection.
The regrid was done in geocentric coordinates, so straight lines were transformed in great circles and the image was not kept rectangular in the source projection (with different longitude).
So I have modified regrid to do it in the source projection.

Note: this replace #1412 to fix CL and pass the checks.

The code used to test:
```
import numpy as np
import matplotlib.path as mpath
import cartopy.crs as ccrs

INTERVAL = 30
MINLAT = -75
MAXLAT = 75
MINLON = -165
MAXLON = 75

PROJ_DATA = {}
PROJ_DATA[0] = ccrs.PlateCarree(central_longitude=0)
PROJS = {}
PROJS[0] = {}
PROJS[0]['Plate Carre 180'] = ccrs.PlateCarree(central_longitude=180)
PROJS[0]['Plate Carre 0'] = ccrs.PlateCarree(central_longitude=0)
PROJS[0]['Plate Carre 225'] = ccrs.PlateCarree(central_longitude=225)
PROJS[0]['Mollweide 180'] = ccrs.Mollweide(central_longitude=180)
PROJS[0]['Polar -45'] = ccrs.NorthPolarStereo(central_longitude=-45)
PROJS[0]['AlbersEqualArea'] = ccrs.AlbersEqualArea()
PROJS[0]['AzimuthalEquidistant'] = ccrs.AzimuthalEquidistant()
PROJS[0]['EckertI'] = ccrs.EckertI()
PROJS[0]['EckertII'] = ccrs.EckertII()
PROJS[0]['EckertIII'] = ccrs.EckertIII()
PROJS[0]['EckertIV'] = ccrs.EckertIV()
PROJS[0]['EckertV'] = ccrs.EckertV()
PROJS[0]['EckertVI'] = ccrs.EckertVI()
PROJS[0]['EqualEarth'] = ccrs.EqualEarth()
PROJS[0]['EquidistantConic'] = ccrs.EquidistantConic()
PROJS[0]['EuroPP'] = ccrs.EuroPP()
PROJS[0]['Geostationary'] = ccrs.Geostationary()
PROJS[0]['Gnomonic'] = ccrs.Gnomonic()
PROJS[0]['InterruptedGoodeHomolosine'] = ccrs.InterruptedGoodeHomolosine()
PROJS[0]['LambertCylindrical'] = ccrs.LambertCylindrical()
PROJS[0]['LambertConformal'] = ccrs.LambertConformal()
PROJS[0]['LambertAzimuthalEqualArea'] = ccrs.LambertAzimuthalEqualArea()
PROJS[0]['Mercator'] = ccrs.Mercator()
PROJS[0]['Miller'] = ccrs.Miller()
PROJS[0]['NearsidePerspective'] = ccrs.NearsidePerspective()
PROJS[0]['OSGB'] = ccrs.OSGB()
PROJS[0]['OSNI'] = ccrs.OSNI()
PROJS[0]['Orthographic'] = ccrs.Orthographic()
PROJS[0]['Robinson'] = ccrs.Robinson()
PROJS[0]['RotatedPole'] = ccrs.RotatedPole(
    pole_latitude=37.5,
    pole_longitude=177.5)  # issue with extent
PROJS[0]['Sinusoidal'] = ccrs.Sinusoidal()
PROJS[0]['SouthPolarStereo'] = ccrs.SouthPolarStereo()
PROJS[0]['Stereographic'] = ccrs.Stereographic()
PROJS[0]['TransverseMercator'] = ccrs.TransverseMercator()
PROJS[0]['UTM'] = ccrs.UTM(zone=31)

PROJS[1] = {}
PROJS[1]['Polar -45'] = ccrs.NorthPolarStereo(central_longitude=-45)
PROJS[1]['Polar 0'] = ccrs.NorthPolarStereo(central_longitude=0)
PROJS[1]['Plate Carre 180'] = ccrs.PlateCarree(central_longitude=180)
PROJ_DATA[1] = PROJS[1]['Polar -45']

# set the longitudes and latitudes to span 0-360 and 90 to -90.
LAT = np.linspace(MAXLAT, MINLAT, (MAXLAT-MINLAT) / INTERVAL + 1)
LON = np.linspace(MINLON, MAXLON, (MAXLON-MINLON) / INTERVAL + 1)
LONGRID, LATGRID = np.meshgrid(LON, LAT)

# make a grid of DATA that representats the function sin(lon)*cos(lat)
DATA = {}
DATA[0] = np.sin(np.deg2rad(LONGRID)) + np.cos(np.deg2rad(LATGRID))+\
    0.2 * np.cos(np.deg2rad(LONGRID)) + 0.2 * np.sin(np.deg2rad(LATGRID))
DATA[1] = DATA[0][:, :5]
# set the extent of the DATA coordinates: the boundaries are extended by half a pixel in
# all directions since the function uses gridline coordinates.
EXTENT = {}
EXTENT[0] = (LON[0]-INTERVAL/2, LON[-1]+INTERVAL/2, LAT[-1]-INTERVAL/2,
             LAT[0]+INTERVAL/2)

# in polar
EXTENT[1] = [-12500000, 12500000, -12500000, 12500000]

CMAP = {}
CMAP[0] = 'viridis'
CMAP[1] = 'cool'

# circle to make the boundary in polar stereographic
THETA = np.linspace(0, 2 * np.pi, 100)
CENTER, RADIUS = [0, 0], 12500000
VERTS = np.vstack([np.sin(THETA), np.cos(THETA)]).T
CIRCLE = mpath.Path(VERTS * RADIUS + CENTER)

# Plot from PROJ_DATA to other PROJS
# ipr : 2 from stero to others
for ipr in range(2):
    for proj in PROJS[ipr]:
        print(proj)
        fig = plt.figure(figsize=(10, 6), dpi=100)
        axe = fig.add_subplot(111, projection=PROJS[ipr][proj])
        if 'Polar' in proj:
            axe.set_xlim(-12500000, 12500000)
            axe.set_ylim(-12500000, 12500000)
            axe.set_boundary(CIRCLE)
        elif 'RotatedPole' in proj:
            axe.set_xlim(-np.pi * 1.001, np.pi * 1.001)
            axe.set_ylim(-np.pi / 2., np.pi / 2.)
        elif 'UTM' in proj:
            axe.set_extent((0, 6, 20, 80))
            # the colors are not as expected, something is wrong
            # probably due to the wrapping of UTM
        else:
            axe.set_global()
        axe.imshow(DATA[ipr], transform=PROJ_DATA[ipr], extent=EXTENT[ipr],
                   origin='upper', cmap=CMAP[ipr])
        axe.set_title(proj)
        axe.coastlines()

```
## Implications
I have tested it with imshow on all projections and it worked well, with the exception of UTM, but probably due to the zonage of UTM.
I don't know of an other usage than imshow.


htonchia:
It fails the test test_griding_data_outside_projection().
This is expected as the purpose of the bug fix is not to regrid in geocentric coordinates but in the source projection. 
On the plot below the 3 disks display the data to regrid.
Here is the expected regrid results:
Capture d’écran 2020-02-24 à 00 04 18
And below the regrid in source projection, which seems a better one but different from expected.
Capture d’écran 2020-02-24 à 00 04 23
So we would need to modify the test accordingly.

pelson:
Thanks for this @htonchia. Fundamentally, I'm much more comfortable having a spherical regridding algorithm than a cartesian one you are proposing here. Having an algorithm in spherical coordinates gives us no concerns with awkward data around the poles (of either source or target projection) and allows us to stitch back together images which may otherwise need to be split to handle the projection (e.g. when you have full globe data in PlateCarree being transformed to PlateCarree(180) there is a split+join needed).

With regards to addressing #1398 I think we would be better looking at suggesting a pcolormesh, which today is inherently non-spherical (though does have the downside of only being native in the target projection) and so doesn't have the curves which are undesired.

By rejecting this PR, I'm effectively saying that: the coordinate system you are providing when doing an imshow (or directly using ``cartopy.img_transform``) controls the coordinates of the pixels of the image, and do not define the underlying coordinate system of the data/field, which is *always* assumed to be mapped to spherical.

I'm open to the idea of customising this in the same way we can for lines (where the coordinate system defines the geometry of the field rather than the specific coordinates provided), but suggest that removing the spherical behaviour altogether is going to lead to a lot of corner cases that we are unlikely to be able to address easily.

I just want to re-iterate my thanks for your PR, which is well researched and described, but state that I'm 👎 on this being merged without customisation options and having the spherical treatment be the default behaviour.

htonchia:
HI @pelson 
Thank you for your comment.

I understand that you are saying that an image is always assumed to be spherical and that there could be issues with pole at interpolation.
Scientific data can be given in geographic coordinates without the projection of the underlying grid, which is usually not geographic but unknown. 
Image could be spherical, in which case I guess they should be dealt as a special case with the source transformation being geodetic.
The images I have ever used have never been spherical.

So the points to evaluate would be:

- What is the definition of an image?
- What is expected according to the documentation?
- which solution simpler (as far as code goes)? 
- Is the interpolation in the source projection robust?
- What do people expect?


According to my experience, an image is gridded data with rectangular pixels of a constant size in the source projection. 

My assessment is:

- An image has rectangular pixels in source projection vs yours that the underlying coordinates of an image are always assumed to be spherical.
- The cartopy documentation mentions  that 'transform: :class:`~cartopy.crs.Projection` or matplotlib transform' is 'The coordinate system in which the given image is rectangular.' So my understanding of the documentation is that the image is rectangular in the transform (or source) projection, not on the sphere, unless transform is geodetic.
- Both solutions are simple.
- Interpolation, in the source projection means interpolating a bunch of irregular spaced points on a regular rectangular grid which is pretty robust versus interpolating in geocentric coordinates which may be less robust. Wrapping seems to be managed the same way by both interpolation.
- My understanding is that people expect square pixels in source projection versus your understanding that people don't expect square pixels in source projection.

You have an other point, that people should use pcolormesh, but pcolormesh has several issues which I haven't experienced with imshow and the fix tested (see code above).

### Wrapping and pixel shape
Wrapping works well and here is the result for your example of using PlateCarree() and PlateCarre(180). As you can see in the plots below made with the proposed fix, there is no need of any split, join. It just works as expected.
But if one doesn't use the fix, then the images will have square pixels in data projection and geodetic pixels in PlateCarree(180), which is a big issue. **The pixel should not change shape when we change the central meridian**. So obviously it won't work correctly when we add image with different central meridian, sometimes they will have square pixels and sometimes not.

Wrapping works well with the fix proposed as shown by the examples of the code above.
Below are the PlateCarree results.
With the fix:
Capture d’écran 2020-03-02 à 17 42 24

Capture d’écran 2020-03-02 à 17 42 30

Without the fix:
Capture d’écran 2020-03-02 à 17 43 56

Capture d’écran 2020-03-02 à 17 44 02

There is an issue in both the original version and the fix one, as the pixels are truncated south and north; so this is not relevant for the comparison between the two. it is a bug elsewhere.

## Conclusion

The algorithm doesn't seem to work better in geocentric coordinates versus in data projection. Both have the same truncation issue in my tests. 

There is no need to stitch and join the data when working in data projection and wrapping is correctly handled. The geocentric algorithm is inconsistent when going to the data central longitude to an other central longitude as the pixels go from rectangular to wobble. 

My general understanding and the definition of cartopy imshow lead me to consider that an image is expected to have rectangular pixels in data projection. 

Finally, the image I have encountered that needed regrinding were Météo France Radar images in North Polar Stereographic projection and geographic images. 
Both are images in projection not in geodetic coordinates.

So the proposed improvement works as well for wrapping, is more consistent when the central meridian changes and provides rectangular pixels in the source projection as expected by most.

pelson:
Thanks for the updates. I've spent a good amount of time working through my own assumptions and prior objections to the change. They boil down to the two following points:

 * The distance metric used for interpolation is changed from elliptical to cartesian (in the source image)
 * In this proposal the target coordinates must be transformed to the source
   coordinates, implying that:
    * the source coordinate system must now have an inverse definition
        * (note: this appears to already be the case with the clipping code further down the regrid function)
    * it is not possible to define an image in a different range to the
      projection (e.g. an image in 0-360 longitude, when the projection is -180-180 [a common case])

The first is the whole point of the original issue and this PR (square pixels, not great circles), and to be fair, is consistent with other cartopy interfaces (e.g. line projections). The second point, regarding an inverse transformation, appears to no longer be a problem with modern Proj4 versions (but I don't understand why). For example, we can define our own projection which
apparently has no inverse definition, yet can happily (continue) to be transformed even with the change proposed here.

The "image defined beyond the range of the projection" issue is real though, and needs to be addressed before we can proceed. I think you can simply transform the source coordinates into its own coordinate system to have the coordinates put into the appropriate range (nb: doing this effectively makes the image ragged, so the kd-tree optimisation mentioned below is not valid). An example of the failing case with this implementation:

```
import cartopy.crs as ccrs
import numpy as np

img_prj = ccrs.PlateCarree()
x_lim, y_lim = [160, 220], [-45, 45]
nx, ny = [6, 3]

# Define x and y coordinates at the center of each pixel.
# ys = np.linspace(y_lim[0], y_lim[1], ny)
ys, dy = np.linspace(y_lim[0], y_lim[1], ny, retstep=True, endpoint=True)
xs, dx = np.linspace(x_lim[0], x_lim[1], nx, retstep=True, endpoint=False)

# Generate an image that shows each pixel clearly.
extent = xs.min()-dx/2, xs.max()+dx/2, min(y_lim)-dy/2, max(y_lim)+dy/2
data = np.linspace(0, 100, nx*ny).reshape((ny, nx))

# Plot the image.
plt.figure(figsize=(10, 10))
ax.coastlines()
ax.set_global()
ax.imshow(data, transform=img_prj, extent=extent, origin='upper')

plt.show()
```

master:

pc_over_current

With this change:

pc_over_proposed


(It is worth pointing out that the cropping of the output is not necessarily the best with the existing implementation - it appears to be cropping at pixel centers rather than the extents themselves. But the implementation proposed here makes it much worse, as can be seen in the image above.)


Another good example of the incorrect clipping but this time the data is in a valid range:

```
img_prj = ccrs.RotatedPole(pole_latitude=50)
x_lim, y_lim = np.array(img_prj.x_limits)/5, np.array(img_prj.y_limits)/5
nx, ny = [6, 3]

# Define x and y coordinates at the center of each pixel.
# ys = np.linspace(y_lim[0], y_lim[1], ny)
ys, dy = np.linspace(y_lim[0], y_lim[1], ny, retstep=True, endpoint=True)
xs, dx = np.linspace(x_lim[0], x_lim[1], nx, retstep=True, endpoint=False)

# Generate an image that shows each pixel clearly.
extent = xs.min()-dx/2, xs.max()+dx/2, min(y_lim)-dy/2, max(y_lim)+dy/2
data = np.linspace(0, 100, nx*ny).reshape((ny, nx))

# Plot the image.
plt.figure(figsize=(10, 10))
ax.coastlines()
ax.set_global()
ax.imshow(data, transform=img_prj, extent=extent, origin='upper')

plt.show()
```

master:

rp_existing

This PR:

rp_proposed

The benefits of this change worth noting are:

 * We may no longer need to use a kdtree - we can use a cartesian nearest neighbour calculation, which would be significantly faster (although the kdtree implementation is already amazingly fast). 
Until we have the ability to toggle between distance metrics, I don't propose we try to optimise this yet.
    * Update: As mentioned previously, this isn't necessarily true if we re-project source pixel
      locations to the source projection in order to handle over-ranged input.
 * We get consistency with other APIs in that the coordinate system provided also controls the
   distance metric of the interpolation


Request for proposal for toggling the distance metric behaviour:

 * I would like to expose control of the distance metric to our users before proceeding with this breaking change. So far we have used the coordinate reference system as the way in which we define the distance metric. Changing the behaviour as proposed in this PR *is* consistent with other cartopy interfaces, and am happy for the default to change. However, a proposal for how to expose the distance metric used in the interpolation in an independent manner from the CRS of the ``extents``` coordinates would be worthwhile as part of this discussion/PR.

In summary, I've spent a long time trying to find cases which this change cannot handle
(incl. limited area source and/or target such as OSGB, projections apparently without defined inverse transformations). The only cases that I can find are those in which images extend beyond the limit of their source projection (common for PlateCarree to have images from 0-360) and those which don't preserve the clipping of the existing implementation (which itself is not perfect).

htonchia:
@pelson has made a comment about clipping.

As far as I can tell, the example 
```
import cartopy.crs as ccrs
import numpy as np

img_prj = ccrs.PlateCarree()
x_lim, y_lim = [160, 220], [-45, 45]
nx, ny = [6, 3]

# Define x and y coordinates at the center of each pixel.
ys, dy = np.linspace(y_lim[0], y_lim[1], ny, retstep=True, endpoint=True)
xs, dx = np.linspace(x_lim[0], x_lim[1], nx, retstep=True, endpoint=False)

# Generate an image that shows each pixel clearly.
extent = xs.min()-dx/2, xs.max()+dx/2, min(y_lim)-dy/2, max(y_lim)+dy/2
data = np.linspace(0, 100, nx*ny).reshape((ny, nx))

# Plot the image.
plt.figure(figsize=(10, 10))
ax.coastlines()
ax.set_global()
ax.imshow(data, transform=img_prj, extent=extent, origin='upper')

plt.show()
```
is not about regrid, as img_prj and the source projection are the same. And one would get the same results with `ax.imshow(data, extent=extent, origin='upper')`.
The issue is with the clipping and the wrapping extent. Of course wrapping uses regrid hence the issue.
This comes from the fact that the extent coordinates are not within the projection domain, so during interpolation the target coordinate is -160 while the source coordinate is at 200.

My point is that the source projection is the projection where the pixels are rectangular and which contains the image without wrapping. If the image is cut in two as above, then it cannot be the source projection. 
To work in the source projection I have modified @pelson accordingly:

```
import cartopy.crs as ccrs
import numpy as np

img_prj = ccrs.PlateCarree(central_longitude=180)
x_lim, y_lim = [160, 220], [-45, 45]
nx, ny = [6, 3]

# Define x and y coordinates at the center of each pixel.
ys, dy = np.linspace(y_lim[0], y_lim[1], ny, retstep=True, endpoint=True)
xs, dx = np.linspace(x_lim[0], x_lim[1], nx, retstep=True, endpoint=False)

# Generate an image that shows each pixel clearly.
extent = xs.min()-dx/2, xs.max()+dx/2, min(y_lim)-dy/2, max(y_lim)+dy/2
data = np.linspace(0, 100, nx*ny).reshape((ny, nx))
xy2 = img_prj.transform_points(
        ccrs.PlateCarree(),
        np.array(extent[0:2]),
        np.array(extent[2:]))
extent_img_proj = xy2[0,0], xy2[1,0], xy2[0,1], xy2[1,1]

# Plot the image.
plt.figure(figsize=(10, 10))
ax = plt.axes(projection=ccrs.PlateCarree(central_longitude=0))
ax.coastlines()
ax.set_global()
ax.imshow(data, transform=img_prj, extent=extent_img_proj, origin='upper')

plt.show()
```

which produces with the change proposed in the PR.
Capture d’écran 2020-03-23 à 16 51 43

Now it is possible to manage this in the regrid process, but I would rather stick with the source projection being the one containing the image without wrapping and with rectangular pixels.

greglucas:
Removing the geocentric pivot was added in this commit: https://github.com/SciTools/cartopy/pull/1808/commits/017f989f02078e8e41f60852713eb3760ef9b60b
Current master reproduces the images found within this PR. Closing as I don't think there is anything else that needs to be done here.

PR #1475, opened 2020-02-24
Better label rotation

stefraynaud:
## Rationale

Labels are rectangles with text and their rotation make them
to overlap the map boundary, which force the current algorithm
to try another rotation angle, and eventually leads to suprising
rotations. This occurs because the labels are also moved along
an ellipse that is defined xpadding and ypadding.

This patch make the labels to move along the rectangle defined by xpadding
and yappding, the one that encloses the ellipse. This
sligthly increases the padding which may be sufficient to prevent
labels to overlap boundaries in most situations.

## Implications

Labels should undergo less exagerated rotation in some circumstances.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -639,8 +639,23 @@ class Gridliner(object):
                         else matplotlib.rc_params['xtick.major.pad'])
             ypadding = (self.ypadding if self.ypadding is not None
                         else matplotlib.rc_params['ytick.major.pad'])
-            dx = ypadding * np.cos(angle * np.pi / 180)
-            dy = xpadding * np.sin(angle * np.pi / 180)
+
+            angle = ((angle+180) % 360 - 180) * np.pi / 180
+            tangle = np.tan(angle)
+            cangle = np.arctan2(xpadding, ypadding)
+            if abs(angle) < cangle:  # right
+                dx = ypadding
+                dy = dx * tangle
+            elif abs(angle) > (np.pi - cangle):  # left
+                dx = -ypadding
+                dy = dx * tangle
+            elif angle >= cangle:  # top
+                dy = xpadding
+                dx = dy / tangle
+            else:  # bottom
+                dy = -xpadding
+                dx = dy / tangle
+
             transform = mtrans.offset_copy(
                 self.axes.transData, self.axes.figure,
                 x=dx, y=dy, units='dots')
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2019, Met Office
+# (C) British Crown Copyright 2011 - 2020, Met Office
 #
 # This file is part of cartopy.
 #
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2019, Met Office
+# (C) British Crown Copyright 2011 - 2020, Met Office
 #
 # This file is part of cartopy.
 #


QuLogic:
Can you post an example of what effect this might have?

stefraynaud:
Here is a view extracted from a [figure](lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/gridliner_labels_inline.png) that is created by unitary tests.
cartopy-rotation

cartopy-rotation2

On the left, that actual behavior, and on the right, the proposed one. Globally: some of the labels are less rotated, and some others appear.

greglucas:
I actually think the addition makes the axes formatting look worse. I personally liked the rotations before... Maybe this is because of the extra locations though, so perhaps it just needs an updated locator now?

stefraynaud:
Here is one of the reason I proposed this PR: #1560

In fact, when can even go further by not trying alternate angles that induce confusion, but play only with the padding, and just not display labels that intersect the boundary with gridline angle.

friedrichknuth:
For this example

```
import cartopy.crs as ccrs

rotated_crs = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=70.0)

ax = plt.axes(projection=rotated_crs)
ax.set_extent((-6, 3, 48, 58), crs=ccrs.PlateCarree())
ax.coastlines(resolution='50m')

gl = ax.gridlines(draw_labels=True,x_inline=False, y_inline=False)
plt.show()
```

the proposed fix in this PR produces 

PR

instead of

current

The thought with #1560 is something closer to

suggested

but also with labels at the top and bottom of the plot, oriented similarly to the way labels seen at the right and left are being aligned in parallel to gridlines.

stefraynaud:
@friedrichknuth Aligning top and bottom labels with the gridline without 90° rotation to make it more readble is choice among others. For instance with this choice, longitude labels will not longer be horizontal in classic rectangular projections (like [this figure](https://scitools.org.uk/cartopy/docs/latest/_images/gridliner-2.png)).
I think, a solution that may work in most situation is:
* to always align the strictly with the gridlines,
* make by default text label as horizontal  as possible (which is already done)
* increase the padding until if the labels crosses the map boundary, until a max padding
* if the max padding is reached, try a 90° rotation of the label that may make it vertical.
* finally don't show the label if it still crosses the map boundary.

It's for debate.

Note:
* problems occur when the label size is large with respect to the padding
* the solution must work with [all projections](https://scitools.org.uk/cartopy/docs/latest/crs/projections.html) and map boundary types

friedrichknuth:
The issue is indeed tricky... in the example provided above

```
gl.top_labels = False
gl.right_labels = False
```
produces

plot

which helps clean things up. What remains are unexpected labels on the right side, despite the call for all labels on the right to be disabled. 

I think this is being discussed in part here https://github.com/SciTools/cartopy/issues/1530#. Perhaps there might be a lead here https://github.com/SciTools/cartopy/issues/1530#issuecomment-618806505 on how to mask right/left/top/bottom values based on if they are lat/lon values. In other words, if labels on a given side are being explicitly disabled, only preserve the most frequent type. 

Perhaps only displaying the most frequent type (lat or lon) on a given side can be the default? Not sure how this plays out in all possible projections/rotations, but it may resolve some of these alignment issues.

Another thought would be to implement a toggle for `gl.rotate_labels` by axes (e.g. . `gl.rotate_xlabels` and `gl.rotate_ylabels`. `False` would mean horizontal and `True` could mean for the text to be aligned parallel to the gridline angle.

stefraynaud:
In fact, the problem of label visibility is also tricky. For instance, what is a "right label" in this [figure](https://scitools.org.uk/cartopy/docs/latest/_images/sphx_glr_gridliner_002.png)?
Whatever, it should be better with the PR #1537 and #1538 that are currently under review.

As for the `rotate_labels` parameter, currently `True` means "we try to align it with the gridline", and `False` means "horizontal". As discussed above, this would be changed to "we strictly align it or hide it". It is also feasible as you suggest to allow a distinction between "x" labels rotation and "y" labels rotation.

stefraynaud:
No longer valid due to #1537

PR #1485, opened 2020-03-05
Update CircleCI config

QuLogic:
## Rationale

CircleCI is turning on Pipelines next week. I just flipped the switch early to confirm that everything is working.

Additionally, I dropped the printing of the link to the documentation in the build, because it stopped working. When you're already on the build page, it's easier to go through the artifacts tab at the top. For a quick link directly from the PR, I've added a GitHub Action that posts a status with the correct link (the Action knows the right way to get the link.) Note, it does note post here yet because the Action is not in this repo, but you can see it if you look at the commit in my repo:
image

## Implications

No surprises next week. A direct link to the built documentation in the commit/PR status.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,8 +32,7 @@ env-run:  &env-setup
 deps-run: &deps-install
   name: Install Python dependencies
   command: |
-    source activate test-environment
-    conda install --quiet \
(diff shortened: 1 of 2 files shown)


QuLogic:
For reference, if you look at the merge commit, you can find these artifact links there, so this should be working now:
image

PR #1466, opened 2020-02-18
Default lonlat formatter

stefraynaud:
## Rationale

Longitude and latitude labels have no meaning for projections other than `PlateCarree`.
However, the default label formatter is always of longitude and latitude degrees type, whetever the gridliner crs is.
A test is now added and the default formatter is of `ScalarFormatter` type as in MPL, except when the crs is `PlateCaree()`.

## Implications

This prevents displaying labels in degrees for projected gridlines.

PR #1492, opened 2020-03-22
Set the default label formatter to lon/lat only for PlateCarree

stefraynaud:
_This PR replaces #1466: I didn't successed in squashing the commits._

## Rationale

Longitude and latitude labels have no meaning for projections other than `PlateCarree`.
However, the default label formatter is always of longitude and latitude degrees type, whetever the gridliner crs is.
A test is now added and the default formatter is of `ScalarFormatter` type as in MPL, except when the crs is `PlateCaree()`.

## Implications

This prevents displaying labels in degrees for projected gridlines.


greglucas:
@stefraynaud it looks like there is a conflict now with the License header that needs to be resolved before merging it in.

greglucas:
I think you must be the most unlucky person with your PRs :)
CI is still failing, but I'm pretty sure this is unrelated and just needs to be rerun.
https://travis-ci.org/github/SciTools/cartopy/builds/674884514
Unfortunately, I can't hit restart on those jobs. Can you try to cycle it by force-pushing an empty commit?

QuLogic:
So I logged out and back in on travis-ci.org, restarted the job, and now magically it's posting here.

greglucas:
Thanks for this update @stefraynaud!

PR #1496, opened 2020-03-25
Add GeoQuadMesh to support set_array() with pcolormesh

greglucas:
## Rationale

I wanted to be able to dynamically update a map with new data using `set_array()` as typically done in MPL animations.

## Implications

- We will now draw a few more Polygon's than before if an initial masked array was passed in because we need to draw every single Polygon for future updating and array sizes being equal.
- We need to store the wrapping mask for future updates adding some additional overhead.


greglucas:
Closes #233 by adding the ability to call `set_array()` on a new GeoQuadMesh collection.

greglucas:
The global_wrap_3 image is currently failing because the boundary is not clipping the polygons on the boundary exactly to produce a black boundary line. Instead, the colors are being aliased out the edges. I think this is related to: https://github.com/matplotlib/matplotlib/issues/15952

QuLogic:
This doesn't fully fix #233 though, as `set_array` was just an example, and it was about all `set_*` methods.

greglucas:
True. I just updated the comment.

I just went and looked to see how many other [methods](https://matplotlib.org/3.2.1/api/collections_api.html#matplotlib.collections.QuadMesh) would need to be dispatched and decided I don't have a desire to implement those all right now. With a new subclassed collection and demonstration, someone should be able to take it from there if they need those extra capabilities and pretty easily extend it.

lib/cartopy/tests/mpl/test_pseudo_color.py
```diff
@@ -30,18 +30,6 @@
 from cartopy.tests.mpl import MPL_VERSION
 
 
-def test_pcolormesh_fully_masked():
```
greglucas, line 33: We are now plotting the entire mesh regardless, so that we can have the vertices for future `set_array` calls. I don't think this is a huge issue because someone can just check for fully masked data themselves beforehand and not issue the plot command if that is what they desire.

greglucas:
Looks like an unrelated test on Python 3.5 is failing here: `test_quiver_plate_carree`

I don't know if any packages were upgraded recently, but it doesn't make a lot of sense because this case should pin most of the minimum versions. If #1575 goes in, then this shouldn't be an issue since Python 3.6 is passing here.

dopplershift:
Looks like same test failure even after rebase.

greglucas:
Looking at it now, my guess is that this is something similar to #1628 would fix. I'm adding a test before the failing one that has an image comparison and perhaps the state is left in an awkward place in Python 3.5. I'm not sure what MPLs ImageComparison decorator is doing back then, but my guess is that the decorator has changed to fix this behavior in the newer versions which is why the Py3.6 version is passing.

dopplershift:
The problem is that master passed on Travis 2 hours ago, so *somehow*, this test failing is related to your PR.

greglucas:
I agree. I _think_ this would be fixed by that other PR, for better or worse since it means I didn't write a great test then :-/

greglucas:
🤦 It turns out that the previously failing test was never run before and had this decorator
`@pytest.mark.xfail(MPL_VERSION < '2.1.0', reason='Matplotlib is broken.')`, which I had stolen for a test above it and the diff's didn't show a removal because it was just moved up in the file.

PR #1505, opened 2020-04-03
Add Python 3.8 to list of supported Python versions

dopplershift:
I have at least one user report that pip on 3.8 doesn't want to install cartopy because it doesn't think it works on 3.8.


dopplershift:
Would be good to have this in for the next beta.

dopplershift:
There's absolutely no reason this should impact coverage.

PR #1512, opened 2020-04-10
Fix CI failures

greglucas:
Changing the plot directive to mpl's version.

**Edit:** CI was not making it past test_pseudo_color even though it was in an xfail, so I changed that to a skipif now instead to just avoid that potential hanging process.

## Rationale

Doc builds are failing and I think it is due to a recent update to sphinx 3.0 in CI. This is an attempt to remedy that.

## Implications

No more local copy of the plot_directive, so we're tied to mpl now.


QuLogic:
All the plots are _really_ small (and scaled up) now: https://1976-5282596-gh.circle-artifacts.com/0/docs/build/html/whats_new.html

greglucas:
Looks like that is because it was only building the thumbnails for those images and not the full-size pngs. I swapped the order in conf which seems to have solved the issue. It looks like the thumbnails are still made for the gallery images.

greglucas:
It looks like the reason Travis may not be showing up on the PRs anymore is that we are using the legacy '.org' hooks. Rather than the new GitHub Actions.
https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798

I think someone with admin privileges will have to move that integration over.

For this specific PR, Travis is now succeeding again: https://travis-ci.org/github/SciTools/cartopy/builds/673561526

QuLogic:
I requested it, but I think an org admin has to approve.

QuLogic:
I _think_ org admins would be @SciTools/admins.

rhattersley:
> I requested it, but I think an org admin has to approve.

Hopefully that's done now.

greglucas:
Thanks, @rhattersley! I see the new org over there now, but the builds are still occurring on the old system: https://travis-ci.org/github/SciTools/cartopy/builds

The new one is empty: https://travis-ci.com/github/SciTools/cartopy

It looks like maybe you'll also need to migrate the repository as well.
https://docs.travis-ci.com/user/migrate/open-source-repository-migration#the-migration-steps
I don't see any of those options under my account, so I assume it has to be an admin as well.

rhattersley:
> It looks like maybe you'll also need to migrate the repository as well.

The migration is proving tricky! Both the "Migrate" button on the [repo page](https://travis-ci.com/github/SciTools/cartopy) and the organisation's [migrate page](https://travis-ci.com/organizations/SciTools/migrate) now let me kick off the process but then nothing seems to happen...

rhattersley:
The ability to migrate is still only available as a beta ... I guess quirks should be expected.

greglucas:
Odd that it is still only in beta even though they seem to be forcing the changeover. I guess if you clicked the migrate then it should just "work". We'll see the next time someone submits a build if it worked :)
Thanks for doing that!

QuLogic:
It does seem to be stuck: https://travis-ci.community/t/cannot-migrate-repository/8110
I hit Migrate in two places and don't see anything happening.

QuLogic:
So FYI, I logged out and back in and restarted #1492, and it started posting statuses there. I still can't migrate to .com though.

PR #1515, opened 2020-04-11
Remove unused HeaderCheck from setup.py.

QuLogic:
## Rationale

It is now a real test, and the check here is very outdated.

## Implications

Less code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,68 +90,6 @@ def find_package_tree(root_path, root_package):
     return packages
 
 
-class MissingHeaderError(Exception):
-    """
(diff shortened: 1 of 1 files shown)


greglucas:
Coverage increased here, but is still showing a red 'x'. Is that just because it is such a low number at 40%? Also, I'm not sure if that is a true 40% or if it is combining the 0% from `/` and 80% from `lib/` in the tree.

QuLogic:
Running locally, it's 80%, so I wouldn't trust it. It will be fixed once we drop Python 2 support (which is the run adding the extra files.)

PR #1516, opened 2020-04-11
Remove support for Python 2

QuLogic:
## Rationale

Python 2 is quite a burden, so we'll be dropping it after the next release. Do not merge until it's out.

## Implications

No more Python 2.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,9 +3,6 @@ environment:
         - PYTHON_VERSION: "3.6"
           CONDA_INSTALL_LOCN: "C:\\Miniconda36-x64"
           PACKAGES: "cython numpy matplotlib-base proj pykdtree scipy"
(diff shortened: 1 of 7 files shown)


QuLogic:
So, coverage still seems to think that the installed version is important. It seems like only one file is hit there `cartopy/examples/__init__.py`, so if we can figure that one out, maybe the rest will all magically go away.

greglucas:
Linking #1237 as something we should look at again with dropping Python 2 support. Pyupgrade will need to be run again since there have been a lot of changes since then.

QuLogic:
Ah, I forgot about that PR. There are definitely some changes that could be taken from there, though I think these branches I made are for different things.

QuLogic:
This is now mergeable since there's a 0.18.x branch.

PR #1517, opened 2020-04-11
Remove six.

QuLogic:
## Rationale

Once Python 2 is dropped, we won't need `six`, so remove it.

## Implications

Less dependencies.


QuLogic:
Will fix stickler after #1516.

PR #1519, opened 2020-04-12
Update what's new for 0.18

QuLogic:
A couple more changes that people did, plus typo fixes. Also add note about @greglucas as a dev.

### DIFF
--- a/docs/source/whats_new.rst
+++ b/docs/source/whats_new.rst
@@ -79,7 +79,7 @@ Features
 
 * Elliott Sales de Andrade added a
   :class:`cartopy.mpl.geoaxes.GeoAxes.GeoSpine` class to replace the
-  :func:`cartopy.mpl.geoaxes.GeoAxes.outline_patch` that defines the map
+  :attr:`cartopy.mpl.geoaxes.GeoAxes.outline_patch` that defines the map
   boundary. (:pull:`1213`)
 
 * Ryan May fixed the Geostationary projection boundary so that geometries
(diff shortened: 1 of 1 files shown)


docs/source/whats_new.rst
```diff
@@ -10,16 +10,22 @@ For a full list of included Pull Requests and closed Issues, please see the
 Features
 --------
 
-* Kevin Donkers and Phil Elson made the AdaptiveScalar to the Natural
-  Earth Features the default. This will make the default features look
-  much nicer when plotting on zoomed in axes. (:pull:`1105`)
+* We are very pleased to announce that Greg Lucas has been added to the cartopy
```
greglucas, line 13: Should `Cartopy` be capitalized? I see below it is actually listed as `CartoPy`.

docs/source/whats_new.rst
```diff
@@ -10,16 +10,22 @@ For a full list of included Pull Requests and closed Issues, please see the
...
+  core development team. Greg (@greglucas) added the NightShade feature in the
```
greglucas, line 14: I'm now a 'maintainer', you're on the core development team I think. I'm not sure the qualifiers on everything though. But, perhaps put your new upgrade in as well?

docs/source/whats_new.rst
QuLogic: I do prefer to capitalize, but we aren't particularly consistent about it, it seems.

docs/source/whats_new.rst
QuLogic: I just copied the existing ones; it said the same thing when I was added. I don't think there are any announcements for maintainer->core dev I could copy.

QuLogic:
Please leave a comment if you want more changes (or open a PR), but I'll merge to tag the next beta.

PR #1522, opened 2020-04-13
Make the gridliner doc to better conform the new labbeling system

stefraynaud:
## Rationale

The gridliner, locators and formatters have been updated, so the doc must be updated to comply with the new features.

## Implications

The gridliner doc now has two differents exemples:

* The first one is new. It shows how to draw labels quickly.
* The second one has been updated to no longer use the LONGITUDE_FORMATTER and LATITUDE_FORMATTER.

The current web page: https://scitools.org.uk/cartopy/docs/latest/matplotlib/gridliner.html

### DIFF
--- a/docs/source/matplotlib/gridliner.rst
+++ b/docs/source/matplotlib/gridliner.rst
@@ -18,6 +18,25 @@ used to determine draw time behaviour of the gridlines and labels.
     :undoc-members:
 
 
+In this first example, gridines and tick labels are plotted in a
+non-rectangular projection, with most default values and
+no tuning of the gridliner attributes:
+
+.. plot::
+    :include-source:
+
+    import matplotlib.pyplot as plt
+    import cartopy.crs as ccrs
+
+    rotated_crs = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=70.0)
+
+    ax = plt.axes(projection=rotated_crs)
+    ax.set_extent([-6, 3, 48, 58], crs=ccrs.PlateCarree())
+    ax.coastlines(resolution='50m')
+    ax.gridlines(draw_labels=True, dms=True, x_inline=False, y_inline=False)
+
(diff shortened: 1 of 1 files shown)


docs/source/matplotlib/gridliner.rst
```diff
@@ -41,8 +61,9 @@ class to produce customized gridlines and tick labels:
     gl.left_labels = False
     gl.xlines = False
     gl.xlocator = mticker.FixedLocator([-180, -45, 0, 45, 180])
-    gl.xformatter = LONGITUDE_FORMATTER
-    gl.yformatter = LATITUDE_FORMATTER
+    gl.ylocator = LatitudeLocator()
+    gl.xformatter = LongitudeFormatter
```
dopplershift, line 65: Should this also be creating an instance?

stefraynaud: You are right. Just a mistake.

dopplershift:
I was hoping that would fix the circle doc failures. I don't think this PR is the cause--as far as I can tell the failure is due to problems making the Python 2 conda environment. @QuLogic thoughts?

QuLogic:
Python 2 docs are just slow because they have to solve a transaction that installs a bunch of old stuff. I just restart until they pass.

QuLogic:
I opened #1523 for the timeout stuff.

greglucas:
Looks good to me. Going to the docs I had to search to find the gridliner plots though. How would you feel about adding your first example to the gallery so it is easier to see how to use the gridliner right away. Maybe even two examples, or put two gridliners on a single plot. One with the gridliner in PlateCarree, and the other one in a different projection just to show that is possible and demonstrate it.

That can be a separate PR though.

dopplershift:
Not sure what's up with coveralls here--this is only a doc change. I'm just going to merge.

PR #1529, opened 2020-04-21
Add gridliner/ticklabels example to the gallery

stefraynaud:
## Rationale

An example added to the gallery to show the power of the new tick labeller
system.

This PR may be seen as an extension of #1522.

## Implications

So users clearly know how it can works.

Should we remove the other gallery example on the subjects?

### DIFF
--- a/lib/cartopy/examples/gridliner.py
+++ b/lib/cartopy/examples/gridliner.py
@@ -1,0 +1,40 @@
+"""
+Gridlines and tick Labels
+-------------------------
+
+Theses examples demonstrate how to quickly add longitude
+and latitude gridlines and tick labels on a non-rectangular projection.
+
+As you can see on the first example,
+longitude labels may be drawn on left and right sides,
+and latitude labels may be drawn on bottom and top sides.
+Thanks to the ``dms`` keyword, minutes are used when appropriate
+to exprimate fractions of degree.
+
+
+In the second example, labels are still drawn at the map edges
+despite its complexity.
+
+"""
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import matplotlib.pyplot as plt
+
+
+def main():
+
+    rotated_crs = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=70.0)
+
+    ax0 = plt.axes(projection=rotated_crs)
+    ax0.set_extent([-3, 2, 48, 52], crs=ccrs.PlateCarree())
+    ax0.add_feature(cfeature.LAND)
+    ax0.gridlines(draw_labels=True, dms=True, x_inline=False, y_inline=False)
+
+
+    plt.figure(figsize=(6.9228, 3))
+    ax1 = plt.axes(projection=ccrs.InterruptedGoodeHomolosine())
+    ax1.coastlines(resolution='110m')
+    ax1.gridlines(draw_labels=True)
+
+if __name__ == '__main__':
+    main()
--- a/lib/cartopy/examples/gridliner.py
+++ b/lib/cartopy/examples/gridliner.py
@@ -1,19 +1,20 @@
 """
-Gridlines and tick Labels
+Gridlines and tick labels
 -------------------------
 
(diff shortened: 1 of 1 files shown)


lib/cartopy/examples/gridliner.py
```diff
@@ -0,0 +1,40 @@
+"""
+Gridlines and tick Labels
+-------------------------
+
+Theses examples demonstrate how to quickly add longitude
```
QuLogic, line 5:
```suggestion
These examples demonstrate how to quickly add longitude
```
(and re-wrap)

lib/cartopy/examples/gridliner.py
```diff
@@ -0,0 +1,40 @@
...
+and latitude gridlines and tick labels on a non-rectangular projection.
+
+As you can see on the first example,
+longitude labels may be drawn on left and right sides,
+and latitude labels may be drawn on bottom and top sides.
+Thanks to the ``dms`` keyword, minutes are used when appropriate
+to exprimate fractions of degree.
```
QuLogic, line 12: exprimate? I don't know what word this is supposed to be.

stefraynaud: exprimante -> display?

lib/cartopy/examples/gridliner.py
```diff
@@ -0,0 +1,40 @@
...
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import matplotlib.pyplot as plt
+
+
+def main():
+
+    rotated_crs = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=70.0)
+    ax0 = plt.axes(projection=rotated_crs)
+    ax0.set_extent([-3, 2, 48, 52], crs=ccrs.PlateCarree())
```
QuLogic, line 29: This appears to be the middle of nowhere (as far as any details in the land shapefiles go); can you change it to somewhere that doesn't end up as a completely land-filled image?

Also, maybe set the land to use `100m` resolution, because otherwise it downloads 10m which is rather large.

lib/cartopy/examples/gridliner.py
```diff
@@ -0,0 +1,40 @@
...
+    ax0.add_feature(cfeature.LAND)
+    ax0.gridlines(draw_labels=True, dms=True, x_inline=False, y_inline=False)
+
+    plt.figure(figsize=(6.9228, 3))
+    ax1 = plt.axes(projection=ccrs.InterruptedGoodeHomolosine())
+    ax1.coastlines(resolution='110m')
+    ax1.gridlines(draw_labels=True)
```
QuLogic, line 36: Need to call `plt.show()` at the end or nothing appears.

lib/cartopy/examples/gridliner.py
```diff
@@ -0,0 +1,40 @@
...
+and latitude labels may be drawn on bottom and top sides.
+Thanks to the ``dms`` keyword, minutes are used when appropriate
+to display fractions of degree.
+
+
+In the second example, labels are still drawn at the map edges
+despite its complexity, and some others are also drawn within the map
+boundary.
+
+"""
```
QuLogic, line 19: Please add some `__tags__`.

lib/cartopy/examples/gridliner.py
stefraynaud: It was centered on the Channel. 
It is now moved sligthly westward with a greater extent that gives continents a not too crude shape with the 1:110m resolution.

lib/cartopy/examples/gridliner.py
stefraynaud: Fixed

lib/cartopy/examples/gridliner.py
stefraynaud: I added tags.

It should be approriate to have a predefined list of tags.
``"Lines and polygons"`` is often used in examples but is not clear of any user as for instance ``"Feature"``.

PR #1533, opened 2020-04-23
Add simple cache mechanism to GoogleWTS class

adrien-berchet:
## Rationale

During my development I made many tests implying fetching tiles from Google. In the end, my IP address got banned for some time...
I propose a very simple cache mechanism to store tiles so that it is not needed to request tiles from the server at each test.
With this, it is possible to write:
`gt_cache = cimgt.GoogleTiles(cache="\tmp\cartopy_cache")`
and all tiles downloaded downloaded by this `GoogleWTS` object will be stored in the given directory for later use.

## Implications

* Avoid being banned from tile providers during heavy load tests.
* Allow to work offline if the tiles have already been cached.

### DIFF
--- a/lib/cartopy/__init__.py
+++ b/lib/cartopy/__init__.py
@@ -5,6 +5,7 @@
 # licensing details.
 
 from ._version import version as __version__  # noqa: F401
+import tempfile
 
 __document_these__ = ['config']
 
@@ -17,9 +18,11 @@ import os.path
 _writable_dir = os.path.join(os.path.expanduser('~'), '.local', 'share')
 _data_dir = os.path.join(os.environ.get("XDG_DATA_HOME", _writable_dir),
                          'cartopy')
+_cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
 
 config = {'pre_existing_data_dir': '',
           'data_dir': _data_dir,
+          'cache_dir': _cache_dir,
           'repo_data_dir': os.path.join(os.path.dirname(__file__), 'data'),
           'downloaders': {},
           }
@@ -67,6 +70,7 @@ Keys in the config dictionary:
 
 del _data_dir
 del _writable_dir
+del _cache_dir
 
 
 # Try importing a siteconfig file which exposes an update_config function,
--- a/lib/cartopy/__init__.py
+++ b/lib/cartopy/__init__.py
@@ -56,6 +56,14 @@ Keys in the config dictionary:
     cartopy will download the appropriate file(s) to a subdirectory of this
     directory, therefore ``data_dir`` should be writable by the user.
 
+``cache_dir``
+    The absolute path to a directory where tiles data are cached when a
+    GoogleWTS sub-class is initialized with `cache=True`. If it is not found
+    cartopy will create it, therefore ``cache_dir`` should be writable by the
+    user. Note that the default cache dir might be accessible by all users,
+    depending on your OS and local configuration. If private cache is
+    mandatory, set cache_dir to a private location.
+
 ``repo_data_dir``
     The absolute path to the directory where the data delivered with the
     cartopy repository is stored.  Typically this will only be set by OS
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -20,9 +20,7 @@ using tiles in this way can be found at the
 from abc import ABCMeta, abstractmethod
 import concurrent.futures
 import io
-import json
 import os
-from uuid import uuid4
 import warnings
 
 from PIL import Image
@@ -52,7 +50,7 @@ class GoogleWTS(metaclass=ABCMeta):
         # osm may reject requests if there are too many of them, in which case
         # a change of user_agent may fix the issue.
         self.cache_path = cache_path
-        self.cache = {}
+        self.cache = set({})
         self._load_cache()
 
     def image_for_domain(self, target_domain, target_z):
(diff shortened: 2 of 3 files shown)


adrien-berchet:
I don't get why coverage dropped so much between the commits b86702c and dddbb63?

greglucas:
@adrien-berchet, thanks for the PR!
I think this has been a highly desired thing for quite some time. Linking in issue #732, that looks like it has a potential solution as well. I haven't had the chance to review that solution or yours though to see what the differences are.

Coverage has been pretty flaky recently and should hopefully be more reliable when we drop Python 2 support. So, I wouldn't read into it too much right now.

adrien-berchet:
Thanks for your feedback.
I did not see #732 which provides a possible solution for this issue, sorry about that. The idea of relying on file names is interesting so I updated this PR to consider this solution and the cache mechanism is now much more simple.
Also, I now save the Numpy array instead of the image to avoid some conversions.

adrien-berchet:
@greglucas I rebased the branch on master

QuLogic:
There was also #795, which doesn't seem to have been completed. I have not compared the implementations.

adrien-berchet:
> Overall, I think this looks really good. One question is whether we want to have a `cache_path` at all? Would it make sense to just put a kwarg `cache=True` and then use the `data_dir` that is already in Cartopy config:
> https://scitools.org.uk/cartopy/docs/latest/cartopy.html#cartopy.config
> Maybe make a new cache dir for your saved files underneath that? Then a user could easily clean up all local Cartopy files and not need to remember what cache_path they used.

Indeed it would be possible to store the cache in `data_dir` but I am afraid that the cache size explodes if it is global and if we don't implement a kind of rotation mechanism (keep only the 1000 last tiles and remove older ones for example).
For this reason I feel it is more relevant to have project specific caches. But we could setup a temporary cache when no cache directory is specified, so the global cache is automatically cleared at each reboot.
What do you prefer?

adrien-berchet:
> There was also #795, which doesn't seem to have been completed. I have not compared the implementations.

Indeed, I did not know it.
As far as I can see, it uses `shelf` to store the tiles while here we use `numpy` I/O functions. I never used `shelf` so I don't know which is better but I guess it is better to use `numpy` functions to store `numpy` arrays.

greglucas:
I have also never used `shelve`, so don't really have anything to compare it against, but it looks like it is more for an object store that you could reload and immediately access attributes/methods. Since you're just loading the data and not objects, I think the numpy save/load would be fine here.

What about adding a `cache_dir` key into the config dictionary? Each user could set that to whatever directory they'd like then, but maybe have a sensible default under or next to the data directory similar to Natural Earth?
https://github.com/SciTools/cartopy/blob/466c901f8f38ab4a26fd2736ccf49c98086de03b/lib/cartopy/__init__.py#L23-L27

Then we could change the keyword argument input to the function to just be cache and allow `cache=True` which would go to the global location, or an explicitly mentioned location too `cache=/tmp/dir`. I'm just thinking that some people may not want to explicitly put the directory every time? But, it would also be much safer to require the explicit handling than blowing up someones disk with a large cache too.

adrien-berchet:
I rebased on master and added a `cache_dir` entry in the config dict. I set a default value to a temp directory, so the cache is cleared at each reboot when no directory is specified. WDYT about it @greglucas ?

lib/cartopy/__init__.py
```diff
@@ -17,9 +18,11 @@
 _writable_dir = os.path.join(os.path.expanduser('~'), '.local', 'share')
 _data_dir = os.path.join(os.environ.get("XDG_DATA_HOME", _writable_dir),
                          'cartopy')
+_cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
 
 config = {'pre_existing_data_dir': '',
           'data_dir': _data_dir,
+          'cache_dir': _cache_dir,
```
greglucas, line 25: Add an entry into the docstring below describing what this new entry is for.

adrien-berchet:
In fact, when `cache=True`, it will create only one cache directory that is common to all sessions in which `cache=True` since I just get the path of global tmp dir (`/tmp`, `C:\temp` or whatever) and create the `cartopy_cache_dir` inside it. So only one is created and the cache is shared among different sessions (except for those in which a custom directory is specified). I think this way we have a good balance between ease of use and disk space.

If you are ok with that I will add the docstrings (I was waiting the final version to write the doc).

lib/cartopy/io/img_tiles.py
```diff
@@ -79,6 +88,21 @@ def fetch_tile(tile):
...
+        return os.path.join(
+            self.cache_path,
+            self.__class__.__name__
+        )
+
+    def _load_cache(self):
+        """Load the cache"""
+        if self.cache_path is not None:
+            if not os.path.exists(self._cache_dir):
+                os.makedirs(self._cache_dir)
```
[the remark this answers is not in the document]
adrien-berchet: We could indeed. But maybe only in the case where `cache=True`, because when the user explicitly sets a directory path we can suppose he knows what he is doing.

adrien-berchet:
Oh and I have another question @greglucas : since we provide a default path in a temp directory, should we set the default value of the cache parameter to True?

adrien-berchet:
Ok, thanks! Let's wait for other feedbacks then :)

lib/cartopy/tests/test_img_tiles.py
```diff
@@ -285,3 +288,91 @@ def test_ordnance_survey_get_image():
...
+
+    # Check that the warning is properly raised (only when cache is True)
+    if cache_dir is True:
+        assert len(w) == 1
+    else:
+        assert len(w) == 0
+
+    # Define expected results
+    x_y_z_f_h = [
+        (30, 18, 6, '30_18_6.npy', '545db25f1aa348ad85e1f437fd0db0d9'),
```
dopplershift, line 328: What's the thinking on checking the hashes? Won't this test break if Google updates the images?

adrien-berchet: I used hashes to check that the proper tiles are downloaded and that the cache is properly used when the same tiles are used a second time (this is why is update the cache images before the second call, so I can check that they are not downloaded again).
Yes it will break if these tiles change. But as far as I know, the tiles returned by the URL stored in `GOOGLE_IMAGE_URL_REPLACEMENT` should never change (it is already used in some other tests).

dopplershift: Fair enough.

greglucas:
Thank you for rebasing, I just reran the doc CI build that was failing and everything looks good now.

QuLogic:
I'm not so sure about tempdir, as that is shared by users, and this doesn't use a user-specific name, nor an exclusive creation mode. I think the user cache may be a better location? Though it's not consistent across platforms (`~/.cache/cartopy` on Linux, somewhere else on Windows.)

adrien-berchet:
Hi @QuLogic 
You're right, I didn't think about that. But is there any issue to share this cache among users? If someone is concerned about cache privacy, it is possible to set a custom cache path to a private location. Maybe we could just speak about this issue in the doc?

adrien-berchet:
Hi there, I just realized that this PR was still pending. I rebased and added a note about cache privacy in __init__.py.

greglucas:
:+1: My approval still stands. I think this is good with the note you've added, especially since this isn't on by default and the user has to opt-in.

dopplershift (approved):

PR #1534, opened 2020-04-23
Updates for MPL 3.3

### LINKED ISSUE #1470: Update pcolormesh with new matplotlib versions
### Description

The current master of matplotlib has changed the keyword arguments for `pcolor*`. See: https://github.com/matplotlib/matplotlib/pull/16258

We will need to change the `_pcolormesh_patched` function. Specifically, the call to `_pcolor_args` and the `shading` option and get rid of `allmatch` when that update from mpl is released.

Note: I didn't find any version checks and if statements to handle cases like this for MPL_VERSION in the code, only in the tests, so I'm also not sure how we want to handle this.

greglucas:
## Rationale

There are quite a few new deprecation warnings when running tests with the current MPL master (soon to be 3.3 release). In particular the pcolormesh now has a new sub-function call order in it, which we have overridden in the geoaxes implementation.

I also got rid of a few deprecation warnings that were showing up in the tests as well.

Closes #1470

## Implications

Fewer warnings when running on recent MPL versions and compatibility with future releases.

### DIFF
--- a/lib/cartopy/mpl/contour.py
+++ b/lib/cartopy/mpl/contour.py
@@ -52,7 +52,14 @@ class GeoContourSet(QuadContourSet):
             # list in-place (as the contour label code does in mpl).
             paths = col.get_paths()
 
-            data_t = self.ax.transData
+            # The ax attribute is deprecated in MPL 3.3 in favor of
+            # axes. So, here we test if axes is present and fall back
+            # on the old self.ax to support MPL versions less than 3.3
+            if hasattr(self, "axes"):
+                data_t = self.axes.transData
+            else:
+                data_t = self.ax.transData
+
             # Define the transform that will take us from collection
             # coordinates through to axes projection coordinates.
             col_to_data = col.get_transform() - data_t
@@ -64,7 +71,7 @@ class GeoContourSet(QuadContourSet):
 
             # The collection will now be referenced in axes projection
             # coordinates.
-            col.set_transform(self.ax.transData)
+            col.set_transform(data_t)
 
             # Clear the now incorrectly referenced paths.
             del paths[:]
--- a/lib/cartopy/mpl/contour.py
+++ b/lib/cartopy/mpl/contour.py
@@ -1,4 +1,4 @@
-# (C) British Crown Copyright 2011 - 2019, Met Office
+# (C) British Crown Copyright 2011 - 2020, Met Office
 #
 # This file is part of cartopy.
 #
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -441,7 +441,7 @@ class GeoAxes(matplotlib.axes.Axes):
                     self._autoscaleXon, self._autoscaleYon) = other
 
     @matplotlib.artist.allow_rasterization
-    def draw(self, renderer=None, inframe=False):
+    def draw(self, renderer=None, **kwargs):
         """
         Extend the standard behaviour of :func:`matplotlib.axes.Axes.draw`.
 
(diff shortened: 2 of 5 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -441,7 +441,7 @@ def hold_limits(self, hold=True):
                     self._autoscaleXon, self._autoscaleYon) = other
 
     @matplotlib.artist.allow_rasterization
-    def draw(self, renderer=None, inframe=False):
+    def draw(self, renderer=None):
```
QuLogic, line 444: This would break for old versions if someone passed `inframe`, no? We could use `**kwargs` for that.

lib/cartopy/tests/mpl/test_caching.py
```diff
@@ -151,6 +151,7 @@ def test_shapefile_transform_cache():
 
 
 def test_contourf_transform_path_counting():
+    plt.figure()
     ax = plt.axes(projection=ccrs.Robinson())
     ax.figure.canvas.draw()
```
QuLogic, line 156: Can use return from `plt.figure()` here instead of `ax.figure`.

lib/cartopy/mpl/geoaxes.py
greglucas: Yes, I was trying to match the call signature of the new `draw()` method, since that is what this is overriding. I added `**kwargs` to the call now and am passing on inframe if it is there for backwards compatibility, but I'm still not sure that is ideal because now we could have people passing in extra keyword arguments as well. I'm open to other thoughts here too.

Note this is probably the most annoying change of the MPL 3.3 update and what made me want to go through this because every pan/zoom will initiate this warning and flood the terminal. So, getting rid of it in some form is desirable.

QuLogic: We don't need to explicitly pass `inframe`, just pass `**kwargs` and Matplotlib will take care of the warning/error if the user passed `inframe`, and nothing will happen if they didn't.

greglucas: Ahh, the easy solution ;) don't know why I didn't do that. Thanks!

lib/cartopy/mpl/contour.py
```diff
@@ -52,7 +52,11 @@ def clabel(self, *args, **kwargs):
             # list in-place (as the contour label code does in mpl).
             paths = col.get_paths()
 
-            data_t = self.ax.transData
+            # The ax attribute is deprecated in MPL 3.3 in favor of
+            # axes. So, here we test if axes is present and fall back
+            # on the old self.ax to support MPL versions less than 3.3
+            data_t = getattr(self, "axes", self.ax).transData
```
QuLogic, line 58: Won't this still warn about `self.ax`, since parameters are evaluated before the function call?

greglucas: Interesting, I guess I didn't think that self.ax would be called before passing it in to a function. However, that appears not to be the case here. The warnings disappear for me in the test suite with this implementation.

My thought is that this would short circuit if "axes" is present in self, just like an if statement.

I was just trying to shorten up:
```python
if hasattr(self, "axes"):
    data_t = self.axes.transData
else:
    data_t = self.ax.transData
```
I can change it to be a little more explicit if desired.

QuLogic: Strange, it should definitely access it:
```
In [2]: class Foo: 
   ...:     @property 
   ...:     def bar(self): 
   ...:         print('accessing bar') 
   ...:     @property 
   ...:     def baz(self): 
   ...:         print('accessing baz') 
   ...:                                                                                                                                               

In [3]: f = Foo()                                                                                                                                     

In [4]: getattr(f, 'baz', f.bar)                                                                                                                      
accessing bar
accessing baz
```

greglucas: I must have had something messed up in my environment then... I have no idea why the warnings weren't being spit out.

I just updated the code to the if/else statement I put above to just be more explicit. Good catch!

dopplershift:
Looks like docs just timed out and I can't restart. Merging...

PR #1536, opened 2020-04-26
Drop Python 2 doc builds on CI

QuLogic:
## Rationale

This times out after 5 hours and I'm not too interested in fixing it.

## Implications

No doc builds on Python 2, but they stopped working anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,28 +85,6 @@ jobs:
       - store_artifacts:
           path: docs/build/html
 
-  docs-python2:
-    docker:
-      - image: continuumio/miniconda:latest
-    steps:
-      - checkout
-
-      - run: *apt-install
-      - run:
-          <<: *env-setup
-          environment:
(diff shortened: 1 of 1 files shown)


QuLogic:
Now it just times out after 5 hours, so instead I'm just going to remove it.

greglucas:
Are you still able to manually upload the Python 2 docs for the RC? We should have some docs for this final release still.

QuLogic:
I don't think there should be any difference between the two.

greglucas:
I just figured they were both uploaded to readthedocs or pages wherever they are hosted. I haven't looked at py2 docs for anything in a longgg time now though.

This is fine with me. It would be removed shortly anyways.

dopplershift:
For the record, I'm pretty sure the Python 2 builds of the docs weren't uploaded anywhere, it was just used to ensure examples and what not were functional on Python 2.

PR #1538, opened 2020-04-27
Use boundary to compute axes domain in gridliner

### LINKED ISSUE #1443: Gridlines not plotted correctly with Mollweide and Orthographic projections
Gridlines are not being plotted correctly with several projections, including Mollweide and Orthographic. The problem is that the gridlines to not continue all the way to pole, but simply stop at something like 85 degrees.

Here is some simple code to demonstrate the problem
```
fig = plt.figure()

```

stefraynaud:
## Rationale
``Gridliner._axes_domain`` discretize the unitary domain to find mi/max of longitudes and latitudes.
This discretization makes appriximative the estimate if bounds.

This PR directly uses native projection boundary to find min and max.
The also is quite simple and I may have missed some subtleties.

I didn't updated the figures since alot of them are modified and I'm waiting for feedback.

## Implications
Gridlines are properly plotted until the boundary.

Fixes #1443

lonlat_range

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -826,56 +826,16 @@ class Gridliner(object):
 
     def _axes_domain(self, nx=None, ny=None):
         """Return lon_range, lat_range"""
-        DEBUG = False
 
         transform = self._crs_transform()
 
-        ax_transform = self.axes.transAxes
-        desired_trans = ax_transform - transform
+        # Boundary in desired transform
+        lonlats = (self.axes.patch.get_transform() -
+                   transform).transform(self.axes.patch.get_path().vertices)
 
-        nx = nx or 100
-        ny = ny or 100
-        x = np.linspace(1e-9, 1 - 1e-9, nx)
-        y = np.linspace(1e-9, 1 - 1e-9, ny)
-        x, y = np.meshgrid(x, y)
-
-        coords = np.column_stack((x.ravel(), y.ravel()))
-
-        in_data = desired_trans.transform(coords)
-
-        ax_to_bkg_patch = self.axes.transAxes - self.axes.patch.get_transform()
-
-        # convert the coordinates of the data to the background patches
-        # coordinates
-        background_coord = ax_to_bkg_patch.transform(coords)
-        ok = self.axes.patch.get_path().contains_points(background_coord)
-
-        if DEBUG:
-            import matplotlib.pyplot as plt
-            plt.plot(coords[ok, 0], coords[ok, 1], 'or',
-                     clip_on=False, transform=ax_transform)
-            plt.plot(coords[~ok, 0], coords[~ok, 1], 'ob',
-                     clip_on=False, transform=ax_transform)
-
-        inside = in_data[ok, :]
-
-        # If there were no data points in the axes we just use the x and y
-        # range of the projection.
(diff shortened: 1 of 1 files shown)


stefraynaud:
I think this approach cannot be used alone, since it will never work for polar plot for example.
Instead, it but to used to make the base algo more robust.

stefraynaud:
Now it works well with the best of the two algo.

dopplershift:
Could you add the image as a test?

lib/cartopy/mpl/gridliner.py
```diff
@@ -833,12 +833,17 @@ def _axes_domain(self, nx=None, ny=None):
...
         x = np.linspace(1e-9, 1 - 1e-9, nx)
         y = np.linspace(1e-9, 1 - 1e-9, ny)
         x, y = np.meshgrid(x, y)
 
+        xyp = (self.axes.patch.get_transform()-ax_transform).transform(
+            self.axes.patch.get_path().vertices)
+        x = np.concatenate((x.ravel(), xyp[:, 0]))
+        y = np.concatenate((y.ravel(), xyp[:, 1]))
+
         coords = np.column_stack((x.ravel(), y.ravel()))
```
QuLogic, line 847: The `.ravel()` should no longer be needed here after the previous step.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -114,6 +114,14 @@ def test_gridliner():
                         top=1 - delta, bottom=0 + delta)
 
 
+@ImageTesting(['gridliner_orthographic'],
+              tolerance=0.2)
```
QuLogic, line 118: Why the non-default tolerance?

QuLogic:
This may need a rebase to fix CI.

lib/cartopy/mpl/gridliner.py
stefraynaud: `x` and `y` are 2D after meshgrid, so we to ravel them before concatenation.

lib/cartopy/tests/mpl/test_gridliner.py
stefraynaud: Don't know!
So I removed the tolerance keyword.

lib/cartopy/mpl/gridliner.py
QuLogic: But you ravel them on 844/845 already.

stefraynaud: I fixed it.

greglucas:
It looks like you've brought some duplicate commits into this. Can you try to rebase onto master instead of merging master into this branch? I'm not sure why Appveyor failed on the gridliner image tests, but Travis didn't. Maybe it will fix itself with a rerun of the CI...

stefraynaud:
@greglucas how to proceed with this rebasing?

greglucas:
I think you have to delete the commits through an interactive rebase, or at least that's how I've done it when things get out of sync between branches.

1. checkout master and fetch/pull the latest master commits from upstream
2. checkout this branch
3. git rebase -i master

That should give you a list of all the commits to choose from and then just keep the ones you want and delete the others. Probably just keeping all of your commits and removing all the merges/other people's commits from the list.

PR #1539, opened 2020-04-28
Update all remaining license headers.

QuLogic:
## Rationale

We're about to remove all `__future__` imports, so might as well drop the years first.

## Implications

Simpler headers.

### DIFF
--- a/lib/cartopy/tests/test_coding_standards.py
+++ b/lib/cartopy/tests/test_coding_standards.py
@@ -19,41 +19,19 @@ import pytest
 import cartopy
 
 
-LICENSE_TEMPLATE = """
-# (C) British Crown Copyright {YEARS}, Met Office
-#
-# This file is part of cartopy.
-#
-# cartopy is free software: you can redistribute it and/or modify it under
-# the terms of the GNU Lesser General Public License as published by the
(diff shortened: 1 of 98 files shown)


lib/cartopy/tests/test_coding_standards.py
```diff
@@ -133,52 +76,33 @@ def test_license_headers(self):
...
-                elif last_change.year > years[1]:
-                    print('The file header at {} is out of date. The last'
-                          ' commit was in {}, but the copyright states it'
-                          ' was {}.'.format(fname, last_change.year,
-                                            years[1]))
-                    failed = True
-
-        if failed:
-            raise ValueError('There were license header failures. See stdout.')
+        assert failed == [], 'There were license header failures.'
```
greglucas, line 105: I think it would help to print the names of the files that failed still. Is that somewhere else, or could you add it in here with just a `.format(failed)` or similar appended to the string.

QuLogic: Pytest will print it out as part of showing that the list is not empty.

greglucas: Ahh, yes! I forgot it would print the comparison variables. Thanks for clarifying.

PR #1542, opened 2020-04-29
add hexbin support

### LINKED ISSUE #884: Support matplotlib hexbin()
Current cartopy does not seem to provide `hexbin` support.  I assume this isn't a simple implementation like how `geoaxes.scatter` works or else it would have been done by now?  I will take a swing at this and see how "bad" things look on the various projections!

akrherz:
closes #884

## Rationale

Allows cartopy users to use matplotlib's `hexbin`

## Implications

The hexbins constructed are in the projection of the axes and **not** that of the source data projection.

This PR appears to expose a bug with how cartopy computes the axes extent when provided longitude values between 0 and 360.  When provided values between -180 and 180, the axes extent covers the globe as expected.  I would think fixing that issue would come with a separate PR?


lib/cartopy/mpl/geoaxes.py
```diff
@@ -1547,6 +1547,32 @@ def scatter(self, *args, **kwargs):
...
+    def hexbin(self, x, y, *args, **kwargs):
+        """
+        Add the "transform" keyword to :func:`~matplotlib.pyplot.hexbin'.
+
+        Other Parameters
+        ----------------
+        transform
+            A :class:`~cartopy.crs.Projection`.
+        """
+        t = kwargs.get('transform', None)
```
greglucas, line 1560: What about just popping this here and doing the `.transform_points()` function every single time?

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -132,6 +132,36 @@ def test_global_scatter_wrap_no_transform():
...
+@ImageTesting(['global_hexbin_wrap'],
+              tolerance=12.4 if MPL_VERSION < '2.1.0' else 0.5)
+def test_global_hexbin_wrap_new_transform():
+    ax = plt.axes(projection=ccrs.PlateCarree())
+    # By default the coastline feature will be drawn after patches.
+    # By setting zorder we can ensure our hexbin polygons are drawn
+    # after the coastlines.
+    ax.coastlines(zorder=0)
+    x, y = np.meshgrid(np.linspace(0, 360), np.linspace(-90, 90))
+    data = np.sin(np.sqrt(x ** 2 + y ** 2))
```
greglucas, line 145: How would you feel about changing the image data to: `sqrt((x - 180)**2 + y**2)`? This makes the plot look nice at least I think because then the wrapped values don't show a discontinuity due to bin edges being weird.

greglucas:
One more thought. Could you also add an example to the gallery showing it working with a few different projections?

lib/cartopy/mpl/geoaxes.py
akrherz: should be done now.

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: I ended up going to x values between -180 and 180, so I think remediates your suggestion.

akrherz:
> One more thought. Could you also add an example to the gallery showing it working with a few different projections?

That would be nice, I just need to grok how the gallery works and what type of example data to feed it.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -130,6 +130,37 @@ def test_global_scatter_wrap_no_transform():
...
+@pytest.mark.natural_earth
+@ImageTesting(['global_hexbin_wrap'],
+              tolerance=12.4 if MPL_VERSION < '2.1.0' else 0.5)
+def test_global_hexbin_wrap_new_transform():
+    ax = plt.axes(projection=ccrs.PlateCarree())
+    # By default the coastline feature will be drawn after patches.
+    # By setting zorder we can ensure our hexbin polygons are drawn
+    # after the coastlines.
+    ax.coastlines(zorder=0)
+    x, y = np.meshgrid(np.linspace(-180, 180, 200), np.linspace(-90, 90, 200))
```
greglucas, line 142: The problem is that now you aren't testing the "wrapping" capabilities here.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -130,6 +130,37 @@ def test_global_scatter_wrap_no_transform():
...
+    data = np.sin(np.sqrt(x**2 + y**2))
+    plt.hexbin(
+        x.flatten(),
+        y.flatten(),
+        C=data.flatten(),
+        transform=ccrs.PlateCarree(),
+        alpha=0.2,
```
greglucas, line 149: Could you also lower the bin size so we can see these are hexes on the global map still? The default of 100 seems quite large to me.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -130,6 +130,37 @@ def test_global_scatter_wrap_no_transform():
...
+    )
+
+
+@pytest.mark.natural_earth
+@ImageTesting(['global_hexbin_wrap'],
```
greglucas, line 154: Probably only need one "good" image that you can compare multiple different methods against? Just make sure your expected data is always the same in each case and that the coordinates are what is changing being passed in.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1540,6 +1540,29 @@ def scatter(self, *args, **kwargs):
         self.autoscale_view()
         return result
 
+    @_add_transform
+    def hexbin(self, x, y, *args, **kwargs):
+        """
+        Add the "transform" keyword to :func:`~matplotlib.pyplot.hexbin'.
```
greglucas, line 1546: In one of your earlier comments you mentioned that the data will always be projected to the map projection and hexes will always be drawn in the map projection (rather than the hexes themselves being transformed). Capturing that in the docstring would be good to let people know as well.

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: I am unsure how to do this as this exposes the problem with how matplotlib computes a grid size, if I reduce the points, only whitespace appears between the hexagons like so:

result-global_hexbin_wrap

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: I dropped the transform test as everything is being transformed now.

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: I am unsure how to specify the data that matches -180 to 180 in `PlateCarree`, the previous 0 to 360 would represent different points.

lib/cartopy/tests/mpl/test_mpl_integration.py
greglucas: I think I caused a mixup in terms. I meant the gridsize argument to MPL, not the gridsize of the numpy arrays. So, adding an extra argument to your call list: `gridsize=20,` produces something like this for me (still keeping your 200 x 200 numpy size array).

test_884

lib/cartopy/tests/mpl/test_mpl_integration.py
greglucas: Yeah, you will have to change your data calculation to make sure that the xs are mapped the same between the two cases. Something similar to this should work for those purposes I think.
```python
    x_wrap = np.where(x > 180, x-360, x)
    data = np.sqrt(x_wrap ** 2 + y ** 2)
```

greglucas:
Looks like you still have two images in the commits and removed the wrapping tests. I think there should be a standard -180 -> 180 plot that you compare everything against. Probably 2 or 3 test cases that should all produce the same image.
1.  axes: PlateCarree, transform = PlateCarree, x=-180, 180
2. axes: PlateCarree, transform = None, x=-180, 180
3. axes: PlateCarree, transform = PlateCarree, x=0, 360

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: Hmmm, I don't understand this plot, it shows 20 hexagons in the horizontal, but there should be one hexagon per data point right?  Maybe I am way over my head here with what matplotlib hexbin is doing :(  Maybe the issue is that dx != dy in my example, more boggling.  Thanks for all your effort here!

greglucas: No problem at all. I'm glad you're sticking with it! The hexbin plot is binning the data like a 2d histogram. So, each of these hexagons is summing all of the weighted data within each region, so if you applied no data argument then it would be the sum of the number of points within each hex. The gridsize parameter then tells MPL to fit 20 hexagons in the horizontal and however many are needed vertically to keep them regular I believe.

akrherz: Thanks for your patience with me, I think I have `Voroni` polygons in my head here with these hexbins, which they are not :(

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: Eh, I can't make it work out to the same resulting image.  There are data differences and the coastlines are even slightly different :(

result-global_hexbin_wrap-failed-diff

greglucas: Hm... That is odd. I wonder if the data limits aren't being updated properly so the extents of the images are slightly different? That also looks like the values for data weren't mapped in the transform properly either. I won't be able to take a closer look at this until this weekend or early next week.

akrherz: I think the issue is what is happening at the anti-meridian.

x range | transformed `xmin` | transformed `xmax`
--- | --- | ---
range(0, 360) | -179 | 180
range(-180, 180) | -180 | 179

Oh life is fun! `-180` appears to be the tough x value to get after transformation.  So I have pushed a slight jitter to the x values to prevent the anti-meridian from getting transformed.

:eyes: Travis-CI :)

greglucas: The transform must choose +180 when wrapping points, but if -180 is present doesn't transform it. Interesting, I didn't ever realize that. Your second case is leaving off the +180 point because range doesn't include the last value.

Rather than an offset of 0.1, I think you could use `arange(-179, 181)` and `arange(0, 360)` That way everything is just defined on the -179 -> 180 points in both cases...?

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -129,6 +129,41 @@ def test_global_scatter_wrap_no_transform():
...
+              tolerance=27.6 if MPL_VERSION < '2.1.0' else 0.5)
+def test_global_hexbin_wrap_transform():
+    ax = plt.axes(projection=ccrs.PlateCarree())
+    ax.coastlines(zorder=2)
+    x, y = np.meshgrid(np.arange(0.1, 360, 1.), np.arange(-90, 91))
+    x_wrap = np.where(x >= 180, x-360, x)
```
greglucas, line 156: Put a short comment here indicating why this wrap is needed for the data to be the same.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1537,6 +1537,32 @@ def scatter(self, *args, **kwargs):
...
+
+        The provided points are transformed into the axes projection and then
+        hexagons are constructed based on those transformed points.
```
greglucas, line 1546: Minor wording suggestion.
```suggestion
        The points are first transformed into the projection of the axes and then
        the hexbin algorithm is computed using the data in the axes projection.
```

lib/cartopy/tests/mpl/test_mpl_integration.py
akrherz: very good, thank you.

greglucas:
Thanks @akrherz!

PR #1543, opened 2020-04-29
Corrected behavior of 2D alpha array

kdpenner:
## Rationale

imshow() does not warp a 2D alpha array; it passes the unwarped alpha array to matplotlib's imshow(), which results in the display of an unwarped image.

## Implications

With these changes imshow() passes a warped 2D alpha array.


lib/cartopy/mpl/geoaxes.py
```diff
@@ -1276,14 +1276,29 @@ def imshow(self, img, *args, **kwargs):
             regrid_shape = self._regrid_shape_aspect(regrid_shape,
                                                      target_extent)
             warp_array = cartopy.img_transform.warp_array
+            unmodified_extent = extent
```
greglucas, line 1279: Can't you still pass extent in down below in the alpha route and not add this new variable?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1276,14 +1276,29 @@ def imshow(self, img, *args, **kwargs):
...
-                                     source_extent=extent,
+                                     source_extent=unmodified_extent,
                                      target_proj=self.projection,
                                      target_res=regrid_shape,
                                      target_extent=target_extent,
                                      mask_extrapolated=True,
                                      )
+            alpha = kwargs.pop('alpha', None)
+            alpha_arr = np.array(alpha)
+            if alpha_arr.ndim == 2:
```
greglucas, line 1290: Could you put `if np.array(alpha).ndim == 2:` here and then just keep using alpha below. Then you wouldn't need the `else` statement at all and could just set `kwargs['alpha'] = alpha` after the if statement.

kdpenner:
> Overall I think this update makes sense. Do you have a test case you can add to show that this does what you are expecting with 2d alpha?

Yes.  Here is a test with no warping of alpha:

t

Here is a test with warping of alpha:

t1

lib/cartopy/mpl/geoaxes.py
greglucas: I just realized extent is returned from the first warp_array call below 🤦Maybe change this to `original_extent`?

kdpenner: No; "extent" is used as input in line 1281 and replaced as output in line 1279, and I haven't tracked where it's used in subsequent code.

greglucas:
Yep, I know what you were getting at and the improvement :) I meant to add that under the `tests/` directory. Can you add something like that to make sure that your code is working and someone doesn't break it in the future?

lib/cartopy/mpl/geoaxes.py
kdpenner: OK---modified in latest commit.

lib/cartopy/mpl/geoaxes.py
kdpenner: OK---will change to `original_extent`.

kdpenner:
Sure, I'll make a test for the next commit.

kdpenner:
Writing a "good" test is harder than I expected. (As always.)

My thought was to test select alphas from the output of `imshow`. The `AxesImage` class has a method, `make_image`, which returns an RGBA array. Its values are integers between 0 and 255. Alpha arrays should have floats between 0 and 1. And the transformation between image value and RGBA value requires both a normalization and a colormap. I think I can't assert that the A in the RGBA output is the alpha from the input.  Ideas?

greglucas:
You could get the data out of the returned object like:
```python
image = ax.imshow(...)
arr = image.get_array()
alpha = image.get_alpha()

assert arr.shape == alpha.shape
```
Then you can make sure that the regridded data is the shape expected and that alpha is the same shape as that, which is the real issue here I think. You probably don't need an explicit image if you just want to throw in some fake data and some fake alphas that you could test this against too.

kdpenner:
I agree that the shape of the alpha array is the issue---`get_array` and `get_alpha` are undocumented for an `AxesImage` object, so I overlooked them. Your proposed test will work, I'll write it up.

greglucas:
I didn't realize those methods were undocumented! It looks like other people have run into the issue with a lack of documentation on that class as well: https://github.com/matplotlib/matplotlib/issues/15503

kdpenner:
I think the rebase worked. (I haven't done it before.)

Issue #1518, opened 2020-04-12
Adopt NEP29: minimum dependency policy


QuLogic:
After Cartopy 0.18, we will be dropping support for Python 2. Currently, for Python 3, we have a minimum of 3.5. This is somewhat problematic on conda-forge because they've already dropped support for it.

For the future, we should implement something like [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) as a minimum dependency policy. For Python, this is a 42-month window and for NumPy, it is a 24-month window.

Assuming that 0.18 is out by the end of the month, and 0.19 takes at least 3 months, that'd be end of July. Extending that policy to other dependencies using 42 months (January 30, 2017), that would mean:
* Python 3.7
* NumPy 1.16
* Cython 0.26 (older than current requirement)
* Proj 5.0.0
* GEOS 3.6.2
* Shapely 1.6b4
* pyshp 1.2.11 (older than current requirement)
* pyepsg 0.3.2 (older than current requirement)
* OWSLib 0.15.0
* Pillow 4.1.0
* Matplotlib 2.0.1
* GDAL 2.2.0
* SciPy 0.19.0
* pytest 3.0.7 (older than current requirement)

Extending that policy to other dependencies using 24 months (July 30, 2018) instead would mean:
* Python 3.7
* NumPy 1.16
* Cython 0.28.5
* Proj 5.2.0
* GEOS 3.6.3
* Shapely 1.7a1
* pyshp 2.0.0
* pyepsg 0.4.0
* OWSLib 0.17.0
* Pillow 5.3.0
* Matplotlib 2.2.3
* GDAL 2.3.2
* SciPy 1.2.0
* pytest 3.7.0

Though note, we probably shouldn't switch minimums to alphas or betas; that's just there for reference. We may also want slightly different policy for system deps like Proj and GEOS.

greglucas:
I'm all for this and would prefer the 24-month version. Python 3.5 is getting pretty old at this point even and we aren't even testing against Python 3.7 yet and this is a minimum version in your findings here.

dopplershift:
I’d be good to go with things just like NEP 29: Longer for Python itself (so we’d be going back to 3.7) and 24 months for everything else.

QuLogic:
Using 24 months for around now (Sept ~15, 2019) instead would mean:

- [x] Python 3.7
- [ ] NumPy 1.18
- [ ] Cython 0.29.13
- [ ] Proj 6.2.0
[...]

PR #1545, opened 2020-04-30
Switch to setuptools-scm instead of versioneer.

QuLogic:
## Rationale

Versioneer seems to break with PEP517 and newer packaging tools. The supported tool is setuptools-scm.

## Implications

Better-working installs.


greglucas:
Overall looks good to me and a positive to get rid of the forked versioneer code, but I'm not an expert on packaging at all. Do we need to pin setuptools_scm to a minimum version or anything like that to still work with Python 3.5 for now?

dopplershift:
I like the change. This is what I've switched to elsewhere.

QuLogic:
Well, CI passes, so I guess 3.5 should be fine? Anyway, if we do #1518, we won't have to worry about 3.5.

PR #1548, opened 2020-05-01
Prepare docs for 0.18

QuLogic:
Exactly what the title says.


QuLogic:
Thanks, did it from the command-line since I had to update the date in a couple places.

Issue #1554, opened 2020-05-08
Return list of Text objects from GeoContourset clabel() method


mhearne-usgs:
### Description

With the upgrade to cartopy 0.18, the behavior of labeling contour lines seems to have changed. Whereas before when I called ax.clabel(), I got back a list of Text instances, now I get None. 

I've traced through the code, and found that the GeoContourset clabel() method does not return anything. If you modify the final line of that method to look like:

`return super().clabel(*args, **kwargs)`

It appears to solve the issue I am seeing. I'll be pushing up a PR soon - hopefully I do that in accordance with your git workflow.

#### Code to reproduce

```

```

#### Traceback 

```

```

<details>
 <summary>Full environment definition</summary>

### conda list

```
```

### pip list

```
```

</details>

QuLogic:
Fixed by #1555.

PR #1555, opened 2020-05-08
Now returning the list of Text instances from parent class clabel() m…

mhearne-usgs:
Closes Issue #1554 

Now returning the list of Text instances from parent class clabel() m…ethod. This preserves the behavior of the MPL Axes clabel() method, which promises to return such a list.


mhearne-usgs:
btw, I've also signed the CLA

QuLogic:
Thanks!

PR #1565, opened 2020-05-19
Avoid reusing the kwargs variable name in draw

### LINKED ISSUE #1563: .add_image very poor quality
### Description

The .add_image command provides a map at very low quality. Can it be improved?
example_Stamen

Similar to post #1048 which has been closed -- the solution there is to add `interpolation='spline36', regrid_shape=2000` as args in `.add_image()` but this does not work any more.

#### Code to reproduce

```
import cartopy
import cartopy.io.img_tiles as cimgt

# Setting projection and ploting the second tilt
projection = ccrs.LambertConformal(central_latitude=39.3,
                                   central_longitude=-101.7)


fig = plt.figure(figsize=(20,20))
ax = fig.add_subplot(1, 1, 1, projection=projection)
ax.set_extent((-104.7, -98.7, 36.3, 42.3))

# Map features
ax.add_image(cimgt.Stamen('toner'), 8)

plt.savefig("example.png", format='png', bbox_inches="tight", dpi=500)

plt.show()
```

#### Traceback 

```

```

<details>

example_OSM
example_Stamen

### conda list

```
```

### pip list

```
```

</details>

ajdawson:
## Rationale

This fixes #1563 by changing the now duplicate variable name `kwargs` in `GeoAxes.draw` so that factory keyword arguments do not overwrite top-level keyword arguments that we pass on to matplotlib. Currently users will see a traceback if you pass a cartopy-specific keyword to an image factory because the factory keywords overrid the draw-level keywords and are mistakenly passed on to matplotlib which doesn't recognise them.

This was introduced in 210ee4becd63d1410e7260533c7412d8284bfb1b (#1534). I have remedied by prefixing factory args/kwargs with `factory_` as it seems the simplest and easiest to read way of doing it.

## Implications

Fixes #1563.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -469,11 +469,12 @@ class GeoAxes(matplotlib.axes.Axes):
         #       caching the resulting image;
         #       buffering the result by 10%...;
         if not self._done_img_factory:
-            for factory, args, kwargs in self.img_factories:
+            for factory, factory_args, factory_kwargs in self.img_factories:
                 img, extent, origin = factory.image_for_domain(
-                    self._get_extent_geom(factory.crs), args[0])
+                    self._get_extent_geom(factory.crs), factory_args[0])
                 self.imshow(img, extent=extent, origin=origin,
-                            transform=factory.crs, *args[1:], **kwargs)
+                            transform=factory.crs, *factory_args[1:],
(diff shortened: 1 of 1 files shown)


QuLogic:
Do tests not exercise this?

ajdawson:
It seems not

QuLogic:
That would happen with a normal merge-up, though adding tests would be good.

PR #1575, opened 2020-05-26
[WIP] Updating minimum versions of dependencies on Travis

greglucas:
Do not merge yet! This needs discussion on whether we want to adopt these policies or not.

## Rationale

The old versions of dependencies are getting difficult to maintain and test against. This adopts the 24-month NEP29 versions discussed in #1518. (OWSlib is pinned to 0.19.1 rather than the 0.17 version as there was an error we running into a couple months ago in the test suite because of that)

## Implications
Not supporting older versions of dependencies anymore.

Easier test suite to maintain.


stefraynaud:
Should we remove the figures created with MPL 1.5 in tests, once this PR is accepted?

greglucas:
Good suggestion, I went through and tried to remove all of the MPL version checks that were less than the minimum supported. Similar could be done with PROJ too I'm guessing.

dopplershift:
I'm 💯 on the concept. Supporting matplotlib 1.5 through 3.3 with image tests is just untenable. Personally, I'd say go ahead and remove the 1.5 test images in this PR as well.

greglucas:
The old images should have all been removed before, just needed a rebase. Let me know if you saw others specifically in there that I should remove too.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -238,8 +209,8 @@ def test_grid_labels():
 
 
 @pytest.mark.skipif(
-    MPL_VERSION < '2.0.0',
-    reason='Impossible to override tight layout algorithm in mpl < 2.0.0')
+    MPL_VERSION < '3.0.0',
```
dopplershift, line 212: I don't follow. The old code used to run on Matplotlib >=2...why do we need to update this rather than flat out remove?

greglucas: Great question! I was wondering the same thing. This is a new test added in the last month or so, and I think our previous CI was only running MPL 1.x and MPL 3.x, never hitting the 2.x builds from what I can see in the Travis setup.

@lukelbd do you know if this test should be possible to run in MPL 2.x? I haven't looked into the tight_layout code on MPLs side at all to know when it was changed and what should be expected.

lukelbd: This change should remain -- I'm not sure why, but turns out the tight layout test will not run successfully until matplotlib >= 3.0. Originally I thought it should work in matplotlib >= 2.0 but [my reasoning was incorrect](https://github.com/SciTools/cartopy/pull/1355#issuecomment-643846634).

After some empirical testing with:

```sh
conda create -c conda-forge/label/testing -c conda-forge -n cartopy-dev-testing cython numpy matplotlib-base=VERSION proj4 pykdtree scipy fiona pillow pytest pytest-xdist filelock pep8 pyshp shapely six requests pyepsg owslib
pip install -e .
echo 'backend: agg' > matplotlibrc
conda activate cartopy-dev-testing
pytest -k 'grid_labels_tight'
```

the last matplotlib-base 2.X version available on conda fails, and the first matplotlib-base 3.X version works.

greglucas: Thanks for the quick response!

I think this is fine for now then, since it is a new feature and we aren't breaking anything that was possible previously. It actually just defaults to the previous 0.18 behavior.

.travis.yml
```diff
@@ -2,14 +2,12 @@ language: python
...
 env:
     matrix:
-        # These ancient versions are linked to an old libgfortran, but that version
-        # isn't pinned in the package metadata
-        - PYTHON_VERSION=3.5
-          PACKAGES="cython=0.28 numpy=1.10.0 matplotlib=1.5.1 nose proj4=4.9.1 scipy=0.16.0 libgfortran=1 owslib=0.19.1"
-          PYTHONHASHSEED=0  # So pytest-xdist works.
+        - NAME="Minimum dependencies."
+          PYTHON_VERSION=3.7
+          PACKAGES="cython=0.28.5 numpy=1.16 matplotlib=2.2.3 proj4=5.2.0 scipy=1.2.0 owslib=0.17"
```
QuLogic, line 7: Might was well sort these at the same time.
```suggestion
          PACKAGES="cython=0.28.5 matplotlib=2.2.3 numpy=1.16 owslib=0.17 proj4=5.2.0 scipy=1.2.0"
```

.travis.yml
```diff
@@ -2,14 +2,12 @@ language: python
...
         - NAME="Latest everything."
-          PYTHON_VERSION=3.6
-          PACKAGES="cython numpy matplotlib-base proj4 pykdtree scipy fiona"
+          PYTHON_VERSION=3.8
+          PACKAGES="cython numpy matplotlib-base pykdtree scipy fiona proj"
```
QuLogic, line 10:
```suggestion
          PACKAGES="cython fiona matplotlib-base numpy proj pykdtree scipy"
```

lib/cartopy/tests/mpl/test_examples.py
```diff
@@ -29,19 +29,14 @@ def new_fn(*args, **kwargs):
 
 
 @pytest.mark.natural_earth
-@ExampleImageTesting(['global_map'],
-                     tolerance=4.5 if MPL_VERSION < '2' else 0.5)
+@ExampleImageTesting(['global_map'], tolerance=0.5)
```
QuLogic, line 32: Remove `tolerance=0.5` as it's the default.

lib/cartopy/tests/mpl/test_examples.py
```diff
@@ -29,19 +29,14 @@ def new_fn(*args, **kwargs):
...
 
 
-if MPL_VERSION < '2':
-    contour_labels_tolerance = 7.5
-elif MPL_VERSION <= '2.0.2':
-    contour_labels_tolerance = 1.24
-elif MPL_VERSION <= '2.1.2':
-    contour_labels_tolerance = 0.63
+if MPL_VERSION < '3':
+    contour_labels_tolerance = 9.8
```
QuLogic, line 39: Confused why this went up.

lib/cartopy/tests/mpl/test_images.py
```diff
@@ -205,8 +192,7 @@ def test_pil_Image():
 @pytest.mark.xfail((5, 0, 0) <= ccrs.PROJ4_VERSION < (5, 1, 0),
                    reason='Proj Orthographic projection is buggy.',
                    strict=True)
-@ImageTesting(['imshow_natural_earth_ortho'],
-              tolerance=4.2 if MPL_VERSION < '2' else 0.5)
+@ImageTesting(['imshow_natural_earth_ortho'], tolerance=0.5)
```
QuLogic, line 195: Drop `tolerance` as it's the default.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -105,8 +98,7 @@ def test_global_pcolor_wrap_no_transform():
 
 
 @pytest.mark.natural_earth
-@ImageTesting(['global_scatter_wrap'],
-              tolerance=12.4 if MPL_VERSION < '2.1.0' else 0.5)
+@ImageTesting(['global_scatter_wrap'], tolerance=0.5)
```
QuLogic, line 101: Drop `tolerance` as it's the default.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -119,8 +111,7 @@ def test_global_scatter_wrap_new_transform():
...
```
QuLogic, line 114: Drop `tolerance` as it's the default.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -596,8 +576,7 @@ def test_quiver_regrid_with_extent():
 
 
 @pytest.mark.natural_earth
-@ImageTesting(['barbs_plate_carree'],
-              tolerance=8 if MPL_VERSION < '2.1.0' else 0.5)
+@ImageTesting(['barbs_plate_carree'], tolerance=0.5)
```
QuLogic, line 579: Drop `tolerance` as it's the default.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -619,8 +598,7 @@ def test_barbs():
 
 
 @pytest.mark.natural_earth
-@ImageTesting(['barbs_regrid'],
-              tolerance=2.9 if MPL_VERSION < '2.1.0' else 0.5)
+@ImageTesting(['barbs_regrid'], tolerance=0.5)
```
QuLogic, line 601: Drop `tolerance` as it's the default.

lib/cartopy/tests/mpl/test_examples.py
greglucas: I have no idea either, hence my statement about the oddities at the beginning... I'll attach the difference image showing the comparison and why I had to increase it.
result-contour_label-failed-diff

When you were testing for 0.18 did you test on the latest MPL 2.x? I'm pretty sure we were missing that from our test suite coverage on CI, so we were testing against 1.x and 3.x only.

greglucas:
Natural Earth download servers are down right now causing the failures.

lib/cartopy/tests/mpl/test_examples.py
QuLogic: I tested every single possible combination of Matplotlib (latest micro only), Proj, and Python, as available in conda.

These are text issues, so FreeType-related. Did you use Matplotlib from the `conda-forge/label/testing` channel to get the one with FreeType 2.6? It looks like they may only have 2.2.2, not 2.2.5 though.

greglucas: On my local environment that I produced that image with I have
`freetype                  2.10.0               he06d7ca_0    conda-forge`
On the CI build it looks like it is pulling from conda-forge (not the /label/testing) and getting a more recent version.
`freetype                  2.10.2               he06d7ca_0    conda-forge`

conda won't let me pin `freetype=2.6` for any of the 2.2.x series of MPL either.

dopplershift: Yeah, we only have 2.2.2 on `conda-forge/label/testing`. You really want to use the versions from there since they're statically linked to freetype 2.6.1, which is what matplotlib uses (currently at least) for their static images for testing. Any reason you made 2.2.3 the minimum on Travis? Could we just make it 2.2.2?

greglucas: I can't install that locally?
```
conda create -n test
conda install --channel conda-forge/label/testing matplotlib=2.2.2
```
gives me freetype 2.10.2 still... Maybe because I'm on OSX though? Explicitly, it completely fails if I add:
`conda install --channel conda-forge/label/testing matplotlib=2.2.2 freetype=2.6`

Just saw 2.2.3 listed on the original Issue and went with it. Maybe we shouldn't even pin to bugfix release versions?

QuLogic: Matplotlib from that channel is statically linked with FreeType; the other package version is not important.

greglucas: Doh! I missed that even though you both said it :(

greglucas: Hm... It doesn't want to pull from that channel.
Here is what CI grabbed:
`matplotlib                2.2.2            py37hb69df0a_2    defaults`

I'm also curious why it is only this one image failing here too and not others with text...

greglucas: Ahh, so I just looked at the testing channel and they only have py36 wheels built for 2.2.2.
https://anaconda.org/conda-forge/matplotlib/files?version=2.2.2&channel=testing
So, I'm not sure what the best option here is. We could add an additional py36 test to the matrix I suppose.

greglucas: Looks like using Py36 solved that problem. I think this is good for a full review now. Leaving the test matrix at Py36 with minimum dependencies and Py38 with latest everything seems reasonable enough for now.

PR #1588, opened 2020-06-19
Globe instances default to WGS84 only if no args were passed

vlas-sokolov:
Addresses #564.

As per title, the PR makes Globe instances default to WGS84 only if no args were passed. Regression test added as well. Looked like low-hanging fruit to me - are there some complications here I'm not aware of?


vlas-sokolov:
Oops, just realised I had to fill in the CLA. Filled and signed.

lib/cartopy/tests/test_crs.py
```diff
@@ -186,6 +186,16 @@ def test_globe(self):
         assert_arr_almost_eq(rugby_pt, (1400915, 1741319), decimal=0)
         assert_arr_almost_eq(footy_pt, (155657, 193479), decimal=0)
 
+    def test_globe_defaults(self):
+        """Regression test for #564"""
+        default_globe = ccrs.Globe()
+        assert default_globe.ellipse == 'WGS84'
+
+        eccentric_globe = ccrs.Globe(semimajor_axis=1000, semiminor_axis=500)
+        assert not eccentric_globe.ellipse  # would need revision with #330
```
QuLogic, line 195: What does this mean? What kind of revision?

vlas-sokolov: If I understood #330 correctly, when #330 gets implemented, this unit test might start misbehaving, because the `.ellipse` attribute might get set with some kind of default or whatever gets passed along from `proj4`. The `assert` statement would get invalidated in that case and would have to be revised, right?

vlas-sokolov: I see! I'll simply remove the comment then.

QuLogic:
It appears there are many tests which expect this default ellipse and will have to add WGS84 explicitly. I guess this could also get a doc note.

vlas-sokolov:
Got it, I'll investigate. Any pointers as to where that doc note should get added?

PR #1592, opened 2020-06-21
Bugfix for masked arrays in imshow

cdholmes:
Geoaxes imshow currently throws an error ("AxisError: axis 2 is out of bounds for array of dimension 0") when trying to plot a masked array with no mask, i.e. img.mask = np.bool_(False). The bug is fixed by testing np.any(img.mask) instead. 

No special handling is required for masked arrays if img.mask is
False, np.bool_(False), or an array of these. imshow previously
only checked for False, so np.bool_(False) was not caught.

Signed-off-by: Christopher Holmes <cdholmes@fsu.edu>


cdholmes:
I signed the CLA after submitting the pull request and getting the "blocked" message.

QuLogic:
Would `np.ma.is_masked` be useful here?

cdholmes:
Yes, using is_masked would make the test even simpler:

`if (np.ma.is_masked(img) and shape[2:3] == (3, )):`

greglucas:
If you rebase and make the suggested change I think the checker will automatically find you and remove the block.

PR #1596, opened 2020-06-23
DOC: add pkg-config instructions for macOS install

tgy:
## Rationale

macOS should be able to install cartopy via `pip install cartopy` smoothly

## Implications

Better instructions will be provided to macOS users when they want to build
cartopy from the source.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -23,6 +23,12 @@ Additional pre-built binaries can be found at a variety of sources, including:
 Building from source
 --------------------
 
+On macOS, make sure you have installed pkg-config and set the `PKG_CONFIG_PATH`
+enviornment variable as follows
+
+    brew install pkg-config
+    export PKG_CONFIG_PATH=/usr/local/bin/pkgconfig
+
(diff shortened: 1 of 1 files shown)


tgy:
I have now signed the CLA.

QuLogic:
Makes sense to add it, but I think maybe it should go in the 'Required dependencies' section?

tgy:
@QuLogic good point. Is that OK now?

INSTALL
```diff
@@ -61,6 +61,16 @@ For macOS, the required dependencies can be installed in the following way::
     # installed, uninstall it by: pip3 uninstall shapely
     pip3 install shapely --no-binary shapely
 
+Still on macOS, make sure you have installed pkg-config and set the
+`PKG_CONFIG_PATH` enviornment variable as follows
```
tgy, line 65:
```suggestion
`PKG_CONFIG_PATH` environment variable as follows
```

QuLogic:
Thanks @tgy!

PR #1607, opened 2020-07-12
Fix limits/threshold of projections with non-default Globe

QuLogic:
## Rationale

Some limits and thresholds are hard-coded, but this causes incorrect projections when the Globe radius is much smaller than the default. Either the map is tiny in very large limits, or lines are not projected correctly.

## Implications

`TransverseMercator` is no longer tiny when using a small Globe radius; other projections do not cut projected lines randomly.


dopplershift:
Code changes look good in general. Looks like this broke the `test_grid_labels` test in `test_gridliner.py`, and angered Travis by trying to generate some HTML with the result image base64-encoded.

QuLogic:
Weird, I thought I saw that failure before making this change and thought it was unrelated, but I guess not.

QuLogic:
The test failure is because the `AzimuthalEquidistant` is now missing a whole bunch of lines. I rebased hoping that your 1000-point PR would have fixed it, but it appears not.
image

lib/cartopy/crs.py
```diff
@@ -841,9 +841,17 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
                 proj4_params += [('approx', None)]
         super().__init__(proj4_params, globe=globe)
 
+        # TODO: Let the globe return the semimajor axis always.
+        a = np.float(self.globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS)
+        b = np.float(self.globe.semiminor_axis or a)
```
greglucas, line 846: Is there a reason for these to be `np.float()` instead of just leaving them as the values that are returned?

lib/cartopy/tests/crs/test_transverse_mercator.py
```diff
@@ -21,16 +22,56 @@ def setup_class(self):
         self.point_b = (0.5, 50.5)
         self.src_crs = ccrs.PlateCarree()
 
+    def check_args(self, approx, proj, other_args):
+        if ccrs.PROJ4_VERSION < (6, 0, 0):
```
greglucas, line 26: Can get rid of this check now since we are dependent on pyproj's PROJ version and looks like we are at 8+
https://pyproj4.github.io/pyproj/3.3.1/

greglucas (commented):
This looks good to me and will help folks when dealing with their moon/astro projections too. Just needs a rebase/update I think.

lib/cartopy/crs.py
QuLogic: I don't see any reason why any more.

lib/cartopy/tests/crs/test_transverse_mercator.py
QuLogic: Yes, removed.

QuLogic:
I made a small tweak to the magic numbers so that calculations all looked consistent.

QuLogic:
I ran all projections with the test from the issue (except a few that don't support a `globe` argument, and Spilhaus), and found we need to do the same with Aitoff and Hammer.

greglucas:
> I ran all projections with the test from the issue (except a few that don't support a `globe` argument, and Spilhaus), and found we need to do the same with Aitoff and Hammer.

Do you want to add these here or is this ready to merge?

greglucas (approved):

QuLogic:
> Do you want to add these here or is this ready to merge?

They're in here as well.

greglucas:
Great! GitHub's review interface didn't show me them obviously, I think some of the classes were collapsed in the diff.

PR #1614, opened 2020-07-23
Upgrading sphinx gallery requirement

greglucas:
Moving the documentation around.
I moved the `/lib/cartopy/examples` directory to `examples/` and created subdirectories for the gallery to organize that way.

Edited a bunch of references and a few missing tag closures etc... to avoid warnings.

## Rationale
I noticed docs were failing on CI, so I went through to clean things up and hopefully start building again.

We are pinning to version 0.2 of sphinx-gallery currently and inserting our own code. This removes all extra source code and put us in line with sphinx-gallery.

## Implications
Hopefully building docs again.


QuLogic:
There was some push back to keep the examples importable #902; I'm not really sure it's worth it though.

greglucas:
I am not sure where the problem actually was... So, my guess is that we actually could put the examples back in the Cartopy library so they are importable. But, the organization with subsections etc... isn't quite as nice then. I was trying to make this more in line with other projects so discoverability and maintenance would be easier and got carried away before I realized the lack of importability used in the tests. I also don't like that the tests import from examples, that just seems little dubious to me because examples should be fluid in some sense, and not used for testing code. I also don't see that running an example through an import is all that much easier than `python path/to/example.py`, I think every example I have ever used I download the python file directly and run it.

I don't have a real strong preference either way, but I don't really want to go through moving all of the links around again if I can avoid it too.

dopplershift:
I'm a fan in general of standardizing things as much as possible. We clearly don't have enough bandwidth right now, so anything to remove cartopy's bespoke infrastructure in favor of things that "handle themselves" is a 💯 from me. In this case, while the intention was to put a lot more into the docs and advertise this ability, it's now 3 years later and still nothing. "..practicality beats purity" and all that.

Having said that, I don't want to move too quickly to give others on the other side of the pond to chime in. At the same time the doc build is busted. Is it possible to fix the doc build (and put that in a separate PR) without doing the sphinx-gallery upgrade?

greglucas:
See the new PR I put up once the artifacts build. I left the examples directory alone over there, but the gallery is no longer organized. That may be OK since there are very few examples and there have been other issues of people wanting to reorganize the docs as well.

I still personally like this PR better, but it is a lot more rip the bandaid off all at once style.

dopplershift:
Thanks for putting in #1616. To be clear, I'm very much in favor of this approach here. I just also want to get CI unstuck so we can get PRs merged in easily.

QuLogic:
I prefer this one as well. If I understood correctly, there's only one example that actually had multiple tags?

greglucas:
That is correct, there was only one example with multiple tags, and I personally think each example should really only be describing a single thing or a single category to keep it simpler.

Looks like no one else had any opinions on either of these PRs. Since everyone that has voiced an opinion so far has supported this one, I'd vote putting this one in and then we can revert it if there is major pushback later. That way we can at least build the docs again.

QuLogic:
I was going to merge, but I guess some conflicts arose.

dopplershift:
I'll second that: this has been around long enough with no commentary that I would merge as well.

greglucas:
Looks like my other test updates conflicted. I just rebased and fixed those.

greglucas:
It looks like PROJ 7.1 is causing an issue with the gridliner_labels_tight image comparison. The difference is the 0-degree label line in the PlateCarree projection is shifted ever so slightly to the right now. The [PROJ changelog](https://github.com/OSGeo/PROJ/releases) doesn't show me anything obvious at first glance and I'm not sure why it would only be appearing in this test and not the others.

dopplershift:
I'm fine if we fix that here or in another PR.

QuLogic:
Let's just merge this.

PR #1616, opened 2020-07-25
DOC: Removing sphinx-gallery version pin

greglucas:
NOTE: This is an alternative to #1614  

This eliminates the examples directory move. But, note that it doesn't have subsection organization anymore in the gallery. Not sure if that is easy to do without making directories explicitly, and also not sure it is worth the time with the other PR already doing that piece.

## Rationale

Alternative proposal to leave examples directory importable

## Implications
Building docs again


greglucas:
Closing because the other option was merged in.

PR #1623, opened 2020-08-02
Updating the shapefile cache test for macosx

greglucas:
This update forces the test to use the Agg backend. When using
the macosx backend the draw() would not get registered and
the cache would never get updated.

## Rationale

On my local testing on a macosx machine I'd get this one failure and ignore it. Now all tests will pass on macosx

## Implications

We could also call switch_backend within the function itself, but I figure if there is already a decorator to handle it, might as well wrap in an empty image comparison.


dopplershift:
Test failures are unrelated to this change.

PR #1624, opened 2020-08-02
TST: Remove lock files from test directory

greglucas:
This cleans up the image test directory by removing the empty '.lock' files that were created.

## Rationale

Each image has an empty '.lock' file after the test suite is run. This is an attempt to clean that up.

## Implications

I am not sure if the `unlink` immediately after the `with` block will enter any race conditions on multi-threaded test runs, but I assume not because it is done immediately? It would be good for someone with more knowledge on the lockfile paradigm to make sure this is OK.

### DIFF
--- a/lib/cartopy/tests/mpl/__init__.py
+++ b/lib/cartopy/tests/mpl/__init__.py
@@ -12,7 +12,7 @@ import glob
 import shutil
 import warnings
 
-import filelock
+import flufl.lock
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import matplotlib.patches as mpatches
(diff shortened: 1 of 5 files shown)


dopplershift:
Why doesn't the `filelock` module handle this for us? It would be most safe to have the `unlink` done on the exit of the context manager.

lib/cartopy/tests/mpl/__init__.py
```diff
@@ -147,6 +147,8 @@ def run_figure_comparisons(self, figures, test_name):
             with filelock.FileLock(result_path + '.lock').acquire():
```
dopplershift, line 147: Try instead changing this to:
```
with filelock.FileLock(result_path + '.lock'):
```
and remove the acquire. Based on this [example](https://github.com/benediktschmitt/py-filelock#examples).

greglucas: This didn't work for me. I think this is dependent on the underlying OS flock command. Apparently they will leave the file around on purpose to keep it thread safe.

I agree it would be nice to do it on exit of the context manager. Additionally, we could drop the filelock dependency and use matplotlib's own private cookbook recipe for `_path_lock`, but then we would be using a private function from them.

QuLogic: We could try `flufl.lock`; it doesn't seem to leave the lock file around.

greglucas: Just tried it out locally and seems to be fine there. Lets see what CI thinks.

PR #1625, opened 2020-08-03
Cleaning up test warnings

greglucas:
This cleans up the warnings from the test suite.

First commit: Deprecation warning from MPL for reusing the same axes. They aren't image tests, so weren't closing the figures. Fixed this by calling close after each non-image comparison test.

Second commit: Catch the expected DownloadWarning with pytest to remove that warning.

## Rationale

Fewer warnings

### DIFF
--- a/lib/cartopy/tests/io/test_downloaders.py
+++ b/lib/cartopy/tests/io/test_downloaders.py
@@ -148,7 +148,8 @@ def test_natural_earth_downloader(tmpdir):
 
     # check that the file gets downloaded the first time path is called
     with CallCounter(dnld_item, 'acquire_resource') as counter:
-        shp_path = dnld_item.path(format_dict)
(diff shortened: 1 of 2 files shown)


greglucas:
This took care of all of the warnings on my local setup (other than for external packages like owslib). Let me know if you saw others I should look at.

I just went and looked at the CI logs and there are other `DownloadWarning`s appearing there. We should probably take a look at the tests and make sure we are only downloading necessary files (i.e. we don't need both 50m and 110m geometries for testing things), or consolidate all of the downloads into a `setup()` call and catch the warnings there.

I'll tackle that in another PR.

QuLogic:
Closing figures should probably get a `cleanup` decorator, but really we should add an `autouse` pytest fixture like Matplotlib does, so we don't need to worry about that.

PR #1628, opened 2020-08-05
Adding an autouse fixture to close plots after each test

greglucas:
Automatically closing figures after each test. Implemented the suggestion from https://github.com/SciTools/cartopy/pull/1625#issuecomment-668380062

## Rationale

Will automatically close figures after each test to not reuse the same figures.

### DIFF
--- a/lib/cartopy/tests/mpl/conftest.py
+++ b/lib/cartopy/tests/mpl/conftest.py
@@ -1,0 +1,21 @@
+# Copyright Cartopy Contributors
+#
+# This file is part of Cartopy and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+import matplotlib.pyplot as plt
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def mpl_test_cleanup(request):
+    """Runs tests in a context manager and closes figures after each test"""
+    try:
+        # Run each test in a context manager so that state does not leak out
+        orig_backend = plt.get_backend()
+        plt.switch_backend('Agg')
+        with plt.rc_context():
+            yield
+    finally:
+        # Closes all open figures and switches backend back to original
+        plt.switch_backend(orig_backend)


QuLogic:
Can we have a `conftest.py` only in the `mpl` directory? We may also want to put a bit of Matplotlib testing reset into the beginning of this function too.

greglucas:
Good suggestion, I just moved everything over to the mpl-specific directory so that it only gets called there and not in every test.

What other testing reset do you think we need? Do you want to run each test in an `rccontext` so that any rcparams don't get propagated as well?

greglucas:
Just realized the actual MPL fixture is not private. I was looking at `_cleanup_cm` this whole time 🤦‍♂️ I just pushed a new version to bring that directly into the new conftest file.

dopplershift:
I'm not sure how I feel about this:
1. Importing from matplotlib's `conftest` feels weird, I wouldn't necessarily consider that public API
2. `mpl_test_settings` is *all* of matplotlib's test settings, beyond using the figure auto-closer.

@QuLogic ?

greglucas:
That is one of the downsides of force-pushing over things with multiple implementations. Sorry about that, Ryan. The previous iteration I had in here was simply:
```python
@pytest.fixture(autouse=True)
def close_plots(request):
    try:
        yield
    finally:
        plt.close()
```

Actually, it looks like using MPLs version doesn't work on Python 3.5 for us either, so maybe the previous iteration is the better route anyways.

greglucas:
I just undid the "simple" version of grabbing MPLs test fixture.

I put the `yield` into a context manager. Of course, on my macosx backend that broke some tests :( The fix for me locally was to force all mpl tests to use the Agg backend by default (The image testing decorator switches to TkAgg when necessary for `show()` calls), is this an OK approach?

lib/cartopy/tests/mpl/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+    """Runs tests in a context manager and closes figures after each test"""
+    try:
+        # Run each test in a context manager so that state does not leak out
+        orig_backend = plt.get_backend()
+        plt.switch_backend('Agg')
+        with plt.rc_context():
+            yield
+    finally:
+        # Closes all open figures and switches backend back to original
+        plt.switch_backend(orig_backend)
```
dopplershift, line 21: I had no idea `switch_backend` closed all the figures, but that makes sense.

lib/cartopy/tests/mpl/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def mpl_test_cleanup(request):
+    """Runs tests in a context manager and closes figures after each test"""
+    try:
+        # Run each test in a context manager so that state does not leak out
+        orig_backend = plt.get_backend()
+        plt.switch_backend('Agg')
```
dopplershift, line 16: Works for me.

lib/cartopy/tests/mpl/conftest.py
greglucas: Me either until I went through MPLs cleanup decorator. Everyones learning something today :)

PR #1629, opened 2020-08-06
Filter warnings that we can expect

greglucas:
This helps to filter out CI logs due to always issuing DownloadWarning and TileMatrixLimits that we know can be emitted in new environments.

I noticed that there is also a DownloadWarning issued when importing the shapereader module. This is due to running the docstring that happens to download the NaturalEarth there. I'm not clear how to filter that warning or not run the docstring while in pytest.

## Rationale

Cleaner logs when running tests.

## Implications

No more warnings about these things.

### DIFF
--- a/lib/cartopy/tests/io/test_ogc_clients.py
+++ b/lib/cartopy/tests/io/test_ogc_clients.py
@@ -129,6 +129,7 @@ class TestWMSRasterSource:
         assert img.shape == (40, 20, 4)
 
 
+@pytest.mark.filterwarnings("ignore:TileMatrixLimits")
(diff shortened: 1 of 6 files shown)


QuLogic:
Needs rebase for doc CI fix?

The doctests are run because we pass `--doctest-modules`; I think this bit is intended.

greglucas:
I did not look at travis for the actual pytest command, thanks for pointing that out! That one doctest is the only import of that shapefile ever taking place, so we may want to change that to use a different one at some point in the future to save extra downloads.

But, this should be rebased and good to go now.

PR #1636, opened 2020-08-20
Changing the interpolation of vectors into the source projection

greglucas:
This PR changes the vector transform to still create an even grid in the target frame, but compute the interpolation in the source frame.

Updated computation:
1. Transforms all source coordinate points into the target projection and makes a regular grid in the target projection.
2. The target grid points are then transformed back to the source projection
3. Interpolation is computed in the source frame.
4. Transform the interpolated source grid vectors into the target projection to return the fully transformed gridded data.

## Rationale
Interpolation should be done in the frame where the data is collected rather than the new projected coordinate system.

## Implications

This is a pretty major behavior change, but somewhat surprisingly only breaks 4 tests. Rather than defaulting to either target frame or source frame, we could add a keyword `interpolation_crs`or something similar to specify where the interpolation should take place as well. That way we could default to the projection space and leave backwards compatibility.


dopplershift (approved):
Looks good to me. I'll leave open for a bit for any additional comments.

greglucas:
No one seems to have had any comments. I think this should still be good to go.

Issue #1510, opened 2020-04-09
Expose the full Table of Content of the web site ?


jypeter:
I got **lost again when trying to navigate the cartopy website**!

This may come from the fact that some pages are listed on the home page, in the _Getting Started_ section, but **some other useful pages are not listed** there, and you can only find them by searching the web site, or by visiting one of the pages listed in _Getting Started_ and then clicking on the **Next/Previous Topic** in the right part of the page. **This is not intuitive at all!** Would it be possible to have a full Table of Content of the site available somewhere?

I confess I don't know what would be the (best) solution. Maybe listing [A], [B] and [C] (see below) in _Getting Started_? Maybe explicitly mentioning [A], [B] and [C] in the _Using cartopy with matplotlib_ would help as well

You'll see below the pages listed in _Getting Started_ (not indented). The indented lines are the pages that are not listed
```
         home page
         [more topics]
         Cartopy
Coordinate reference systems in Cartopy
Cartopy projection list
Using cartopy with matplotlib
	  [A] More advanced mapping with cartopy and matplotlib
	  [B] Cartopy Matplotlib integration reference document
	  [C] Cartopy map gridlines and tick labels
The cartopy Feature interface
Understanding the transform and projection keywords
Using the cartopy shapereader
Cartopy developer interfaces
	  The cartopy module
```

QuLogic:
There's an outline page, and an index page, and it seems like none of these are shown there either. I don't know why or how that happens though, but I also just got lost trying to find a page.

jypeter:
Well, there are _outline_ and _index_ links at the top and bottom of all the pages, but:
- No global Table of Content (or I have not found it yet)
- Some links are missing on the outline page (#1322)
- The _outline_ word is not helpful enough (in my opinion), as it does not really describe what is behind the link (#1319)

stefraynaud:
Yep, the site is far from being clear.

Possible **content**?

* Installation/setup
* User guide
  * Coordinate reference systems in Cartopy
  * Cartopy projection list
  * etc..
* Reference guide
  * API
  * What's new
  * Appendices
* Getting involed

A good use of the `.. toctree::` makes the toc always accessible.
A use of the toctree `:caption:` option may also help to clarify.
The **theme** may also be updated. A RTD-like theme?

Keeping most of the actual content, it should be not that difficult to reorganised the doc.

greglucas:
[...]

PR #1652, opened 2020-09-09
Doc switch to pydata theme and reorganization

greglucas:
## Rationale

There are quite a few issues in the tracker open about documentation and organization. https://github.com/SciTools/cartopy/search?q=documentation&type=Issues
One such issue: https://github.com/SciTools/cartopy/issues/1510

This update keeps most of the content from the previous site, but reorganizes it with the pydata theme and uses more general :toctree: calls for an API reference rather than inline descriptions.

## Implications

Currently, there is not a version_switcher in the pydata theme. I did rework the nav-item's and could insert it to look like this:

But, that would involve overwriting the main pydata theme navbar and potentially getting out of sync with it.


QuLogic:
This looks fairly nice, though I have really looked at it in depth.

greglucas:
ping: @jypeter and @stefraynaud
You both had suggestions/comments on the linked documentation issue. Feel free to review this and suggest changes.

stefraynaud:
@greglucas This is a huge improvement the doc!

Minor suggestions :
- I'm not sure the "Getting involed" block is at the right place. Maybe an alternative is to merge the home page with the current Readme file, and include it in the index page with `.. include::` directive.
- Please add the module names in the title of the API sections.

Nice work.

@jypeter you can have a look to the doc by clicking on the "Show all checks" link of the CI status at the Bottom of this page, then on the "Details" link in front of "ci/circleci: docs-python3 artifact".

greglucas:
Excellent, thank you for the feedback!

1. I agree, the entire landing page feels a bit awkward at the moment and "Getting Involved" really isn't important to a first-time user just arriving at the docs. The current README has some circular language pointing to the docs and extra copyright/licensing things that I think should get pushed back a page. Getting Started may be a better landing page, but more verbose. I'll give this some thought and try to reorg it a bit (so far I just tried to re-layout the current content, not rewrite content).

2. Where do you want to see module names? In the API reference page itself, or the subsections from there? Something like: `Matplotlib interface (cartopy.mpl)`? I was trying to loosely follow [numpy](https://numpy.org/devdocs/reference/index.html) and [pandas](https://pandas.pydata.org/docs/reference/index.html). It looks like numpy includes some, pandas doesn't. We also have quite a few nested layers of modules it seems like, so I'm not real clear on where to draw the line for each one. (Also note, currently the documentation for each *class* page does include the full [module path](https://2358-5282596-gh.circle-artifacts.com/0/docs/build/html/reference/generated/cartopy.mpl.ticker.LongitudeFormatter.html#cartopy.mpl.ticker.LongitudeFormatter).

stefraynaud:
OK.
As for the module names, the case of numpy is special since almost everything is available under `numpy.`.
In the case of pandas, the module name is not shown in the module title, however the module/section content show full declaration paths so it make it clear in which submodule we are.
For instance, it is not clear on the overview page that we are in the in the `cartopy.mpl` module whereas this becomes clear in the item page: 
* overview : https://2358-5282596-gh.circle-artifacts.com/0/docs/build/html/reference/matplotlib.html#gridlines-and-ticks 
* item : https://2358-5282596-gh.circle-artifacts.com/0/docs/build/html/reference/generated/cartopy.mpl.geoaxes.GeoAxes.html#cartopy.mpl.geoaxes.GeoAxes

greglucas:
Thanks @stefraynaud, that makes sense. I moved some of those to the section headers (`cartopy.mpl`), but others I just more explicitly wrote out in the subsection instead (`geoaxes.Geoaxes`). I'm not sure how we should deal with the doubly nested structures like `cartopy.io.shapereader`, would you want to see the full `io.shapereader.ShapeReader` path, or is it obvious enough as is? Let me know what you think of this new version.

jypeter:
Thanks @greglucas and @stefraynaud !

stefraynaud:
Its seems that this docstring code is not well processed by sphinx:

````
        Keyword Parameters
        ------------------
````

greglucas:
That should just be under the `Parameters` section. I've moved that now.

I think there are a lot of other fixes like updating types/intersphinx references better, but I'd rather hold off on those for separate PRs. This is a major overhaul already and I'd like to keep it somewhat self-contained to only updating the documentation format/structure for now.

greglucas:
Plugging this PR for a review as well, I think it is a major improvement to the doc situation.

Rendered version: https://2692-5282596-gh.circle-artifacts.com/0/docs/build/html/index.html

QuLogic:
It looks like the What's new for 0.19 is missing from the index?

QuLogic:
The extra macOS instructions in the Required Dependencies section appear to be not indented like they were before, but really, they should be code styled.

greglucas:
Thanks for pointing those out! I updated the MacOS extra instructions to follow the same style as above and added in the v0.19 release notes.

QuLogic (approved):

greglucas:
I'm going to merge this so I don't have to keep rebasing. We can start making smaller doc PRs on top of this now. Thanks for the feedback and improvements, everyone!

Issue #1578, opened 2020-06-05
10m land feature read incorrectly by basic reader


massoncl:
### Description
Depending on the extent of a map, adding the land features will color the entire plot (including the ocean).
Related to #803, #579, #492.
This issue seems to have appeared on the 0.18 version, although I haven't tested it. 
The issue is present when plotting with the NearsidePerspective projection, but not PlateCarree (I haven't tried others). 

#### Code to reproduce

```
import cartopy
import matplotlib.pyplot as plt

nearside = cartopy.crs.NearsidePerspective(central_longitude=0, central_latitude=0)
geo = cartopy.crs.PlateCarree()
ax = plt.axes(projection=nearside)
extent = 7
ax.set_extent([-extent, extent, -extent, extent], crs=geo)
ax.add_feature(cartopy.feature.LAND)
ax.coastlines()
```
produces 
![image](https://user-images.githubusercontent.com/44751719/83857468-53329180-a71b-11ea-8b04-a68b2f67b6a7.png)

setting the extent to 8 in the above code gives the expected result :
![image](https://user-images.githubusercontent.com/44751719/83857533-69d8e880-a71b-11ea-8cf6-91a4f2ecc0c3.png)

<details>
 <summary>Full environment definition</summary>

### Operating system
Windows 10

### Cartopy version
0.18.0

</details>

greglucas:
Which versions of PROJ and GEOS do you have? I am unable to reproduce this.

clem-masson:
I'm using PROJ 7.0.0 and GEOS 3.8.1. I installed everything in a new miniconda environment, with the following commands : 
```
conda create --name test_cartopy python=3.8.3
conda activate test_cartopy
conda install cartopy
```
Here's the output of `conda list`, just in case : 
<details>

```
brotlipy                  0.7.0           py38he774522_1000
bzip2                     1.0.8                he774522_0
ca-certificates           2020.4.5.2           hecda079_0    conda-forge
cartopy                   0.18.0           py38hde6099d_1    conda-forge
certifi                   2020.4.5.2       py38h32f6830_0    conda-forge
cffi                      1.14.0           py38h7a1dbc1_0
cfitsio                   3.470                he774522_5
chardet                   3.0.4                 py38_1003
cryptography              2.9.2            py38h7a1dbc1_0
curl                      7.69.1               h2a8f88b_0
cycler                    0.10.0                   py38_0
expat                     2.2.9                he025d50_2    conda-forge
[...]

PR #1653, opened 2020-09-09
Update dependency to pyshp v2.1.1 to correctly display shapefiles with multi-polygon holes

karimbahgat:
## Rationale

The new v0.18 introduced problems with rendering some shapefiles (#1578). This was due to how the pyshp dependency converted the internal shapefile shape structure to GeoJSON for multipolygons with holes, sometimes assigning the holes to the wrong polygon. A new version of PyShp has since fixed this issue, including several other shapefile bug fixes, and this PR updates the dependency to use the newest PyShp version. The PR also cleans up some of the code that uses pyshp. 

## Implications

This bugfix PR will make it a requirement for all cartopy users to upgrade their pyshp version to >= v2.1.1, in order to ensure that all Cartopy users get the correct rendering (#1578). I'm not sure to what degree requiring such an upgrade is a nuisance, or if the added benefit is sufficient for such a change. 

As part of cleaning up the code that uses pyshp, I also removed the io.shapereader.Record's `fields` argument and private `._fields` attribute, as this did not seem to be used anywhere else in the codebase, and shouldn't have any benefit, since the record's fieldnames are already stored as the keys to the `attributes` dict. Anything that relies on this seemingly unused `fields` arg/attr would break, but I see no reason why anyone would need it and the attribute is private anyway.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -53,7 +53,7 @@ class Record:
     their associated geometry.
 
     """
-    def __init__(self, shape, attributes):
+    def __init__(self, shape, attributes, fields):
         self._shape = shape
 
         self._bounds = None
@@ -67,6 +67,8 @@ class Record:
 
         self.attributes = attributes
         """A dictionary mapping attribute names to attribute values."""
+        
+        self._fields = fields
 
     def __repr__(self):
         return '<Record: {!r}, {!r}, <fields>>'.format(
@@ -160,9 +162,12 @@ class BasicReader:
         Return an iterator of :class:`~Record` instances.
 
         """
+        # Ignore the "DeletionFlag" field which always comes first
+        fields = self._reader.fields[1:]
+        field_names = [field[0] for field in fields]
         for shape_record in self._reader.iterShapeRecords():
             attributes = shape_record.record.as_dict()
-            yield Record(shape_record.shape, attributes)
+            yield Record(shape_record.shape, attributes, field_names)
 
 
 class FionaReader:
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -160,7 +160,7 @@ class BasicReader:
         Return an iterator of :class:`~Record` instances.
 
         """
-        for shape_record in self._reader.iterShapeRecords()):
+        for shape_record in self._reader.iterShapeRecords():
             attributes = shape_record.record.as_dict()
             yield Record(shape_record.shape, attributes)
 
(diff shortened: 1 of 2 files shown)


greglucas:
We are currently following [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) with a 24-month dependency policy. See issue https://github.com/SciTools/cartopy/issues/1518 for discussion around that.

Can the hard dependency be removed and just allow users to benefit if they upgrade pyshp themselves? That being said, I think all of the refactors are good, and we probably need a deprecation for the fields keyword in case someone downstream is using that even though we aren't.

dopplershift:
I like the clean-ups as well. Are they compatible with pyshp < 2? I think that's another consideration for what version of `pyshp` we support here.

karimbahgat:
Good point @greglucas, does requiring a newer version constitute a dependency break? If cartopy was to require it, would that mean inserting a deprecation warning and then adding the version bump requirement after 24 months? Maybe it would be best to just leave this as an optional benefit for those that have the newer version. I think it depends how serious you think issue #1578 is, if it's an acceptable Cartopy bug for those with the older pyshp versions? As I point out in [pyshp issue 202](https://github.com/GeospatialPython/pyshp/issues/202), the bug is only potentially relevant for multipolygons with holes, and even then I believe most of the time the polygon rings are ordered in the way assumed by the code, not resulting in the reported error, though I cannot be certain. 

@dopplershift, some of the code cleanup would not in fact be compatible with pyshp < 2, specifically len(shapefile) and .as_dict(), so I would have to redo those differently if you don't want to require pyshp 2+. Then I should probably also make a new PR, that only focuses on the code refactoring.

karimbahgat:
Also, iterShapes and iterShapeRecords apparently were only available since pyshp 1.2.2, which would break in some cases since cartopy requires only pyshp>=1.1.4. So most of the code refactoring depends what is done regarding version requirement. Ultimately I would say the code refactoring is less important, more important is whether or not to upgrade pyshp. Not sure when v1.1.4 is from, but v1.1.7 is from back in 2013. At least at some point an upgrade should be considered, there are plenty of bug fixes as well as read efficiency improvements.

greglucas:
Looking at the analysis of https://github.com/SciTools/cartopy/issues/1518 again, the 24-month dependency rule would put pyshp at 2.0.0. So, my vote would be to update the requirements file to require `pyshp >= 2` and take advantage of all your refactors here. In fact, all of the minimum requirements in the requirements file are not even currently tested, so I have no idea if numpy 1.10 would even work anymore honestly.

karimbahgat:
Ok, just to get some closure on this, I think the PR should be ready for merging now, if everyone agrees on the version issue. I changed so the dependency only require pyshp 2 or bigger, and I undid removing the fields arg, as that should probably be another PR and needs a deprecation warning.

dopplershift (approved):

dopplershift:
Looks good. Thanks @karimbahgat !

Issue #1654, opened 2020-09-10
set_clim() doesn't update color limits for cells crossing the antimeridian


LiamBindle:
### Description

@sdeastham and I have been trying SciTools/cartopy#1622 which fixes streaking of data that crosses the antimeridian. Thanks so much for this, it works great and solves a longstanding problem of ours!

We noticed that `set_clim()` doesn't appear to updated color limits for cells that cross the antimeridian. Do you know why this might be? Below are some snippets to reproduce.

#### Code to reproduce
Data files: [data.zip](https://github.com/SciTools/cartopy/files/5204366/data.zip)
- data/grid.npz contains `x` and `y` arrays which are corners of the cubed-sphere grid (6 curvlinear meshes that cover the globe, we call each of these sub-meshes a face). These arrays are shape `(6, 25, 25)`.
- data/data.npy is surface level ozone from GEOS-Chem on a C24 cubed-sphere grid.

---

You can reproduce the issue like so
```python
import cartopy
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import pyproj
import xarray as xr
import numpy as np

print(f"caropty version: {cartopy._version.version}")

data = np.load('data/data.npy')
grid = np.load('data/grid.npz')

vmin=np.quantile(data, 0.01)
vmax=np.quantile(data, 0.99)

plt.figure()
ax = plt.axes(projection=ccrs.PlateCarree())
ax.coastlines()
ax.set_global()

for nf in [0, 1, 2, 3, 4, 5]:
    x = grid['x'][nf,...] % 360
    y = grid['y'][nf,...]
    d = data[nf, ...]
    im = plt.pcolormesh(x, y, d, transform=ccrs.PlateCarree())
    im.set_clim(vmin, vmax)

plt.show()
```
![image](https://user-images.githubusercontent.com/12772306/92796547-e7457200-f376-11ea-81fd-ffde1c7e6ff6.png)
Notice the dark cells (affecting cells crossing the AM) in the figure above. 

---

Instead, if we use the `norm` kwarg to `pcolormesh()` it behaves as expected.

```python
import cartopy
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import pyproj
import xarray as xr
import numpy as np

print(f"caropty version: {cartopy._version.version}")

data = np.load('data/data.npy')
grid = np.load('data/grid.npz')

vmin=np.quantile(data, 0.01)
vmax=np.quantile(data, 0.99)
[...]

Issue #1371, opened 2019-09-21
Document test procedure


greglucas:
I find myself forgetting exactly what commands to run for the pytest test suite here. I think some kind of testing section in the readme just giving a few commands on how to test would be very useful.

I was running
`pytest --pyargs cartopy.tests`
or
`pytest --pyargs cartopy.tests.specific_test_file.py`
which I'm not even entirely sure is the best/right way to run the tests.

For example from matplotlib:
https://github.com/matplotlib/matplotlib#test
Maybe even a developer's guide page if you want to get really detailed?
https://matplotlib.org/devel/coding_guide.html#testing

dopplershift:
I usually end up looking at `.travis.yml`, which has: `pytest -n 4 --pyargs cartopy`. Actually, last time I think `pytest lib/cartopy` worked fine for me.

But yes, it would be good to document.

QuLogic:
This depends on how you've installed cartopy. As a full install, using `--pyargs` and a module path is the way to go. If you've done an editable install, then just running `pytest` by itself, or specifying a file in the repo works (because the 'install' is those same files.)

I have a tox config that could help codify things, but it's got some personal config that I don't know how to drop yet, plus it's got 312 environments (so far) that'd be a bit overwhelming for people.

PR #1656, opened 2020-09-11
Combine wrapped collection into GeoMesh data when get_array is called

sdeastham:
When calling get_array with a QuadMesh object (from pcolormesh), the returned array only included the unmasked data. This change fills in the masked data when present. This follows from the discussion in https://github.com/SciTools/cartopy/issues/1654.

## Rationale

The `pcolormesh` `set_array` method now correctly handles the wrapped, masked collections which can be generated when polygons cross the antimeridian. However, the `get_array` method still only returns the data for the standard cells (those which did not cross the antimeridian). This caused unexpected behavior when running, for example:

```
im = ax.pcolormesh(...)
im.set_array(im.get_array()*1.0)
```

This would result in the same plot being shown, but now with the cells at 180E missing. This fix modifies the `get_array` call to compensate.

## Implications

Calling `get_array` now returns the complete array, with the masked elements filled in.

### DIFF
--- a/lib/cartopy/mpl/geocollection.py
+++ b/lib/cartopy/mpl/geocollection.py
@@ -16,6 +16,14 @@ class GeoQuadMesh(QuadMesh):
     # come from GeoAxes.pcolormesh. These methods morph a QuadMesh by
     # fiddling with instance.__class__.
 
+    def get_array(self):
+        # Retrieve the array - use copy to avoid any chance of overwrite
+        A = super(QuadMesh, self).get_array().copy()
+        # If the input array has a mask, retrieve the associated data
+        if hasattr(self, '_wrapped_mask'):
+            A[self._wrapped_mask] = self._wrapped_collection_fix.get_array()
+        return A
+
     def set_array(self, A):
         # raise right away if A is 2-dimensional.
         if A.ndim > 1:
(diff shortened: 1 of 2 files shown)


greglucas:
Looks like a good way to do it to me, @sdeastham!

Can you add a test for this too? As simple as just putting `mesh.get_array() == Z_input` for a wrapped and not-wrapped case.

Also, it looks like you've got one failing test currently. That is actually one that I just put in to make sure the cells were being wrapped properly. A good fix for that test would be to assert that the `_wrapped_collection` attribute is present in that case (which implies it was wrapped).

greglucas:
ping @sdeastham. I think all that is needed is updated tests here.

sdeastham:
@greglucas Got it. Sorry about the delay - I've never used pytest before and am having terrible trouble getting it to do anything except throw `ImportError`s...

greglucas:
No worries at all. I think that error can be fixed by adding the pyargs argument to pytest: `pytest --pyargs cartopy`. See #1371 for a few different ways to run the tests. It depends on how you installed your local version as well, so if you installed an editable version with `pip install -e .` or just a standard installation `pip install .`

sdeastham:
@greglucas Thanks again for your patience! I've attempted to build some tests which test that `get_array` actually returns the expected data (albeit 1-D), as well as modifying the `diagonal_wrap` test to prevent the failure you mentioned. I wasn't 100% certain on what you intended with that one, so please do let me know if you'd like anything changed (and indeed if you see changes needed with anything else in the new tests).

greglucas (approved):
@sdeastham, glad you stuck with it and got all the tests working. This looks good to me! Thanks for contributing and hope to see you back again!

PR #1662, opened 2020-09-18
Add option to remove 'E', 'W', 'N', 'S' from axis labels. Fixes #1635

### LINKED ISSUE #1635: Feature request: option to remove 'E', 'W', 'N', 'S' from axis labels
### Description

I would like to be able to format my axis labels so that there is no N/S/E/W after the label. 

Is this possible? It seems like a simple option could be added to the Longitude/Latitude formatters. 

In Basemap there was an option for labelstyle="+/-" that could do exactly this. 
<!--

gcaria:
## Rationale

This change fixes #1635 (Add option to remove 'E', 'W', 'N', 'S' from axis labels).

## Implications
 
Users will be able to to remove 'E', 'W', 'N', 'S' from axis labels.

## Checklist

 * I have signed the Contributor Licence Agreement (CLA).

 * The example I'd add is based on https://scitools.org.uk/cartopy/docs/v0.15/examples/tick_labels.html
I wonder if adding a subplot to this example would be a good idea.

 * A test should be added to lib/cartopy/tests/mpl/test_ticker.py. Should I add it to this pull request ?

### DIFF
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -275,6 +275,7 @@ class LatitudeFormatter(_PlateCarreeFormatter):
 
         """
         super().__init__(
+            direction_label=direction_label,
             degree_symbol=degree_symbol,
             number_format=number_format,
             transform_precision=transform_precision,
@@ -285,15 +286,10 @@ class LatitudeFormatter(_PlateCarreeFormatter):
             auto_hide=auto_hide,
         )
 
-        self._direction_labels = direction_label
-
     def _apply_transform(self, value, target_proj, source_crs):
         return target_proj.transform_point(0, value, source_crs)[1]
 
     def _hemisphere(self, value, value_source_crs):
-
-        if not self._direction_labels:
-            return ''
 
         if value > 0:
             hemisphere = 'N'
@@ -403,6 +399,7 @@ class LongitudeFormatter(_PlateCarreeFormatter):
             labels = [lon_formatter(value) for value in ticks]
         """
         super().__init__(
+            direction_label=direction_label,
             degree_symbol=degree_symbol,
             number_format=number_format,
             transform_precision=transform_precision,
@@ -412,7 +409,6 @@ class LongitudeFormatter(_PlateCarreeFormatter):
             seconds_number_format=seconds_number_format,
             auto_hide=auto_hide,
         )
-        self._direction_labels = direction_label
         self._zero_direction_labels = zero_direction_label
         self._dateline_direction_labels = dateline_direction_label
 
@@ -446,9 +442,6 @@ class LongitudeFormatter(_PlateCarreeFormatter):
         return _PlateCarreeFormatter._format_degrees(self, self._fix_lons(deg))
 
     def _hemisphere(self, value, value_source_crs):
-
-        if not self._direction_labels:
-            return ''
 
         value = self._fix_lons(value)
         # Perform basic hemisphere detection.
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -101,7 +101,7 @@ class _PlateCarreeFormatter(Formatter):
             label = self._format_minutes(mn) + label
 
         if not self._auto_hide_degrees or not label:
-            label = self._format_degrees(deg) + hemisphere + label
+            label = sign + self._format_degrees(deg) + hemisphere + label
 
         return label
 
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -21,8 +21,8 @@ class _PlateCarreeFormatter(Formatter):
 
     _target_projection = ccrs.PlateCarree()
 
-    def __init__(self, degree_symbol='\u00B0', number_format='g',
-                 transform_precision=1e-8, dms=False,
+    def __init__(self, direction_label=True, degree_symbol='\u00B0',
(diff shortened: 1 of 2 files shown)


lib/cartopy/mpl/ticker.py
```diff
@@ -274,10 +281,16 @@ def __init__(self, degree_symbol='\u00B0', number_format='g',
...
         )
 
+        self._direction_labels = direction_label
+
     def _apply_transform(self, value, target_proj, source_crs):
         return target_proj.transform_point(0, value, source_crs)[1]
 
     def _hemisphere(self, value, value_source_crs):
+
+        if not self._direction_labels:
```
greglucas, line 291: What about moving this up to the formatting base class? Only call `self._hemisphere()` if necessary

Up above, something like this.
`hemisphere = self._hemisphere() if self._direction_labels else ''`

greglucas (commented):
This looks correct. Could you add the test into this PR as well? No need for an additional image test, just check against the tick format strings returned in the several cases you brought up in the Issue comments.

If you want to add an example too that would be great!

lib/cartopy/mpl/ticker.py
gcaria: This would mean adding the `_direction_labels` argument to `_PlateCarreeFormatter` and removing it from the other classes where I've just added it, correct ?

greglucas: I think it would just mean adding it to the _PlateCarreeFormatter as an additional option there too, you'd still want the explicit keyword arguments in this function I think.

lib/cartopy/mpl/ticker.py
```diff
@@ -422,6 +446,10 @@ def _format_degrees(self, deg):
         return _PlateCarreeFormatter._format_degrees(self, self._fix_lons(deg))
 
     def _hemisphere(self, value, value_source_crs):
+
+        if not self._direction_labels:
```
greglucas, line 450: You shouldn't need these anymore since it is taken care of in the PlateCarree class?

lib/cartopy/tests/mpl/test_ticker.py
```diff
@@ -46,6 +46,17 @@ def test_LatitudeFormatter():
     assert result == expected
 
 
+def test_LatitudeFormatter_direction_label():
+    formatter = LatitudeFormatter(direction_label=False)
+    p = ccrs.PlateCarree()
+    formatter.axis = Mock(axes=Mock(GeoAxes, projection=p))
+    test_ticks = [-90, -60, -30, 0, 30, 60, 90]
+    result = [formatter(tick) for tick in test_ticks]
+    expected = ['90\u00B0', '60\u00B0', '30\u00B0', '0\u00B0',
```
greglucas, line 55: I would expect this to now include the negative signs instead of the W/E labels?
```suggestion
    expected = ['-90\u00B0', '-60\u00B0', '-30\u00B0', '0\u00B0',
```

greglucas (changes requested):
I think you'll need to add some logic in to add negative signs into the formatter rather than the W/S hemisphere labels now.

gcaria:
Thanks for your feedback @greglucas  !

As you suggested now I have to add the extra logic to make sure negative signs are added when `direction_label` is set to `False` (as expected now the tests fail)

greglucas:
There are at least two potential paths you can take to get the negatives showing up and passing the tests.

1. Change the format string directly to start with a negative if the hemisphere would have been W/S. (Something around line 90 probably)
2. Currently we are passing abs(value) into the formatters, you could try and drop the absolute value and change the logic around that. There may be consequences with the `_dms` formatter, but I don't know for sure.

gcaria:
I'm having a bit of a beginner's issue here, sorry for the digression from the issue.

What's the procedure to run tests on a package with local edits (like it's the case for my fork) ?
Should I run `python setup.py` and then run `pytest` ?

For some reason my changes don't show up in the tests (even if I make some silly edit on purpose),
so I'm wondering whether I'm just not following the right procedure or there's something wrong with my virtual environments.

greglucas:
Cartopy is not a trivial package to build/test locally, so you're beyond the beginner questions already :) I even opened up an issue myself because I kept getting confused too! See #1371 for some other advice on that as well.

When I am doing local work I'll usually build an editable install with `pip install -e .` and then run pytest on whatever file I want to test.

gcaria:
Sorry for the long delay, ironically I was stuck with an old laptop and gave up installing cartopy there.

Now the original test that deals with direction_label=False passes, and I've also added a test for the case where dms=True is combined with direction_label=False which passes.

Since now other tests of other directories do not pass, I wonder whether I've broken something or these other tests need to be updated.

greglucas (approved):
Looks good to me now with the negative sign updates! There are a lot of linter fixes and small commits mixed in, can you squash and rebase those to just one or two logical commits?

gcaria:
Ok, I've squashed the fix-linting commits ! I'm not sure if you agree the test for `dms=True` + `direction_label=False` is actually needed.

Out of curiosity, there were tests failing last time I pushed (example tests I think), how can they have magically disappeared ?

greglucas:
I think it is good to leave the `dms=True` test in here. That will make sure that the combination of the added keyword argument and `dms` behaves as expected.

When I went to review your PR before and saw the failures they were all internet download errors on the CI systems, so I just restarted the job for you and it passed. Looks like all of the checks passed after your rebase again, so all is good!

greglucas:
Thank you, @gcaria!

Issue #1642, opened 2020-08-25
Add capability for non-inline grid labels to appear inside map


karlwx:
I want to be able to adjust the position of the grid tick labels to be inside the map boundary. Setting the gridliner xpadding works to move the labels away from the map, but setting the xpadding to a negative number results in hidden labels. I found that that non-inline labels that overlap the map boundary are hidden by default, so I experimented with setting the visibility of all labels to true. However, that resulted in the labels appearing at a 45-degree angle.

I'm essentially trying to do something [like this](https://stackoverflow.com/questions/47873389/matplotlib-move-tick-labels-inside-plot-area). Note: I am not looking to use the inline labels option, since (at least to my knowledge) there isn't a way to fix their positions to a particular x or y offset from the map boundaries. Changing the xpadding doesn't have any effect with the inline labels. Maybe this should be an option instead?

Example code:
```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

plt.figure(figsize=(12,9))
ax = plt.axes(projection=ccrs.Miller())
ax.set_extent([-130,-65,20,55], crs=ccrs.PlateCarree())
ax.coastlines(resolution='50m')
gl = ax.gridlines(draw_labels=True, color='gray', linestyle='--')
gl.top_labels, gl.right_labels = False, False
gl.xlabel_style, gl.ylabel_style = {'weight': 'bold'}, {'weight': 'bold'}
#gl.xpadding = -10     # Ideally, this would move labels inside the map, but results in hidden labels
```
No xpadding:
![Screenshot from 2020-08-25 10-51-30](https://user-images.githubusercontent.com/45180714/91191431-c74d5780-e6c2-11ea-84a2-4f45720b50a3.png)

xpadding set to -10 and visibility hard-coded to True results in rotated labels:
![Screenshot from 2020-08-25 10-50-55](https://user-images.githubusercontent.com/45180714/91191434-c7e5ee00-e6c2-11ea-9842-3080a4c06aa5.png)

greglucas:
A similar request was asked on gitter, so I'll add my comment here as well.
I think you need to set the keyword `rotate_labels` to False in your second example.

karlwx:
I went back and did some experimenting again. For whatever reason, when I tried the same code again, now the labels are not rotated and I get the desired result:
![inmap_labels](https://user-images.githubusercontent.com/45180714/93896323-c8bc7080-fcbe-11ea-9d17-71facd29af36.png)
[...]

PR #1666, opened 2020-10-15
Feature labels on top of map

emsterr:
## Introduction
This pull request is based on the issue #1642: Add capability for non-inline grid labels to appear inside map

## Rationale

As a satellite image provider I need to minimize the size of images and meet the customers requirements on how products shall be presented.

## Implications

With this change all non-inline labels will be shown on top of the map and not outside. This occurs if x-, y- or both padding are negative. All labels that is shown outside the map is removed.

## Input
```python
top = 49.3457868  # north lat
left = -124.7844079  # west long
right = -66.9513812  # east long
bottom = 24.7433195  # south lat
plt.figure(figsize=(35, 35))
for i, proj in enumerate(TEST_PROJS, 1):
    if isinstance(proj, tuple):
       proj, kwargs = proj
    else:
       kwargs = {}
        ax = plt.subplot(7, 4, i, projection=proj(**kwargs))
        try:
            ax.set_extent([left, right, bottom, top], crs=ccrs.PlateCarree())
        except Exception:
            pass
        ax.set_title(proj, y=1.075)
        if ccrs.PROJ4_VERSION[:2] == (5, 0) and proj in (
            ccrs.Orthographic,
            ccrs.AlbersEqualArea,
            ccrs.Geostationary,
            ccrs.NearsidePerspective,
        ):
            # Above projections are broken, so skip labels.
            # Add gridlines anyway to minimize image differences.
            ax.gridlines()
        else:
            gl = ax.gridlines(draw_labels=True, clip_on=True)
            gl.xpadding = -5
            gl.ypadding = -5
        ax.coastlines(resolution="110m")
plt.subplots_adjust(wspace=0.35, hspace=0.35)
```

Code is the same as in the unit test.

## Output
gridliner_labels_intersect_map

## Note
* I cant seem to figure out why a couple of the numbers in Mollweide, Sinusoidal and InterruptedGoodeHomolosine do intersect with the map bbox.
* Unit test provided
* All test passes


stefraynaud:
In think a special keyword like ìnside` must be added to `Gridliner.__init__` to activate this behavior of labels, and consider `self.xpadding` and `self.ypadding` as always positive.
I added this feature to my local version of PR #1537, which is finally a major revision of the labelling system that solves numerous issues.

emsterr:
Ok. So you have implemented this functionality in your branch ```stefraynaud:bug_label_visibility```? I also need the feature where labels outside of map is removed completely. Since i use ```bbox_inches="tight", pad_inches=0,``` to remove borders  when saving images output.

What is your suggestion on how we proceed?

emsterr:
@stefraynaud, I will just use my local solution until you have rolled out your PRs. You have a timeline for the next release?

stefraynaud:
@emsterr Except a few tests I must add, I mostly dépends on the CI success and on the review process.

stefraynaud:
@emsterr You should push forward this PR to make it merged as soon as possible.

emsterr:
@stefraynaud sorry I'm a bit confused, what do you mean 'push forward'? isn't it ready to be merged?

lib/cartopy/mpl/gridliner.py
```diff
@@ -799,17 +811,30 @@ def remove_path_dupes(path):
...
                     if ((lonlat == 'lon' and self.x_inline) or
                             (lonlat == 'lat' and self.y_inline)):
                         # TODO: When Matplotlib clip path works on text, this
                         # clipping can be left to it.
                         if outline_path.contains_point(center):
                             visible = True
-                    # Non-inline must not run through the outline.
-                    elif not outline_path.intersects_path(this_path):
+                    elif not outline_path.intersects_path(this_path,
+                                                          filled=is_filled):
```
greglucas, line 830: What about adding the boolean tests directly into the argument here instead of creating another variable?

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -365,3 +367,40 @@ def test_gridliner_line_limits():
     for path in paths:
         assert (np.min(path.vertices, axis=0) >= (xlim[0], ylim[0])).all()
         assert (np.max(path.vertices, axis=0) <= (xlim[1], ylim[1])).all()
+
+
+@pytest.mark.natural_earth
+@ImageTesting([grid_label_intersect_map_image],
```
greglucas, line 373: Do we need another 1 MB image comparison test? Or can this be checked by comparing against the positions of the labels and whether they are visible or not?

Or maybe just a single smaller image would be fine.

Hakkeh4x0r: Yes that could work. I'll make some new test tomorrow.

Hakkeh4x0r:
@greglucas i did some changes based on stefraynaud feedback.

The user should only operate padding values as positive integer and adjust labels inside with the class input parameter. 

However, in the class the _x/_ypadding is set to negative if labels_inside is true. The same goes for the initial value _x/_ypadding.

greglucas (commented):
I'm a little confused now. Why can't we allow negative values in the padding? What if I want x labels inside, but ylabels outside the plot? I guess it seems to me that if I want inside_labels, then I should add a negative padding. So, then in this code I assume it you could actually do the test
`if xpadding < 0` then `inside_x_labels = True` and make all your adjustments later based on that? So, rather than adding a new keyword argument, we could just implicitly determine that based on whether the padding arguments are positive or negative?

emsterr:
@greglucas i do agree with you. As you mention in your comment, I inital implemented this with the possibility to operate with negative input values, but as stefraynaud sugested (and had implemented in his solution) i changed it. 

I will revert to the previous implementation and remove the **labels_inside** input variable.

As you mention, Regarding the x and y padding,  _What if I want x labels inside, but ylabels outside the plot?_, should the padding top/bottom and left/right be adjusted as one, or is it a solution to seperate all of them. Example where you want bottom to be inside but not top.

x_top_padding, x_bottom_padding...
 
It would give more freedom, but, I think its a bit cumbersome and messy to seperate all sides.

Suggestions?

stefraynaud:
I think that we code that is compute the effective padding must revisited to reach the requested behavior:
https://github.com/emsterr/cartopy/blob/e0a05d8bd4082474bcbcbf024a2d57312188edd2/lib/cartopy/mpl/gridliner.py#L736
```python
dx = ypadding * np.cos(angle * np.pi / 180)
dy = xpadding * np.sin(angle * np.pi / 180)
```
Only the xpadding must be used as a distance for xlabels and only the ypadding must be used as a distance for ylabels.
```python
# For xlabels
dx = xpadding * np.cos(angle * np.pi / 180)
dy = xpadding * np.sin(angle * np.pi / 180)

# For ylabels
dx = ypadding * np.cos(angle * np.pi / 180)
dy = ypadding * np.sin(angle * np.pi / 180)
```
I also have to take this into account in my PR.
My two cents.

emsterr:
@stefraynaud, with the implementation i had where negative padding was possible, the behaviour was as expected.

What will differ with this implementation?
````python
# For xlabels
dx = xpadding * np.cos(angle * np.pi / 180)
dy = xpadding * np.sin(angle * np.pi / 180)

# For ylabels
dx = ypadding * np.cos(angle * np.pi / 180)
dy = ypadding * np.sin(angle * np.pi / 180)
````

stefraynaud:
@emsterr I have trouble to find the implementation you had and I suspect that I misunderstood the meaning of x and y padding. So sorry for the noise and I will also update my PR accordingly.

emsterr:
No worries, i squased my commits to the later implementation, so its not in the branch. I will implement the former solution again.

emsterr:
@stefraynaud God morning. Did you get a PR regarding negative padding into v0.19?

stefraynaud:
Hi @emsterr my PR includes it and I'm about to update the baseline images once my ssh connexion is allowed.
However, I'm not sure it's gonna be included in 0.19.

emsterr:
@stefraynaud 
Ok, from my perspective it would be good to get either yours or mine PR into v0.19 so i can remove my fork that is used by my software. Dont think i need to change to much if your version is merged.

As long as one of the PRs gets in i can take my shoulders down.

@greglucas, whats your thought on this?

smartlixx:
V0.19 has been released. You should aim at V0.20.

emsterr:
that's unfortunate. Hopefully it will take a shorter amount of time from v0.19 to v0.20 as from v0.18 to v0.19.

@greglucas Are there any plans on increasing the amount of releases in the near future?

greglucas:
Yeah, I'd like to get to about a 6 month release schedule to more closely track other package's nominal release schedule as well. But, it all depends on volunteer hours submitting and reviewing PRs and we can't guarantee anything for the schedules.

QuLogic:
Closed by #1537

QuLogic:
OK, I'm not actually sure whether this really is replaced by #1537, but @stefraynaud included it in the list.

stefraynaud:
@QuLogic I confirm: a negative padding should do the trick.

Issue #1671, opened 2020-10-21
Label formatting with dms=True outputs hemisphere behind degrees (36°**N**29’40”), What I need is to move it to the end (36°29’40”**N**).


emsterr:
### Description

Label formatting with `dms=True` outputs hemisphere behind degrees (36°**N**29’40”), What I need is to move it to the end (36°29’40”**N**).

From ticker.py in method `_format_value(..)` this is set in the last if-test. 

```python
class _PlateCarreeFormatter(Formatter):

...

 def _format_value(self, value, original_value):
        hemisphere = self._hemisphere(value, original_value)

        if not self._dms:
            return (self._format_degrees(abs(value)) +
                    hemisphere)

        value, deg, mn, sec = self._get_dms(abs(value))

        # Format
        label = ''
        if sec:
            label = self._format_seconds(sec)

        if mn or (not self._auto_hide_minutes and label):
            label = self._format_minutes(mn) + label

        if not self._auto_hide_degrees or not label:
            label = self._format_degrees(deg) + hemisphere + label

        return label
```
My local solution was just to add the hemisphere at the end ´label = self._format_degrees(deg) + label + hemisphere` to obtain the correct output, but that is of course not possible in Cartopy production. So my question is, can i create a valid solution and request a PR?

I was thinking making a dynamic solution where it was possible to send in the order wanted. 

Pseudo code: `example_method(format_order="DMSH"...)` or something in that direction.

greglucas:
It looks like the degree formatting is a relatively new addition in #1117. https://github.com/SciTools/cartopy/commit/924fd87eff9a8d6a8682954189099b82145ad334
I would tend to agree with you that I'd expect the hemisphere label at the end. I can't find anything on the web where the current format is even used, so maybe it was just an oversight?
(I don't use DMS myself, so I'm not an expert on this at all)

emsterr:
I havent seen the way it is labeled now before, but it might be a reason. @stefraynaud maybe you have some insight? 

Should it be changed so it is at the end of the label, or maybe create a dynamic format method so we dont mess up for someone else? For all I know there is someone out there requesting it as it is.

stefraynaud:
I agree that the new format is logical and must be adopted.
Maybe this could do the job:
```python
        if not self._auto_hide_degrees or not label:
            label = self._format_degrees(deg) + label + hemisphere 
```
The hemisphere symbol is now at the right.

emsterr:
[...]

PR #1672, opened 2020-10-23
change hemisphere location i gridlines labels.

emsterr:
change hemisphere location i gridlines labels. The hemisphere is now moved to the end of the label text, instead of after the degree symbol. (12°N34'45" -> 12°34'45"N)

Discussed in Issue #1671 

## Rationale
The usual presentation of degrees, minutes, seconds  is with hemisphere at the end of the text.


greglucas (approved):
Looks good to me. I'm going to approve and leave it up for a bit in case anyone else has an opinion that this needs a deprecation notice. I'd rather just fix it and not add bloat with more keywords though, so this gets my vote.

greglucas:
@emsterr, it looks like there are some merge conflicts, can you rebase this onto the latest master to update the labels?

emsterr:
@greglucas, I will merge it in and verify.

emsterr:
Hello @greglucas, I have been away for some weeks, but will continue where i left of. If I need to hurry due to new release, just let me know.

greglucas (approved):
Looks good to me. It'd be good to get one more approval since this is actually breaking what was done, but I think it is worth it to change the strings without any deprecation dance since I couldn't find anywhere else that style was used.

greglucas:
Thanks, @emsterr! I think this is the most accepted formatting, so a good change to the defaults.

PR #1680, opened 2020-11-18
Restore PEP-517 compliance

### LINKED ISSUE #1552: pip install of cartopy 0.18.0 fails when installing numpy at the same time
### Description
I am provisioning a docker image using pip. In a single pip command, I installed a number of packages, including cartopy and numpy. This worked in versions prior to 0.18.0, and no longer works with 0.18.0.

#### Code to reproduce

In a docker image with vanilla python3 install and no pip packages installed, run

```
pip3 install --upgrade pip   && pip3 install --no-cache-dir cartopy==0.18.0 numpy
```

#### Traceback 

```
    ERROR: Command errored out with exit status 1:
     command: /usr/local/pyenv/versions/3.7.6/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-cdx7ek5c/cartopy/setup.py'"'"'; __file__='"'"
'/tmp/pip-install-cdx7ek5c/cartopy/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(cod
e, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-377te2k_
         cwd: /tmp/pip-install-cdx7ek5c/cartopy/
    Complete output (12 lines):
      File "/tmp/pip-install-cdx7ek5c/cartopy/setup.py", line 43, in <module>
        import numpy as np
    ModuleNotFoundError: No module named 'numpy'

      File "/tmp/pip-install-cdx7ek5c/cartopy/setup.py", line 45, in <module>
        raise ImportError('NumPy 1.10+ is required to install cartopy.')
    ImportError: NumPy 1.10+ is required to install cartopy.
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

<details>

### Operating system

Ubuntu 18.04
Python 3.7.6 installed via pyenv

### Cartopy version

0.18.0

</details>

nbren12:
## Rationale

pip installs are not currently compliant with PEP-517. This PEP is relied upon by external tooling. This behavior was reverted in 454a9118f711d111b36026b126724cdb8dd705e7 because it made it harder to test cartopy against a set particular version of numpy. IMO it is not necessary to remove PEP-517 compliance to enable this testing pattern. Using `python setup.py develop` versus `pip install` is enough. See this example:

```
bash-5.0$ python -m venv .env
bash-5.0$ source .env/bin/activate
(.env) bash-5.0$ pip install numpy==1.18
Collecting numpy==1.18
  Using cached https://files.pythonhosted.org/packages/f0/14/f71a89e03578084111e352f464d9f3b7f701ebbecbd1a60e89c96983ef77/numpy-1.18.0-cp37-cp37m-macosx_10_9_x86_64.whl
Installing collected packages: numpy
Successfully installed numpy-1.18.0
(.env) bash-5.0$ pip install .
Processing /Users/noah/workspace/cartopy
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  Complete output from command /Users/noah/workspace/cartopy/.env/bin/python /Users/noah/workspace/cartopy/.env/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/l1/hgxljmwd51vdlxhnnd272mqm0000gn/T/tmpd1r2bc61:
  MY NUMPY VERSION IS 1.19.4

  ----------------------------------------
Command "/Users/noah/workspace/cartopy/.env/bin/python /Users/noah/workspace/cartopy/.env/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/l1/hgxljmwd51vdlxhnnd272mqm0000gn/T/tmpd1r2bc61" failed with error code 1 in /private/var/folders/l1/hgxljmwd51vdlxhnnd272mqm0000gn/T/pip-req-build-i17lx8mw
(.env) bash-5.0$ python setup.py  develop
MY NUMPY VERSION IS 1.18.0
```

## Implications

Cannot pip install cartopy with one `pip` command.

## Checklist

 * If you have not already done so, ensure you've read and signed the Contributor Licence Agreement (CLA).
   (See the [governance page](http://scitools.org.uk/governance.html) for the CLA and what to do with it).

 * If this is a new feature, please provide an example of its use in the description. We may want to make a
   follow-on pull request to put the example in the gallery!

 * Ensure there is a suitable item in the cartopy test suite for the change you are proposing. 

-->

Resolves #1552.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ cp-run: &cp-install
   name: Install Cartopy
   command: |
     source activate test-environment
-    pip install -ve .
+    python setup.py develop
 
 doc-run: &doc-build
   name: Build documentation
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "numpy>=1.10" "cython>=0.28"]
+requires = ["setuptools >= 40.6.0", "numpy>=1.10", "cython>=0.28"]
 build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,17 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "numpy>=1.10", "cython>=0.28"]
+requires = [
+    "wheel",
+    "setuptools >= 40.6.0",
+    "Cython>=0.29.2",
+    "numpy==1.13.3; python_version=='3.5' and platform_system!='AIX'",
+    "numpy==1.13.3; python_version=='3.6' and platform_system!='AIX'",
+    "numpy==1.14.5; python_version=='3.7' and platform_system!='AIX'",
+    "numpy==1.17.3; python_version>='3.8' and platform_system!='AIX'",
+    "numpy==1.16.0; python_version=='3.5' and platform_system=='AIX'",
+    "numpy==1.16.0; python_version=='3.6' and platform_system=='AIX'",
+    "numpy==1.16.0; python_version=='3.7' and platform_system=='AIX'",
+    "numpy==1.17.3; python_version>='3.8' and platform_system=='AIX'",
+    # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
+    "numpy; python_version>='3.9'",
+]
 build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 40.6.0", "numpy>=1.10" "cython>=0.28"]
+build-backend = "setuptools.build_meta"
--- a/INSTALL
+++ b/INSTALL
@@ -56,7 +56,7 @@ for these packages can be found at https://github.com/conda-forge/feedstocks.
 For macOS, the required dependencies can be installed in the following way::
 
     brew install proj geos
-    pip3 install --upgrade cython numpy pyshp
+    pip3 install --upgrade pyshp
     # shapely needs to be built from source to link to geos. If it is already
     # installed, uninstall it by: pip3 uninstall shapely
     pip3 install shapely --no-binary shapely
@@ -78,12 +78,6 @@ to the core packages.
 Further information about the required dependencies can be found here:
 
 **Python** 3.5 or later (https://www.python.org/)
-
-**Cython** 0.28 or later (https://pypi.python.org/pypi/Cython/)
-
-**NumPy** 1.10 or later (https://numpy.org/)
-    Python package for scientific computing including a powerful N-dimensional
-    array object.
 
 **GEOS** 3.3.3 or later (https://trac.osgeo.org/geos/)
     GEOS is an API of spatial predicates and functions for processing geometry
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,10 @@ from distutils.spawn import find_executable
 from distutils.sysconfig import get_config_var
 
 from setuptools import Command, Extension, convert_path, setup
+import numpy
+
+print("MY NUMPY VERSION IS", numpy.__version__)
+sys.exit(1)
 
 """
 Distribution definition for Cartopy.


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 40.6.0", "numpy>=1.10", "cython>=0.28"]
```
graingert, line 2: This will fail if pip builds a wheel based on a numpy with a different ABI to the numpy installed due to install_requires

nbren12: Sure, but does this need to be addressed by this PR? This sounds like more of an issue with building/distributing wheels in general and not really specific to this PR. Generalizing a bit, it sounds like your concern is that a wheel will be incompatible with future versions of one of the dependencies, numpy in this case. This problem is not specific to numpy, and could occur if any of cartopy's dependencies make a breaking change.

graingert: Yes, this is different because the whl created and cached will be ABI incompatible and so a simple reinstall won't fix it

nbren12: I understand that numpy is both a build time requirement and a runtime requirement, which means that any given wheel may not work for all the versions of numpy that it claims to work with. 

Are you more concerned about the build process failing or a later user installation failing? 

I think the former is pretty unlikely. Numpy would have to issue an ABI breaking release between when the wheel is built and tested. As for the latter, I still don't see how this is specific to numpy since no build artifact (even an sdist) can ensure forward compatibility against all future releases of dependencies. For instance, the [xarray v0.16.1  sdist on pypi](https://files.pythonhosted.org/packages/70/54/d3416261ec3e427996f196a1a5489c553d4a4624eb61932d090f6d7c7680/xarray-0.16.1.tar.gz) claims to work with any pandas > 0.25. This is a lie, it doesn't work with pandas 1.0.0. This is why pinning package versions is so useful for application developers.

graingert: When you install a package from pypi it builds the wheel and caches it locally. If you install the same package and version in another virtual environment with a numpy with a different ABI it will install but not run correctly

nbren12: I understand that, but doesn't this problem exists for the current master? How is this specific to this PR?

graingert: With this PR it's more likely to occur as the wheel will only compile with the latest numpy and make it impossible to use with older versions

graingert: Here's some more details: https://github.com/numpy/numpy/issues/5888#issuecomment-313283876

nbren12: Also, the spirit of this PR is to facilitate using external tooling (e.g. poetry) that would explicitly allow you to avoid this potentially inevitable situation when using floating releases. The issue is not specific to numpy ABI-related incompatibilities, it could occur for any type of breaking change (e.g. pure python API breaking change). I think the issue you are raising is that a source distribution (sdist) could have better forward compatibility than a compiled artifact (wheel). I agree, but this is again an issue with wheels in general, and not this PR.

graingert: > it could occur for any type of breaking change (e.g. pure python API breaking change).

No it cannot. This is specific to numpy ABI compatibility

graingert: https://github.com/numpy/numpy/issues/5888#issuecomment-503563618

> If I recall correctly, SciPy pins its build system to the oldest numpy it wishes to support, but tests against later numpy versions.

cartopy should copy scipy in this regard

graingert:
```suggestion
requires = ["setuptools >= 40.6.0", "numpy==1.10", "cython>=0.28"]
```

graingert: or even better the correct version for each py version:

https://github.com/scipy/scipy/blob/master/pyproject.toml#L6

graingert:
```suggestion
requires = [
    "setuptools >= 40.6.0",
    "cython>=0.28",
    "numpy==1.16.5; python_version=='3.6'",
    "numpy==1.16.5; python_version=='3.7'",
    "numpy==1.17.3; python_version=='3.8'",
    # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
    "numpy; python_version>='3.9'",
]
```

graingert: see also NEP 29: https://github.com/SciTools/cartopy/issues/1518

nbren12: > cartopy should copy scipy in this regard

Sounds good to me. It sounds like strategy will ensure the widest possible range of compatibility.

graingert:
```suggestion
requires = [
    "wheel",
    "setuptools",
    "Cython>=0.29.18",
    "numpy==1.16.5; python_version=='3.7'",
    "numpy==1.17.3; python_version=='3.8'",
    # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
    "numpy; python_version>='3.9'",
]
```

if you want to follow NEP29

.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ cp-run: &cp-install
   name: Install Cartopy
   command: |
     source activate test-environment
-    pip install -ve .
+    python setup.py develop
```
graingert, line 56:
```suggestion
    pip install -ve .
```

pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[build-system]
+requires = [
+    "wheel",
+    "setuptools >= 40.6.0",
+    "Cython>=0.29.2",
+    "numpy==1.13.3; python_version=='3.5' and platform_system!='AIX'",
```
graingert, line 6: I looked around a bit more and found this even nicer solution:

https://github.com/scipy/oldest-supported-numpy

pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+    "numpy==1.13.3; python_version=='3.6' and platform_system!='AIX'",
+    "numpy==1.14.5; python_version=='3.7' and platform_system!='AIX'",
+    "numpy==1.17.3; python_version>='3.8' and platform_system!='AIX'",
+    "numpy==1.16.0; python_version=='3.5' and platform_system=='AIX'",
+    "numpy==1.16.0; python_version=='3.6' and platform_system=='AIX'",
+    "numpy==1.16.0; python_version=='3.7' and platform_system=='AIX'",
+    "numpy==1.17.3; python_version>='3.8' and platform_system=='AIX'",
+    # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
+    "numpy; python_version>='3.9'",
```
graingert, line 15:
```suggestion
    "oldest-supported-numpy",
```

graingert:
@nbren12 I've got a new version here https://github.com/SciTools/cartopy/pull/1681 using oldest-supported-numpy

nbren12:
@graingert Your PR looks good to me. Should I close this one?

graingert:
@nbren12 I'm using #1681 in production now and it seems to work well

greglucas:
This was superseded by #1681 which was just merged in. Looks like it fixed your issues @nbren12?

nbren12:
Indeed. I'll close this PR. Thanks!

PR #1691, opened 2020-12-11
TST: Using GitHub Actions for CI testing

greglucas:
Adding the yaml configuration file to use GitHub Actions for testing.

## Rationale

TravisCI may no longer be supporting open-source the way it previously did with free resources.

## Implications

New test integration platform and needing to learn the GitHub Actions yaml language and platform.


.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,93 @@
...
+        PACKAGES="$PACKAGES setuptools_scm_git_archive shapely"
+        conda install $PACKAGES
+        conda info -a
+        conda list
+
+    - name: Install Cartopy
+      run: |
+        MPL_CONFIG_DIR=~/.config/matplotlib
+        mkdir -p $MPL_CONFIG_DIR
+        echo "backend" ":" "agg" > $MPL_CONFIG_DIR/matplotlibrc
```
QuLogic, line 62: This doesn't work?
```suggestion
        echo "backend : agg" > $MPL_CONFIG_DIR/matplotlibrc
```

.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,93 @@
+name: Cartopy Testing
```
QuLogic, line 1: I'd just name it
```suggestion
name: Tests
```

.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,93 @@
...
+      run: |
+        MPL_CONFIG_DIR=~/.config/matplotlib
+        mkdir -p $MPL_CONFIG_DIR
+        echo "backend" ":" "agg" > $MPL_CONFIG_DIR/matplotlibrc
+        pip install --no-deps -e .
+        python -c "import cartopy; print('Version ', cartopy.__version__)"
+
+    - name: Testing
+      id: test
+      # Check that the downloader tool at least knows where to get the data from (but don't actually download it)
```
QuLogic, line 68: This comment is only about the first command in `run`, and can go there.

.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        python-version: [3.6, 3.8]
+    defaults:
+      run:
+        shell: bash -l {0}
```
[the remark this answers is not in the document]
greglucas: I was initially testing on Windows as well. Removed it now.

.github/workflows/ci-testing.yml
greglucas: I put it above so that it didn't get "run" on the command line due to the pipe operator, but looks like that doesn't harm anything either.

.github/workflows/ci-testing.yml
greglucas: Copy/pasted from before. Not sure why it was that way, good catch.

.github/workflows/ci-testing.yml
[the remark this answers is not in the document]
greglucas: I take it back. CI fails if I don't use this when installing Cartopy in the setup.py file at `file=sys.stderr`. So, maybe to access stderr you need a login shell?

greglucas:
Coveralls on GitHub reports 85%, but on TravisCI it reports 90% which is the reason for the drop once I removed the TravisCI builds.

.github/workflows/ci-testing.yml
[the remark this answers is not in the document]
dopplershift: Conda needs that.

dopplershift:
Was that a bug or did we suddenly stop running some important config? Or do we have code we can remove because we're not supporting some combination we were hacking around?

greglucas:
It looks like we are now testing a new file: `_geodesic.pyx` in https://coveralls.io/builds/35765159/source?filename=lib/cartopy/geodesic/_geodesic.pyx vs https://coveralls.io/builds/35766611 that only tests the `__init__.py` file from the geodesic package. That new file isn't fully covered which is what is causing the overall coverage to decrease. So, nothing has actually decreased as far as I can tell, just now testing more things.

dopplershift (approved):

QuLogic (approved):
Best to just start testing with it.

Issue #490, opened 2014-10-07
Scale bar


ThomasLecocq:
Hi all

is there a way to add a scale bar in a corner of the plots ?

Thanks

Tom

gauteh:
I'd really like to see something like this too.

sebhahn:
anybody working on this feature?

bladeoflight16:
There's an existing library for adding a scalebar to matplotlib that might be of help: https://pypi.python.org/pypi/matplotlib-scalebar.

eogit:
This function may be useful:

# plot a scale bar with 4 subdivisions on the left side of the map
def scale_bar_left(ax, bars=4, length=None, location=(0.1, 0.05), linewidth=3, col='black'):
    """
    ax is the axes to draw the scalebar on.
    bars is the number of subdivisions of the bar (black and white chunks)
    length is the length of the scalebar in km.
    location is left side of the scalebar in axis coordinates.
    (ie. 0 is the left side of the plot)
    linewidth is the thickness of the scalebar.
    color is the color of the scale bar
    """
    # Get the limits of the axis in lat long
    llx0, llx1, lly0, lly1 = ax.get_extent(ccrs.PlateCarree())
    # Make tmc aligned to the left of the map,
    # vertically at scale bar location
    sbllx = llx0 + (llx1 - llx0) * location[0]
    sblly = lly0 + (lly1 - lly0) * location[1]
    tmc = ccrs.TransverseMercator(sbllx, sblly)
    # Get the extent of the plotted area in coordinates in metres
    x0, x1, y0, y1 = ax.get_extent(tmc)
    # Turn the specified scalebar location into coordinates in metres
    sbx = x0 + (x1 - x0) * location[0]
    sby = y0 + (y1 - y0) * location[1]

    # Calculate a scale bar length if none has been given
    # (Theres probably a more pythonic way of rounding the number but this works)
    if not length:
        length = (x1 - x0) / 5000  # in km
        ndim = int(np.floor(np.log10(length)))  # number of digits in number
        length = round(length, -ndim)  # round to 1sf

        # Returns numbers starting with the list
        def scale_number(x):
            if str(x)[0] in ['1', '2', '5']:
                return int(x)
            else:
                return scale_number(x - 10 ** ndim)

        length = scale_number(length)

    # Generate the x coordinate for the ends of the scalebar
    bar_xs = [sbx, sbx + length * 1000 / bars]
    # Plot the scalebar chunks
    barcol = 'white'
[...]

PR #1693, opened 2020-12-16
Addition of new functionalities for Cartopy.gridliner

PhilipeRLeal:
Dear all,

I would like to make a pull request in cartopy in order to add new functionalities for cartopy.

## Rationale

These new assets have been derived from my personal struggle with cartopy's gridline and the continuous migration from the past basemap towards cartopy.

In this new branch, the gridline class has gained new assets for easier usage, especially regarding ticklabel_settings. 

## The list of the new functionalities related to gridline ticklabel settings:

1. addition of a method for easy setting the decimal separators from the gridline ticklabels
2. addition of a method for easy setting the maximum number of ticks in the gridline

## Changes in respect to basemap->cartopy migration:

There has been a lot of discussion regarding scalebars (rulerbars) for geographical data in python, especially regaring cartopy: 

1. [ref 1 - stackoverflow] (https://stackoverflow.com/questions/32333870/how-can-i-show-a-km-ruler-on-a-cartopy-matplotlib-plot)
2. [ref 2 - stackoverflow] (https://stackoverflow.com/questions/39786714/how-to-insert-scale-bar-in-a-map-in-matplotlib)
3. [ref 3 - cartopy Issue #490] (https://github.com/SciTools/cartopy/issues/490)

Therefore, in this branch, I also developed a scalebar module, from which the user can draw a geographical rulerbar much similar to the ones observed in Geographical Information Systems (GIS). At the moment, I have set the module inside the cartopy.mpl subdirectory. That way, the manipulation of the geographical information becomes more organized in a same subpackage.

One aspect that may cause some constraints is that the scalebar module requires the [geopy ](https://geopy.readthedocs.io/en/stable/) package. In future versions, its dependency may be broken by altering the geopy functions applied in the scalebar module.

## Implications

addition of a dependency of geopy package solely for the scalebar module.


PhilipeRLeal:
I state that I have now signed the Contributor's Licence Agreement (CLA) document.

greglucas:
Thanks for the PR @PhilipeRLeal!

There are a lot of extraneous changes here to what you're trying to accomplish which makes it hard to review (i.e. linting changes in unrelated files). It looks like there are two things you're trying to put in, a new formatter and a new scalebar? I'd suggest making these two separate smaller PRs as I think they are two distinct features.

Can you rebase and squash into one or two meaningful commits to make the review easier?

You have two tests failing, one license file header and one in the new test you added test_scalebar. See here for those failures: https://travis-ci.org/github/SciTools/cartopy/jobs/750254210

The SciTools CLA checker will automatically get updated when someone has the time to review your submitted CLA and they'll automatically remove it I think.

PhilipeRLeal:
> rebase

Dear GregLucas, thank you for your suggestions.

This is my first commit to a open source library as cartopy. I still don't know what "rebase" and "squash" mean.

By any chance, "rebase" would be changing the pull request's Title, or hash code (i.e.:  #1693)? If so, how can I do it?
The same question for "squash".

Nevertheless, you are right about the things I am trying to put in this pull request. There are two changes that I am proposing.

First, I am trying to add a new customization option for the gridlines. This problem is presented in a comment from the stackoverflow [(see here)](https://stackoverflow.com/questions/54580278/why-cant-one-set-to-false-specific-axis-ticklabels-ex-xlabels-top-ylabels). At the current time, cartopy only supports a fixed set of geographical units for the xticklabels and yticklabels (i.e.: "W" for west coordinates, "E" for east coordinates, "S" for south coordinates, "N" for north coordinates), with no easy way to change them to a user's defined version. Also, there is no easy way to convert the decimal symbol of the coordinates (i.e.: dot "."). For users like me who come from Latin countries, there is a great demand for latin symbols, therefore, instead of using [W,E,S,N and dot"."], one should use the following respectively [O, L, S, N and ","].

Here is a relation of the provided symbols:

O (from Oeste) == West
L (from Leste) == East
N (from Norte) == North
S (from Sul) == South

I state here that all prior symbols [W,E,S,N and dot"."] have been kept as they were (in international format). Therefore, no changes have been made to the common cartopy's gridline usage. This new Pull request only allows the user to easily change those symbols into any user's defined string/symbol.

# --------------------------------------------------------------------------------------------------------------- #

Lastly, I am also proposing a method to draw a scalebar to cartopy's geoaxes. This addition would be directly callable from the geoaxes, and it should be capable of evaluating metric distances (i.e.: meters and km units) in axesTransAxes units.

Prior to cartopy's development, the mpl.basemap had something similar, which allowed easy evaluation and plotting of a scalebar into one's geographical Map. Nevertheless, despite continuous development in cartopy's, this functionality has not been formally proposed.

This scalebar method should also be prone to user's modifications as, for example, setting the decimal separator and unit symbology (i.g.: "km", "mi", etc.).

# --------------------------------------------------------------------------------------------------------------- #

Both new additions have been tested, and are currently running in my computer. Nevertheless, I use Windows, and therefore, I can't properly create a build for the package prior a commit due to GEOS interface. That is one of the main issues that I am struggling with in this pull request.

I believe that this new pull request can truly help Python's community, specially cartopy's.

I put myself at your disposal, and I hope to hear from you soon.

Sincerely,

Philipe Riskalla Leal

greglucas:
Welcome to the open source community :)

For rebasing and squashing I was referring specifically to `git` (not GitHub). So on the command line (or with some IDEs maybe through an interface) `git rebase -i master` when you're on your branch. **However, I just noticed that you're on your own `master` branch, so you would want to checkout a new branch locally first: `git checkout -b scalebar`, then start working on your rebase onto the other remote master branch.**

The whole reason for that is right now I see 97 commits and 100+ files changed, so git needs to know what it is merging against to properly see new vs old commits.

The reason for requesting two separate PRs for this work is that they are distinct entities and maybe one will be accepted and another not, or one needs lots of changes but the other doesn't. It is just easier to work on smaller chunks at a time.

PR #1700, opened 2020-12-23
apply NEP29

### LINKED ISSUE #1518: Adopt NEP29: minimum dependency policy

graingert:
Fixes #1518


graingert:
Looks like the image changed with v3.7 but I don't have a Mac to generate it on

greglucas:
As a general comment, I think this is a good thing, but that it should wait until after the 0.19 release to go in. It'd be nice to keep the lower Python support on 0.19 since that will be our first release to officially drop Python 2 and make it an easier transition if people are still running Python 3.6.

The matplotlib `conda` is downloading isn't linked to the proper freetype, which I think may be what is actually causing the image failure (MPL should be downloaded from the label/testing channel). However, I'm not sure why the `Upload Images` step isn't getting run... I put that in there so we could easily get the image results out of the CI system to re-use in situations like this. I swear I tested it on my fork and it worked for me, so I'm not sure what changed.

Specifically, this isn't getting called: https://github.com/SciTools/cartopy/blob/7237260b04f6d803c4b0f0f6095a27a4f83adaef/.github/workflows/ci-testing.yml#L81
So maybe the `test` step conclusion isn't `failure`, or it fails fast on this job? No need for you to look into this one if you don't want, just putting this down for future reference.

greglucas:
@graingert, v0.19 is out now, so I think this would be OK to put in now and drop some older support dependencies. Can you rebase and we'll see if the macos build is still failing?

QuLogic:
It looks like there are still conflicts?

setup.py
```diff
@@ -179,7 +179,7 @@ def get_proj_libraries():
     proj_libraries = ["proj"]
     if os.name == "nt" and (6, 0, 0) <= proj_version < (6, 3, 0):
         proj_libraries = [
-            "proj_{}_{}".format(proj_version[0], proj_version[1])
+            f"proj_{proj_version[0]}_{proj_version[1]}"
```
QuLogic, line 182: You can't make this change to `setup.py`; it needs to remain Python 2 compatible.

graingert: No it doesn't, the requires_python will prevent pip from downloading the sdist

QuLogic: It does; read the comment at the top.

greglucas:
Updated and incorporated in https://github.com/SciTools/cartopy/pull/1844

PR #1701, opened 2020-12-24
FIX: Addressing flake8 complaints

greglucas:
Not sure why Stickler isn't finding these... I just ran `flake8 lib/cartopy` to find these two issues. Maybe we should add that to a separate GitHub Action?


dopplershift (approved):

dopplershift:
I have no problem with a linting GitHub action--that's what we do on MetPy. There's a tool called `reviewdog` that makes it easy to turn flake8 issues into PR annotations.

PR #1706, opened 2021-01-07
SciPy dependency missing

xoolive:
See here: https://github.com/SciTools/cartopy/blob/545832ebfe18f7173559e2403fbd35e0ee900233/lib/cartopy/img_transform.py#L17

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,3 +1,4 @@
 numpy>=1.13.3
 shapely>=1.5.6
 pyshp>=2
+scipy


dopplershift:
Scipy is currently listed in `requirements/plotting.txt`--so it's "optional", as described in `INSTALL`. I'm not sure if that's a good thing or not. Certainly we'd need to adjust those if we were to go this way.

xoolive:
I raised upon this issue when I put a dependency on cartopy in one of my packages and CI failed because of a scipy dependency missing, raised at the line mentioned above. So I assume there is probably something to be fixed somewhere... 😁

xoolive:
Should we leave this PR open or not?

QuLogic:
We take those requirements and define extras out of them, so if you need to do plotting, you should depend on `cartopy[plotting]` instead of plain `cartopy`.

xoolive:
Using cartopy without plotting sounds really weird to me...

greglucas:
I agree with @xoolive, I'm not sure if anyone uses Cartopy without plotting/MPL anymore? Pyproj will handle the PROJ transforms and Shapely/pygeos can handle the GEOS integrations. I think it would be reasonable to bump `cartopy[plotting]` into the main requirements.

PR #1710, opened 2021-01-10
Transform points

### LINKED ISSUE #1705: If point is outside the limits of Orthographic projection, the point defaults to the center of the plot
Is there a way to programmatically clip it instead?
```
import cartopy.crs as ccrs

ax = plt.subplot(projection=ccrs.Orthographic())
ax.coastlines()
ax.scatter(-120, 80, transform=ccrs.PlateCarree())
ax.set_global()
```
image

greglucas:
Closes #1705 

## Rationale

When a single point was passed to `transform_points()` PROJ returns a bad status that is not caught and therefore we didn't update the result array properly.

1st commit: Fixes this to return nan's in the length-1 case
2nd commit: Combines the logic from transform_point and transform_points to remove duplicate code
3rd commit: Adds a test for these cases

## Further considerations

PROJ returns 'inf' when the length of the array is greater than 2, but we fill in nan's when it is length-1 and PROJ returned a bad status code. This is inconsistent in the fact that repeating the same coordinates gives two different values depending on the situation. Should we consider changing the nan-fill into an inf-fill instead so we are consistent? Converting all inf's to nan's didn't directly work:  `result[~np.isfinite(result)] = np.nan` causes failures for two gridliner_inline tests (shapely bad GEOS geometries)

### DIFF
--- a/lib/cartopy/_crs.pyx
+++ b/lib/cartopy/_crs.pyx
@@ -353,26 +353,11 @@ cdef class CRS:
 
         """
         cdef:
-            np.ndarray[np.double_t, ndim=1] cx, cy
-            int status
-        cx = np.array([x])
-        cy = np.array([y])
-        if src_crs.is_geodetic():
-            cx *= DEG_TO_RAD
-            cy *= DEG_TO_RAD
-        status = _safe_pj_transform(src_crs, self, 1, 1, cx, cy, None)
+            np.ndarray[np.double_t, ndim=2] result
 
-        if trap and status == -14 or status == -20:
-            # -14 => "latitude or longitude exceeded limits"
-            # -20 => "tolerance condition error"
-            cx[0] = cy[0] = np.nan
-        elif trap and status != 0:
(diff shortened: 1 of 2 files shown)


lib/cartopy/_crs.pyx
```diff
@@ -446,15 +436,21 @@ cdef class CRS:
...
-            status = _safe_pj_transform(src_crs, self, npts, 3,
-                                        result[:, 0], result[:, 1],
-                                        result[:, 2])
+        if npts == 0:
+            return result
+        status = _safe_pj_transform(src_crs, self, npts, 3,
+                                    result[:, 0], result[:, 1],
+                                    result[:, 2])
+
+        if trap and status == -14 or status == -20:
```
QuLogic, line 445: So I'm fairly certain this was wrong before, but the expression doesn't make sense here. It should be 
```suggestion
        if trap and status in (-14, -20):
```

QuLogic (approved):

PR #1721, opened 2021-02-02
MNT: Removing numpy float/bool conversions

greglucas:
Numpy 1.20 raises a deprecation warning when using `np.float` and `np.bool` instead of Python's builtin `float`/`bool`. Replacing those throughout the codebase.

Should fix the current test failures due to additional warnings being raised.


greglucas:
I have no idea what is causing the gridliner failures now... It is likely some dependency update causing the issues (because the newer dependency tests pass), but I can't figure out what specifically.

QuLogic (approved):

QuLogic:
We can fix that later, I think.

PR #1724, opened 2021-02-03
Skipping bad GEOS version and updating Appveyor Python version

greglucas:
Pinning GEOS to 3.8.1, I'm not entirely clear why this is causing the gridliner to fail only on the lowest Python version.

Updating appveyor to use Python 3.9 to keep more current on all packages.

## Rationale

Passes tests

## Implications

The gridliner and GEOS 3.9 may have some bad interactions that are hidden by this pinning. But, it is likely only on 3.6/older package versions.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,6 +1,6 @@
 environment:
     matrix:
-        - PYTHON_VERSION: "3.6"
+        - PYTHON_VERSION: "3.9"
           CONDA_INSTALL_LOCN: "C:\\Miniconda36-x64"
           PACKAGES: "cython numpy matplotlib-base proj pykdtree scipy"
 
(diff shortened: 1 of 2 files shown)


greglucas:
It is actually something with GEOS 3.9. Apparently, the Python 3.9 conda's don't have GEOS 3.9 yet as I assumed... 

This may be fixed when shapely 1.8.0 is released (maybe a little ways off still though)? Ref: https://github.com/Toblerity/Shapely/issues/1079

For now, pinning to GEOS 3.8.1 for testing. If anyone has a better fix for this that would be great.

greglucas:
Update: My guess is this may be solved by the 3.9.1 GEOS bugfix release.
xref: https://github.com/libgeos/geos/issues/408
We'll have to wait until that version gets released for the CI systems to know for sure though.

greglucas:
I added a commit that skips the failing tests with GEOS version 3.9.0. It looks like 3.9.1 does indeed fix this, the macos-latest 3.6 build installs the 3.9.1 version.

QuLogic:
Is this updating the minimum, or updating the maximum run? I'm confused how this change has to do with minimums.

greglucas:
This was initially only failing on the minimum versions, so the title was different. I just updated it. This all boiled down to a bad GEOS 3.9.0, and it also keeps the Appveyor version bump because that was grabbing the bad version still. We are still testing the minimum dependencies on other CIs.

QuLogic (approved):

PR #1742, opened 2021-03-08
ci: Don't allow proj 8 yet

QuLogic:
## Rationale

Since proj 8 removed the `proj_api.h` header and we still use it, we need to pin <8 temporarily.

Also, remove check for Python 2 doc artifacts, since that build is gone.

## Implications

CI will build again.


greglucas (approved):
We should probably also add some documentation somewhere saying that we currently only support PROJ<8. That can come in another PR though.

PR #1743, opened 2021-03-09
Updating tests for MPL 3.4 release

greglucas:
## Rationale

Fix failing tests on release candidate.

Commits:
- added some private attributes to the sentinel renderer to satisfy the new raster path that checks for those.
- set `snap=False` for all pcolormesh image tests. We could update the images as well, but I figured this would be cleaner and less bloat on the repo for now.
- Update the contour label example image for 3.4. There were a few minor location changes in the locations of the labels and also a more major change of the "-120" location. It all seems to be auto-located, so I'm not sure how much that really matters? I updated the image test for 3.4

## Implications

All tests pass


QuLogic (approved):
Thanks for getting to this.

PR #1745, opened 2021-03-12
IO - Add .prj and .cpg to NaturalEarth unzipped files

### LINKED ISSUE #1744: Missing .prj for NaturalEarth data
### Description

With the current default downloader from ``cartopy``, not all elements from the NaturalEarth zip container are extracted.
Currently, [only shp, dbf, and shx](https://github.com/SciTools/cartopy/blob/8b106fec9511911eb49a7f78b65407895c37bfc3/lib/cartopy/io/shapereader.py#L319) are present.

This means that when loading the file (e.g. in my case with ``geopandas.read_file``) leads to a ``GeoDataFrame`` that has its ``crs`` element set to ``None`` instead of ``"epsg:4326"``.

### Feature request

Unless there is a good reason not to include the .prj file (that has already been downloaded anyways), I'd suggest to include it in the list of extensions [there](https://github.com/SciTools/cartopy/blob/8b106fec9511911eb49a7f78b65407895c37bfc3/lib/cartopy/io/shapereader.py#L319).
Given that it takes virtually no space, I'd also suggest to include the .cpg in case it might also be of use to somebody somewhere.

I can provide a PR if necessary.

tfardet:
Fixes #1744 

## Rationale

Adding .prj files is necessary to have the CRS data attached.
Adding the .cpg file may be useful (unknown) and the files are super lightweight anyways.

## Implications

Now loading data has CRS data attached (e.g. via the ``crs`` attribute in GeoPandas).
Since data is only added, this should not affect any existing code (unless they specifically rely on the absence of CRS information but that would be really weird).

I checked on the [NaturalEarth repo](https://github.com/nvkelso/natural-earth-vector) that all zip files should contain .prj and .cpg files.

Since the zip files downloaded contained the .prj and .cpg files already, this does not lead to any additional data load except for a (tiny) additional use of space on the local drive.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -316,7 +316,7 @@ class NEShpDownloader(Downloader):
         natural earth zip file.
 
         """
-        for ext in ['.shp', '.dbf', '.shx']:
+        for ext in ['.shp', '.dbf', '.shx', '.prj', '.cpg']:
             yield ('ne_{resolution}_{name}'
                    '{extension}'.format(extension=ext, **format_dict))
 


greglucas (approved):
Looks good to me.

PR #1747, opened 2021-03-13
Feature: add etopo plotting to stock_img

### LINKED ISSUE #1734: Plot etopo
I'd like to suggest a new function for cartopy. As far as I'm aware, plotting an etopo bathymetry map with cartopy isn't easy. Basemap has a convenient function to do this: https://basemaptutorial.readthedocs.io/en/latest/backgrounds.html#etopo.

It would be great if cartopy has something similar!

smartlixx:
## Rationale

Close #1734 

## Implications

None as far as I can see.

## Example

```python
import cartopy.crs as ccrs

fig = plt.figure(figsize=[4, 6])
ax2 = fig.add_subplot(2, 1, 2, projection=ccrs.Orthographic())

ax1.stock_img(name='etopo')
ax1.coastlines()
ax2.stock_img(name='etopo')
ax2.coastlines()

plt.show()
```
The output is 
Cartopy_etopo1


dopplershift:
Currently the PR is showing an empty etopo1.jpg committed? Any reason that needs to be included?

smartlixx:
The file etopo1.jpg is not empty. its size is 1.28M (64 dpi). it is the picture of the bathymetry data to plot. I took it from Basemao repo, because the original file downloaded from NOAA is about 2M (96 dpi). This file definitely will increase the size of cartopy. Alternatively, we can download it on the fly from NOAA website. Please let me know your opinion.

greglucas:
I think grabbing it from the source is better whenever possible, that way we get any updates they have. It also looks like there are multiple "versions" of etopo, focusing on ice, or other features too? It looks like the original data is 300+MB! I didn't see the downsampled version immediately on NOAA's site, do you have a link to that?

smartlixx:
> I think grabbing it from the source is better whenever possible, that way we get any updates they have. It also looks like there are multiple "versions" of etopo, focusing on ice, or other features too? It looks like the original data is 300+MB! I didn't see the downsampled version immediately on NOAA's site, do you have a link to that?

When you click on the thumbnail on https://www.ngdc.noaa.gov/mgg/global/, you will get to the downsampled version of etopo1: https://www.ngdc.noaa.gov/mgg/image/color_etopo1_ice_low.jpg. 

Yes, there are several versions of this dataset, depending on using the ice top or the ice bottom for altitude. The above link should suffice for our purpose. If we choose to grab it on the fly, its size is about 2M.

smartlixx:
Now I have modified the code to download the etopo1 image from NOAA website, and delete the etopo1.jpg file from the commit.

dopplershift:
Can you rebase your commits to "squash" together the original commit adding the file and the one deleting it? Otherwise, there will be a copy of the 1.3MB file sitting in the history for this repo.

smartlixx:
> Can you rebase your commits to "squash" together the original commit adding the file and the one deleting it? Otherwise, there will be a copy of the 1.3MB file sitting in the history for this repo.

Thanks. I squashed all the commits into one. 

The test failure message below seems not to be related with this PR.
```
FAILED lib/cartopy/tests/mpl/test_ticks.py::test_set_xticks_no_transform - Fi...
```

lib/cartopy/tests/mpl/test_images.py
```diff
@@ -179,6 +179,15 @@ def test_stock_img():
     ax.stock_img()
 
 
+@pytest.mark.xfail((5, 0, 0) <= ccrs.PROJ4_VERSION < (5, 1, 0),
+                   reason='Proj Orthographic projection is buggy.',
+                   strict=True)
+@ImageTesting(['imshow_etopo1_ortho'], tolerance=0.7)
```
QuLogic, line 185: It appears that you haven't added this test image?

smartlixx: Done.

marqh:
this PR appears to be reviwed and have passing tests, but there are conflicts between it and master

@smartlixx please may i suggest that you `rebase` this PR branch with respect to upstream/master

i would hope that reviewers would then be able to re-assess and potentially merge this onto master

@dopplershift @QuLogic is this reasonable from your points of view?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -988,9 +989,13 @@ def stock_img(self, name='ne_shaded'):
...
 
-        Currently, the only (and default) option is a downsampled version of
-        the Natural Earth shaded relief raster.
+        Currently, there are 2 options:
 
+        1. 'ne_shaded'(default) a downsampled version of the Natural Earth
+           shaded relief raster.
+        2. 'etopo' a downsampled version of global relief model of Earth's
+           surface that integrates land topography and ocean bathymetry. This
+           option is the same as the etopo from Basemap.
```
greglucas, line 998: Can you put in a link to each of these for the documentation so people can see what they are going to get?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -988,9 +989,13 @@ def stock_img(self, name='ne_shaded'):
...
         """
```
greglucas, line 999:
```python
if name not in ('ne_shaded', 'etopo'):
    raise ...
```
Then the if/elif below only need to update the path with a common return call.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -988,22 +989,38 @@ def stock_img(self, name='ne_shaded'):
...
+
         if name == 'ne_shaded':
             import os
             source_proj = ccrs.PlateCarree()
             fname = os.path.join(config["repo_data_dir"],
                                  'raster', 'natural_earth',
                                  '50-natural-earth-1-downsampled.png')
+        elif name == 'etopo':
+            import os
+            source_proj = ccrs.PlateCarree()
```
greglucas, line 1013: Move these outside of the if/elif clause because they are the same for both cases?

greglucas (approved):
Looks good to me, just one minor suggestion.

As a side comment/follow-up PR, would this make sense to have as a RasterSource feature artist? So, make it usable outside of the stock_img function, and have stock_img add the RasterSource features instead.

smartlixx:
> Looks good to me, just one minor suggestion.
> 
Thanks for your comments. I have updated the PR to incorporate the new changes since my last commit, and the PR has passed all the test. Hopefully it can be accepted (quite a long time has passed). Or if it is OK, I can also combine this PR with #2230.

> As a side comment/follow-up PR, would this make sense to have as a RasterSource feature artist? So, make it usable outside of the stock_img function, and have stock_img add the RasterSource features instead.

I can see there is a `RasterSource` class in `io/__init__.py` (https://github.com/SciTools/cartopy/blob/main/lib/cartopy/io/__init__.py#L310-L365) but most functions are not implemented yet. This may be a good starting point, but it's beyond my hands at this moment.

rcomer:
Hi, I’m just checking on the status of this one.  It looks like it was ready to go but unfortunately now has conflicts with the main branch again.  @smartlixx do you have the bandwidth to rebase this again?

Issue #1732, opened 2021-02-16
ax.set_xmargin doesn't seem responsive


ahuang11:
```
import os
import matplotlib.pyplot as plt

from cartopy import config
import cartopy.crs as ccrs


fig = plt.figure(figsize=(8, 12))

# get the path of the file. It can be found in the repo data directory.
fname = os.path.join(config["repo_data_dir"],
                     'raster', 'sample', 'Miriam.A2012270.2050.2km.jpg'
                     )
img_extent = (-120.67660000000001, -106.32104523100001, 13.2301484511245, 30.766899999999502)
img = plt.imread(fname)

ax = plt.axes(projection=ccrs.PlateCarree())
plt.title('Hurricane Miriam from the Aqua/MODIS satellite\n'
          '2012 09/26/2012 20:50 UTC')

# set a margin around the data
ax.set_xmargin(5.25)
ax.set_ymargin(2.5)

# add the image. Because this image was a tif, the "origin" of the image is in the
# upper left corner
ax.imshow(img, origin='upper', extent=img_extent, transform=ccrs.PlateCarree())
ax.coastlines(resolution='50m', color='black', linewidth=1)

# mark a known place to help us geo-locate ourselves
ax.plot(-117.1625, 32.715, 'bo', markersize=7, transform=ccrs.Geodetic())
ax.text(-117, 33, 'San Diego', transform=ccrs.Geodetic())

plt.show()
```
![image](https://user-images.githubusercontent.com/15331990/108110661-3051ce80-7059-11eb-9ef1-00c6079c405d.png)

smartlixx:
Actually, `set_ymargin` is not working as expected as well (there is no extra space at the bottom). This has been not working ever since v0.15, see [](https://scitools.org.uk/cartopy/docs/v0.15/matplotlib/advanced_plotting.html). A simple test with pure `Matplotlib` shows that it is not `Matplotlib`'s problem.

## A simplified example

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

fig = plt.figure(figsize=(8, 12))
ax = plt.axes(projection=ccrs.PlateCarree())
plt.title('Example of set_xmargin and set_ymargin')

ax.set_extent([90,150,10,50])

ax.set_xmargin(3)
ax.set_ymargin(5)

ax.coastlines(resolution='50m', color='black', linewidth=1)
ax.gridlines(draw_labels=True)

plt.show()
```
The output is 

[...]

PR #1750, opened 2021-03-17
Update advanced_plotting.rst to fix margins in the plot

smartlixx:
## Rationale
Based on the discussion in #1732, the code in advanced_plotting.rst needs to be updated to fix the margin problem.


dopplershift (approved):
Seems reasonable to get the plot working.

PR #1752, opened 2021-03-21
Initial port to proj8

manisandro:
This is work in progress, I'm working on it as time allowes as I need it to finish the proj-8 update in Fedora. The code compiles, but there are still many test failures which I need to investigate.

### DIFF
--- a/lib/cartopy/_crs.pxd
+++ b/lib/cartopy/_crs.pxd
@@ -4,8 +4,6 @@
 # See COPYING and COPYING.LESSER in the root of the repository for full
 # licensing details.
 
-from ._proj4 cimport projPJ
-
 
 cdef class CRS:
     """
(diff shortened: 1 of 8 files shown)


QuLogic:
Originally, I would have said that we should support both, but Proj 5 is now 3 years old, so I think we can switch APIs directly, unless anyone else has any objections?

That being said, I don't think the new API should break anything, so these test failures will probably need to be figured out first.

dopplershift:
I agree with @QuLogic on both accounts.

QuLogic:
@manisandro are you able to continue on this?

manisandro:
@QuLogic Currently I don't have any time to work on this

QuLogic:
Can you sign the CLA so we can use this?

QuLogic:
Replaced by #1808.

Issue #1761, opened 2021-04-13
Mapbox streets


lmoresi:
### Description

The interface to access the `mapbox.streets` tiles has been deprecated and attempting to use the current MapboxTiles interface is doomed (returns a 410 error)

#### Code to reproduce

```
import cartopy.io.img_tiles as cimgt
# previous interface which has been deprecated
mapbox_streets = cimgt.MapboxTiles(map_id='streets',  access_token='blah')

```

Below is a potential change to the MapboxTiles class but is not backwards compatible so not something I can suggest a fix for as a PR. 

```
class MapboxTiles(GoogleWTS):
    """
    Implement web tile retrieval from Mapbox.
    For terms of service, see https://www.mapbox.com/tos/.
    """
    def __init__(self, access_token, map_id):
        """
        Set up a new Mapbox tiles instance.
        Access to Mapbox web services requires an access token and a map ID.
        See https://www.mapbox.com/api-documentation/ for details.
        Parameters
        ----------
        access_token
            A valid Mapbox API access token.
        map_id
            An ID for a publicly accessible map (provided by Mapbox).
            This is the map whose tiles will be retrieved through this process.
        """
        self.access_token = access_token
        self.map_id = map_id
        super().__init__()

    def _image_url(self, tile):
        x, y, z = tile
        url = ('https://api.mapbox.com/styles/v1/{id}/tiles/{z}/{x}/{y}'
               '?access_token={token}'.format(z=z, y=y, x=x,
                                              id=self.map_id,
                                              token=self.access_token))
        return url


# new call
mapbox_streets  = MapboxTiles(map_id='mapbox/streets-v11',  access_token='blah')
Mapbox_satellite = MapboxTiles(map_id='mapbox/satellite-v9',  access_token='blah')


```

<details>

### Cartopy version

0.18 and current 0.19.x branch

</details>

greglucas:
If their service is returning a 410 I'm not sure there is much we can do about backwards compatibility.

dopplershift:
I agree with @greglucas

lmoresi:
The Mapbox satellite map still responds to the existing / old url pattern. The new pattern works for both. I am not sure about how the cartopy project would normally deal with a creeping incompatibility of this kind. 

This illustrates the change in the url pattern and the id better matches what Mapbox now describes as their identifier (no longer guaranteed to begin `mapbox.`

```
# OLD code
    
#     def _image_url(self, tile):
#         x, y, z = tile
[...]

PR #1763, opened 2021-04-15
Mapbox api changes

lmoresi:
## Rationale

See the discussion on issue #1761. 

The mapbox API has changed and breaks the current MapboxTiles class. The old interface remains active for the mapbox satellite map style but not for the others. The current Mapbox API documentation guides users to adopt the new approach for all the different styles.

The fix is relatively straightforward - update the url - but note that the previous assumption that all maps were described by `mapbox.map_id` is no longer valid. The styles are now `mapbox/name-version` but I am not sure if that prefix change is robust or long lasting and you may want to specify a pattern more like the original one (i.e. add the mapbox/ to the map id automagically)

## Implications

This is not backward compatible because the API is (selectively) deprecated and returns a 410 error for old-style urls.

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -450,8 +450,6 @@ class MapboxTiles(GoogleWTS):
         return url
 
 
-
-
 class MapboxStyleTiles(GoogleWTS):
     """
     Implement web tile retrieval from a user-defined Mapbox style. For more
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -421,8 +421,12 @@ class MapboxTiles(GoogleWTS):
             A valid Mapbox API access token.
         map_id
             An ID for a publicly accessible map (provided by Mapbox).
-            This is the map whose tiles will be retrieved through this process.
-
+            This is the map whose tiles will be retrieved through this process and is specified
+            through the Mapbox Styles API (https://docs.mapbox.com/api/maps/styles/)
+            Examples:
+              map_id='mapbox/streets-v11' 
+              map_id='mapbox/outdoors-v11' 
+              map_id='mapbox/satellite-v9'
         """
         self.access_token = access_token
         self.map_id = map_id
@@ -430,11 +434,20 @@ class MapboxTiles(GoogleWTS):
 
     def _image_url(self, tile):
         x, y, z = tile
-        url = ('https://api.mapbox.com/v4/mapbox.{id}/{z}/{x}/{y}.png'
+
+        # Previous version of the API (particularly note mapbox.{id} is replaced by mapbox/id)
+        # url = ('https://api.mapbox.com/v4/mapbox.{id}/{z}/{x}/{y}.png'
+        #        '?access_token={token}'.format(z=z, y=y, x=x,
+        #                                       id=self.map_id,
+        #                                       token=self.access_token))
+
+        url = ('https://api.mapbox.com/styles/v1/{id}/tiles/{z}/{x}/{y}'
                '?access_token={token}'.format(z=z, y=y, x=x,
                                               id=self.map_id,
                                               token=self.access_token))
         return url
+
+
 
 
 class MapboxStyleTiles(GoogleWTS):
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -421,11 +421,12 @@ class MapboxTiles(GoogleWTS):
             A valid Mapbox API access token.
         map_id
             An ID for a publicly accessible map (provided by Mapbox).
-            This is the map whose tiles will be retrieved through this process and is specified
-            through the Mapbox Styles API (https://docs.mapbox.com/api/maps/styles/)
+            This is the map whose tiles will be retrieved through this process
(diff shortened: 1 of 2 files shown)


lmoresi (commented):
I have responded to the sticklerbot requests

lib/cartopy/io/img_tiles.py
```diff
@@ -421,22 +421,37 @@ def __init__(self, access_token, map_id, cache=False):
...
         """
         self.access_token = access_token
         self.map_id = map_id
         super().__init__(cache=cache)
 
     def _image_url(self, tile):
         x, y, z = tile
-        url = ('https://api.mapbox.com/v4/mapbox.{id}/{z}/{x}/{y}.png'
+
+        # Previous version of the API
```
smartlixx, line 439:
```suggestion
        # Previous version of the API
```

lib/cartopy/io/img_tiles.py
```diff
@@ -421,22 +421,37 @@ def __init__(self, access_token, map_id, cache=False):
             A valid Mapbox API access token.
         map_id
             An ID for a publicly accessible map (provided by Mapbox).
-            This is the map whose tiles will be retrieved through this process.
-
+            This is the map whose tiles will be retrieved through this process
+            and is specified through the Mapbox Styles API
+            (https://docs.mapbox.com/api/maps/styles/)
+            Examples:
+              map_id='mapbox/streets-v11'
```
greglucas, line 428: Is there a reason to keep the `mapbox/` in the id, versus just making this `streets-v11`? It looks like all Mapbox Styles have the prefix `mapbox/` so you could put that into the base string.

lib/cartopy/io/img_tiles.py
```diff
@@ -421,22 +421,37 @@ def __init__(self, access_token, map_id, cache=False):
...
+        #                                       token=self.access_token))
+
+        url = ('https://api.mapbox.com/styles/v1/{id}/tiles/{z}/{x}/{y}'
                '?access_token={token}'.format(z=z, y=y, x=x,
                                               id=self.map_id,
                                               token=self.access_token))
         return url
 
 
+
```
greglucas, line 453: I don't know why Flake8 didn't catch this, but I don't think these should be there. Only two line breaks between classes.

lib/cartopy/tests/test_img_tiles.py
```diff
@@ -210,8 +210,9 @@ def test_mapbox_tiles_api_url():
     token = 'foo'
     map_name = 'bar'
     tile = [0, 1, 2]
-    exp_url = ('https://api.mapbox.com/v4/mapbox.bar'
-               '/2/0/1.png?access_token=foo')
+    exp_url = ('https://api.mapbox.com/styles/v1/bar/tiles'
+               '/2/0/1?access_token=foo')
+
```
greglucas, line 215: Remove this blank line too, should only be one when inside a function.

greglucas (commented):
Looks good to me, I think this is worth the Cartopy API break since it is keeping up with an external API change that we don't have control over.

Can you rebase onto master again to get the latest updates to pass CI?

lib/cartopy/io/img_tiles.py
lmoresi: Sorry to be slow responding here (teaching !).  I left it this way because it matched the documentation that is provided by mapbox and so is a copy/paste from their list (and I also wondered if they were reserving the option to have other urls by keeping it this way). But whatever the community wants and is used to, that's fine with me.

lmoresi:
@greglucas - I may need some help here. I did the rebase and the fixes which are now sitting in my fork but I can't work out how to update those changes into this PR !

lib/cartopy/io/img_tiles.py
greglucas: No worries at all!

Their documentation is a little bit confusing to parse, but I _think_ the `mapbox/` section is the `{username}` portion, so they are giving examples with that as the username and we should really just hardcode that on our end?

I played around with this to make an API call: https://docs.mapbox.com/playground/static/

and on page load I get:
`https://api.mapbox.com/styles/v1/{username}/{style_id}/static`
and moving around I get:
`https://api.mapbox.com/styles/v1/mapbox/streets-v11/static`

greglucas (commented):
It looks like you're close! You've just got duplicate commits in there. That usually happens when you rebase to a point a little bit too far back in the history. Make sure you rebase onto the latest `upstream/master` point and **not** something like `HEAD~3`

```
git fetch upstream
git checkout mapbox-api-changes
git rebase -i upstream/master
# Force push if the log/history looks good to you locally
# git push --force
```

lmoresi:
@greglucas - I still seem to have duplicate commits which, I think, were from github interacting with my command line git. I'm not sure - should I just start again with the PR or do you want to fix so this history is preserved !!?

greglucas:
I rebased the commits and pushed to your branch to fix up the history a bit. I think you were close, but it looked like you commented out a commit line during your rebase rather than changing the `pick` into a `drop` or removing the line. Let me know if this looks good to you now or not.

lmoresi:
Looks good - thanks for the help !

greglucas:
Thank you for the contribution, @lmoresi!

PR #1783, opened 2021-05-09
MAINT: Changing PlateCarree projection to be more lat/lon like

### LINKED ISSUE #1644: Ensure consistent globe radius (PlateCarree)
### Description

With PROJ 6+ using the new API we see this error with PlateCaree:
```
Error creating Transformer from CRS.: (Internal Proj Error: proj_create_operations: Source and target ellipsoid do not belong to the same celestial body)
```
More information about this issue:

- https://lists.osgeo.org/pipermail/proj/2019-March/008390.html
- https://lists.osgeo.org/pipermail/proj/2019-May/008586.html

Related: #709

This is a blocker for #1140 and #1477

### TODO:
Ref: https://github.com/SciTools/cartopy/pull/1252#issuecomment-508869567

 - [ ]  Make a new lat/lon projection that takes degrees
 - [ ]  Add an EquidistantCylindrical projection that works like everything else with regards to globes
 - [ ]  Deprecate PlateCarree, or at least its current behavior

greglucas:
This allows the semimajor axis to be WGS84 (or any other globe radius) and still converts the units back and forth to degrees. Another option is to subclass `PlateCarree` into a new `LonLat(PlateCarree)` class with these conversions as well.

Fixes #1644, @snowman2 can you check that this actually would help you with your branch updating Cartopy to use pyproj?


snowman2:
Nice 👍. I will try this out when I get a free moment.

greglucas:
The modern proj header yells about the transforms not belonging to the same celestial body, so this seems like it may help getting the other potential solutions across the line.

greglucas:
Superseded by #1808

PR #1787, opened 2021-05-12
Improved setup.py.

hhsprings:
## Rationale

Two fixes, one for the problem when the path contains spaces and one for vcpkg where the proj executable is missing.

I signed the SciTools CLA after making the pull request.


lgolston:
With the v0.22 release, I think this PR can be closed now as setup.py has substantially been simplified to where this is not applicable anymore.

greglucas:
agreed, thanks @lgolston for following up on this. @hhsprings if there are other things that would help with setup and installation with the new pyproject.toml we'd be happy for new PRs.

PR #1790, opened 2021-05-17
change LineString facecolor to None

### LINKED ISSUE #1782: Default styler fills LineStrings
### Description

Default styler fills LineStrings. IMO this behavior is unexpected; I expect LineStrings to be plotted as line strings. The workaround is to set `edgecolor` and `facecolor`.

#### Code to reproduce

```python
from shapely.geometry import LineString
import cartopy.crs as ccrs


geoms = [LineString([(-40, 10), (-30, 15), (-30, 20)]),
         LineString([(-40, -10), (-30, -15), (-30, -20)])]

robin = ccrs.Robinson()

fig = plt.figure(figsize=(10, 7))
ax = fig.add_subplot(1, 1, 1, projection=robin)
ax.add_geometries(geoms, crs=ccrs.PlateCarree())
ax.coastlines()
plt.show()
```

t

smartlixx:
## Rationale

close #1782  by changing LineString geometries facecolor to None.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -758,13 +758,11 @@ class GeoAxes(matplotlib.axes.Axes):
             styler_kw = styler
 
             def styler(geom):
-                styler_g = styler_kw if styler_kw else dict()
-                styler_g = styler_g.copy()
+                styler_g = styler_kw.copy() if styler_kw else {}
 
                 if isinstance(geom, sgeom.LineString):
                     styler_g['facecolor'] = 'none'
-                    if ('edgecolor' not in kwargs) or (
-                            kwargs['edgecolor'] == 'face'):
+                    if kwargs.get('edgecolor', 'face') == 'face':
                         styler_g['edgecolor'] = mpl.rcParams['patch.edgecolor']
                     else:
                         styler_g['edgecolor'] = kwargs['edgecolor']
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -754,7 +754,7 @@ class GeoAxes(matplotlib.axes.Axes):
         """
         styler = kwargs.pop('styler', None)
 
-        if (not callable(styler)) and ('facecolor' not in kwargs):
+        if not callable(styler) and 'facecolor' not in kwargs:
             styler_kw = styler
 
             def styler(geom):
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -753,6 +753,20 @@ class GeoAxes(matplotlib.axes.Axes):
 
         """
         styler = kwargs.pop('styler', None)
+        has_LineString = False
+        try:
+            for g in iter(tuple(geoms)):
+                if isinstance(g, sgeom.LineString):
+                    has_LineString = True
+                    break
+        except TypeError:
+            if isinstance(geoms, sgeom.LineString):
+                has_LineString = True
+
+        if has_LineString:
+            kwargs['facecolor'] = 'none'
+            if 'edgecolor' not in kwargs or kwargs['edgecolor'] == 'face':
+                kwargs['edgecolor'] = mpl.rcParams['patch.edgecolor']
         feature = cartopy.feature.ShapelyFeature(geoms, crs, **kwargs)
         return self.add_feature(feature, styler=styler)
 
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -753,20 +753,25 @@ class GeoAxes(matplotlib.axes.Axes):
 
         """
         styler = kwargs.pop('styler', None)
-        has_LineString = False
-        try:
-            for g in iter(tuple(geoms)):
-                if isinstance(g, sgeom.LineString):
-                    has_LineString = True
-                    break
-        except TypeError:
-            if isinstance(geoms, sgeom.LineString):
-                has_LineString = True
 
-        if has_LineString:
-            kwargs['facecolor'] = 'none'
-            if 'edgecolor' not in kwargs or kwargs['edgecolor'] == 'face':
-                kwargs['edgecolor'] = mpl.rcParams['patch.edgecolor']
+        if (not callable(styler)) and ('facecolor' not in kwargs):
+            styler_kw = styler
+
+            def styler(geom):
+                styler_g = styler_kw if styler_kw else dict()
+                styler_g = styler_g.copy()
+
+                if isinstance(geom, sgeom.LineString):
+                    styler_g['facecolor'] = 'none'
+                    if ('edgecolor' not in kwargs) or (
+                            kwargs['edgecolor'] == 'face'):
+                        styler_g['edgecolor'] = mpl.rcParams['patch.edgecolor']
+                    else:
+                        styler_g['edgecolor'] = kwargs['edgecolor']
+                    return styler_g
+                else:
+                    return styler_kw
+
         feature = cartopy.feature.ShapelyFeature(geoms, crs, **kwargs)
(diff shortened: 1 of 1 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -753,6 +753,20 @@ def add_geometries(self, geoms, crs, **kwargs):
 
         """
         styler = kwargs.pop('styler', None)
```
greglucas, line 755: What about updating the `styler` here? You can pass a callable that updates the color that way. You might be able to put the `isinstance` check in that way and it would act on each geometry then.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -753,6 +753,20 @@ def add_geometries(self, geoms, crs, **kwargs):
...
+        has_LineString = False
+        try:
+            for g in iter(tuple(geoms)):
+                if isinstance(g, sgeom.LineString):
+                    has_LineString = True
+                    break
```
greglucas, line 761: Doing the check this way, if you have one LineString anywhere you'll get the new styling on all of the other geometries too. It seems like the style update should only be applied to LineString geometries themselves, not the entire collection?

lib/cartopy/mpl/geoaxes.py
smartlixx: OK. But maybe it should be clarified that we only do styler update for LineString when 
(1) styler is not set (`None`); or
(2) `facecolor` is not set, otherwise we should assume the user wants to fill the LineString by explicitly specifying `facecolor`. 

And, what if user passes a callable to styler?

greglucas: I agree with all of that. If a user passes a callable for styler then this should have no effect. You could probably add an `if callable(styler): return styler` in there somewhere.

This should also get some tests to make sure it is covering those cases you mentioned and that we are taking the right paths.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -753,6 +753,25 @@ def add_geometries(self, geoms, crs, **kwargs):
 
         """
         styler = kwargs.pop('styler', None)
+
+        if (not callable(styler)) and ('facecolor' not in kwargs):
```
QuLogic, line 757:
```suggestion
        if not callable(styler) and 'facecolor' not in kwargs:
```

lib/cartopy/mpl/geoaxes.py
```diff
@@ -753,6 +753,25 @@ def add_geometries(self, geoms, crs, **kwargs):
...
+            styler_kw = styler
+
+            def styler(geom):
+                styler_g = styler_kw if styler_kw else dict()
+                styler_g = styler_g.copy()
```
QuLogic, line 762:
```suggestion
                styler_g = styler_kw.copy() if styler_kw else {}
```

lib/cartopy/mpl/geoaxes.py
```diff
@@ -753,6 +753,25 @@ def add_geometries(self, geoms, crs, **kwargs):
...
+
+                if isinstance(geom, sgeom.LineString):
+                    styler_g['facecolor'] = 'none'
+                    if ('edgecolor' not in kwargs) or (
+                            kwargs['edgecolor'] == 'face'):
```
QuLogic, line 767:
```suggestion
                    if kwargs.get('edgecolor', 'face') == 'face':
```
But why would `'face'` mean patch edgecolor?

smartlixx: Because `edgecolor = face` means using `facecolor` value for edge. This suggests the user specifies facecolor, and would like to fill the LineString.

greglucas: I'm almost tempted to raise here. Why are we allowing a linestring to be filled at all?

greglucas (commented):
It has been a while since I looked at this, but it looks like it still needs tests.

smartlixx:
> It has been a while since I looked at this, but it looks like it still needs tests.

Before I move on, I have to address the current test error, which I have no clue how to address. Any help would be appreciated.

```
================================== FAILURES ===================================
_________________ Test_Axes_add_geometries.test_styler_kwarg __________________
self = <cartopy.tests.mpl.test_axes.Test_Axes_add_geometries object at 0x0000005D80019910>
ShapelyFeature = <MagicMock name='ShapelyFeature' id='401595362560'>
add_feature_method = <MagicMock name='add_feature' id='401595271776'>
    @mock.patch('cartopy.mpl.geoaxes.GeoAxes.add_feature')
    @mock.patch('cartopy.feature.ShapelyFeature')
    def test_styler_kwarg(self, ShapelyFeature, add_feature_method):
        ax = GeoAxes(plt.figure(), [0, 0, 1, 1],
                     map_projection=ccrs.Robinson())
        ax.add_geometries(mock.sentinel.geometries, mock.sentinel.crs,
                          styler=mock.sentinel.styler, wibble='wobble')
    
        ShapelyFeature.assert_called_once_with(
            mock.sentinel.geometries, mock.sentinel.crs, wibble='wobble')
    
>       add_feature_method.assert_called_once_with(
            ShapelyFeature(), styler=mock.sentinel.styler)
C:\Miniconda3-x64\envs\test-environment\lib\site-packages\cartopy\tests\mpl\test_axes.py:108: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\Miniconda3-x64\envs\test-environment\lib\unittest\mock.py:919: in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <MagicMock name='add_feature' id='401595271776'>
args = (<MagicMock name='ShapelyFeature()' id='401595547712'>,)
kwargs = {'styler': sentinel.styler}
expected = call(<MagicMock name='ShapelyFeature()' id='401595547712'>, styler=sentinel.styler)
actual = call(<MagicMock name='ShapelyFeature()' id='401595547712'>, styler=<function GeoAxes.add_geometries.<locals>.styler at 0x0000005D80F8C790>)
_error_message = <function NonCallableMock.assert_called_with.<locals>._error_message at 0x0000005D80F7B790>
cause = None
    def assert_called_with(self, /, *args, **kwargs):
        """assert that the last call was made with the specified arguments.
    
        Raises an AssertionError if the args and keyword args passed in are
        different to the last call to the mock."""
        if self.call_args is None:
            expected = self._format_mock_call_signature(args, kwargs)
            actual = 'not called.'
            error_message = ('expected call not found.\nExpected: %s\nActual: %s'
                    % (expected, actual))
            raise AssertionError(error_message)
    
        def _error_message():
            msg = self._format_mock_failure_message(args, kwargs)
            return msg
        expected = self._call_matcher(_Call((args, kwargs), two=True))
        actual = self._call_matcher(self.call_args)
        if actual != expected:
            cause = expected if isinstance(expected, Exception) else None
>           raise AssertionError(_error_message()) from cause
E           AssertionError: expected call not found.
E           Expected: add_feature(<MagicMock name='ShapelyFeature()' id='401595547712'>, styler=sentinel.styler)
E           Actual: add_feature(<MagicMock name='ShapelyFeature()' id='401595547712'>, styler=<function GeoAxes.add_geometries.<locals>.styler at 0x0000005D80F8C790>)
```

greglucas:
I think the problem is that the sentinel isn't callable, so then we go into your if clause and build up the dictionary instead. If you are able to make new tests that test all of the various paths, then you can probably replace this test, or you could potentially change it to a `Mock()` object that is callable...

```python
>>> import unittest.mock
>>> callable(unittest.mock.sentinel)
False
>>> callable(unittest.mock.Mock())
True
```

PR #1812, opened 2021-07-11
make threshold writable for RotatedPole

blazing216:
## Rationale

Make `threshold` writable for the projection `RotatedPole`, which makes it easier to adjust the resolution of the great circle path. Related to Issue #8.

## Implications

`threshold` of projection `RotatedPole` can be changed after its initialisation using
```python
import cartopy.crs as ccrs
prj = ccrs.RotatedPole(180, 37)
# this line raises AttributeError: can't set attribute before this change
prj.threshold = 0.01 
```
Similar changes can be made to other projections as well if this one is approved.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1376,11 +1376,16 @@ class RotatedPole(_CylindricalProjection):
                         ('o_lat_p', pole_latitude),
                         ('lon_0', 180 + pole_longitude),
                         ('to_meter', math.radians(1))]
+        self._threshold = 0.5
         super().__init__(proj4_params, 180, 90, globe=globe)
 
     @property
     def threshold(self):
-        return 0.5
+        return self._threshold
+
+    @threshold.setter
+    def threshold(self, t):
+        self._threshold = t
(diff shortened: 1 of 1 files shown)


blazing216:
I have now signed the SciTools Contributor License Agreement

greglucas:
Is there a reason you chose to only do RotatedPole? It seems like this would be good to add to the Projection superclass so all Projections can modify the threshold.

blazing216:
> Is there a reason you chose to only do RotatedPole? It seems like this would be good to add to the Projection superclass so all Projections can modify the threshold.

Not really. I just randomly pick one projection to see if you guys are fine with this kind of change. I can try to add to the Projection superclass.

blazing216:
I made a new pull request (#1815) to replace this one.

dopplershift:
For future reference, you can always just make the changes and push them to the same branch to update the PR.

blazing216:
> For future reference, you can always just make the changes and push them to the same branch to update the PR.

Thanks @dopplershift!

PR #1814, opened 2021-07-14
DOC: Remove Iris and NetCDF dependencies from docs

### LINKED ISSUE #1813: Remove doc dependency on Iris
### Description

Currently, Iris is required to build the docs and that means that Python in the CI can't use the latest version. It is only used in the "Advanced Plotting" examples, and there is really nothing special there that isn't better understood by going to the Iris docs themselves.
https://scitools.org.uk/cartopy/docs/v0.13/matplotlib/advanced_plotting.html

I suggest removing that example. This could be done in #1652, or as a separate PR.

greglucas:
Closes #1813

This removes the Iris and NetCDF dependencies from the doc builds. The Iris example is removed and the NetCDF example is changed to use Scipy instead.

This page is already hard to find on the normal docs and these dependencies don't add useful information to our docs.
https://scitools.org.uk/cartopy/docs/v0.19/matplotlib/advanced_plotting.html

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,7 +1,3 @@
 sphinx
-netCDF4
-iris
-iris-sample-data
-cf_units
 sphinx-gallery
 beautifulsoup4
--- a/docs/source/matplotlib/advanced_plotting.rst
+++ b/docs/source/matplotlib/advanced_plotting.rst
@@ -2,13 +2,7 @@ More advanced mapping with cartopy and matplotlib
 =================================================
 
 From the outset, cartopy's purpose has been to simplify and improve the quality of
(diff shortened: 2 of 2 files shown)


greglucas:
New rendered page: https://2690-5282596-gh.circle-artifacts.com/0/docs/build/html/matplotlib/advanced_plotting.html

dopplershift (approved):

dopplershift:
Any idea if the mac 3.6 failure is related?

greglucas:
That failure is separate. Seems like the tolerance barely increased over the limits. The fix for that is over in the pyproj PR: https://github.com/SciTools/cartopy/pull/1808/commits/f010c05513f131307ee95217f3f43e9bb792dad5

PR #1815, opened 2021-07-14
Writable threshold to projection superclass

blazing216:
## Rationale

Plotting smooth great circle paths may require reducing the `threshold` of projections. `threshold` is readable only and mostly hard-coded.

Writable `threshold` makes changing the `threshold` easier. Related to #8.

## Implications

A new class member `_threshold` is added to `Projection`.

Finding a suitable `threshold` may be confusing to some users.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -99,7 +99,7 @@ class Projection(CRS, metaclass=ABCMeta):
 
     @property
     def threshold(self):
-        return self._threshold
+        return getattr(self, '_threshold', 0.5)
 
     @threshold.setter
     def threshold(self, t):
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -99,12 +99,7 @@ class Projection(CRS, metaclass=ABCMeta):
 
     @property
     def threshold(self):
-        try:
-            t = self._threshold
-        except AttributeError:
-            t = 0.5
-            self._threshold = t
-        return t
+        return self._threshold
 
     @threshold.setter
     def threshold(self, t):
@@ -846,6 +841,8 @@ class TransverseMercator(Projection):
                 proj4_params += [('approx', None)]
         super().__init__(proj4_params, globe=globe)
 
+        self._threshold = 1e4
+
     @property
     def boundary(self):
         x0, x1 = self.x_limits
@@ -947,6 +944,7 @@ class UTM(Projection):
         if southern_hemisphere:
             proj4_params.append(('south', None))
         super().__init__(proj4_params, globe=globe)
+        self._threshold = 1e2
 
     @property
     def boundary(self):
@@ -1207,6 +1205,8 @@ class LambertConformal(Projection):
         maxs = np.max(points, axis=0)
         self._x_limits = mins[0], maxs[0]
         self._y_limits = mins[1], maxs[1]
+
+        self._threshold = 1e5
 
     def __eq__(self, other):
         res = super().__eq__(other)
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -707,7 +707,7 @@ class PlateCarree(_CylindricalProjection):
         x_max = a_rad * 180
         y_max = a_rad * 90
         # Set the threshold around 0.5 if the x max is 180.
-        self._threshold = x_max / 360
+        self.threshold = x_max / 360
         super().__init__(proj4_params, x_max, y_max, globe=globe)
 
     def _bbox_and_offset(self, other_plate_carree):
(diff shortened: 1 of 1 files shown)


lib/cartopy/crs.py
```diff
@@ -1235,10 +1224,6 @@ def __hash__(self):
     def boundary(self):
         return self._boundary
 
-    @property
-    def threshold(self):
```
greglucas, line 1239: Move this one up to the `__init__` too?

lib/cartopy/crs.py
```diff
@@ -955,10 +956,6 @@ def boundary(self):
                                  (x1, y1), (x1, y0),
                                  (x0, y0)])
 
-    @property
-    def threshold(self):
```
greglucas, line 959: To the `__init__`?

lib/cartopy/crs.py
```diff
@@ -841,10 +846,6 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
                 proj4_params += [('approx', None)]
         super().__init__(proj4_params, globe=globe)
 
-    @property
-    def threshold(self):
-        return 1e4
```
greglucas, line 846: To the `__init__`?

lib/cartopy/crs.py
```diff
@@ -101,6 +97,19 @@ def x_limits(self):
     def y_limits(self):
         pass
 
+    @property
+    def threshold(self):
+        try:
```
greglucas, line 102: I'm a bit confused why you don't define `self._threshold` above? Then you shouldn't need the try/except.

blazing216: > I'm a bit confused why you don't define `self._threshold` above? Then you shouldn't need the try/except.

@greglucas Thank you for taking the time for reviewing my pull request! What do you mean exactly by 'above'? There seems no `__init__` method in `Projection` where I could define `self._threshold`.

greglucas: Can you add that method? Or perhaps a class-level attribute?

dopplershift: Or you could replace this entire method with:
```python
@property
def threshold(self):
    return getattr('_threshold', 0.5)
```

dopplershift (commented):
Really, we could avoid having this as a property at all and just directly us a `.threshold` attribute. `@property` exists so that we can later convert that simple attribute to something that can be dynamically computed without breaking users of the attribute API.

dopplershift:
Actually, based on #697, it might make sense to go ahead with property-based implementation. There seems to be a case that making `threshold` dynamic based on limits might be a decent win.

lib/cartopy/crs.py
blazing216: Using the `getattr` method sounds good.

greglucas (approved):
Looks good to me, my only minor comment would be that the subclasses could be `self.threshold = value`, instead of setting the private attribute directly (remove the underscore). That would enable consistency in case anything else gets added in the super-class' `threshold()` or someone renames the private variable that stores it.

dopplershift:
I agree with @greglucas, unless there's a reason to use the private API, all of those should use the public (i.e. `.threshold`) API.

blazing216:
I have changed the `self._threshold` in the subclasses to `self.threshold`.

lib/cartopy/crs.py
```diff
@@ -1056,7 +1051,7 @@ def __init__(self, central_longitude=0.0,
                                        np.array([min_latitude, max_latitude]))
         self._x_limits = tuple(limits[..., 0])
         self._y_limits = tuple(limits[..., 1])
-        self._threshold = min(np.diff(self.x_limits)[0] / 720,
+        self.threshold = min(np.diff(self.x_limits)[0] / 720,
                               np.diff(self.y_limits)[0] / 360)
```
QuLogic, line 1055: Re-align.

lib/cartopy/crs.py
```diff
@@ -1329,13 +1310,10 @@ def __init__(self, central_longitude=0.0, globe=None):
         proj4_params = [('proj', 'mill'), ('lon_0', central_longitude)]
         # See Snyder, 1987. Eqs (11-1) and (11-2) substituting maximums of
         # (lambda-lambda0)=180 and phi=90 to get limits.
+        self.threshold = 0.5
```
QuLogic, line 1313: That's the default now, so can be removed?

greglucas:
@blazing216, looks like there are a few minor comments to address still, but otherwise I think this is good to go if you've still got time to update it.

blazing216:
@greglucas I am sorry, Greg, I was doing some paperwork for my new job. I will try to address the comments today (Irish time)

lib/cartopy/crs.py
blazing216: Removed

lib/cartopy/crs.py
blazing216: Aligned. Thanks!

greglucas:
No problem at all, @blazing216, thanks for addressing that so quickly! Thanks for the contribution!

Test failures are HTTP request failures unrelated to this PR.

PR #1818, opened 2021-07-19
available c, ls, lw in gridliner

muchojp:
## Rationale

I noticed that ax.gridlines `lw` option didn't work. In order to check the reason, I explore the source code and found it. 
If this is remained, beginners will be confused it with such a small thing. So I think I should change it a little.

## Implications

Nothing is remained without being able to use c, ls, lw options in gridliner.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -663,10 +663,15 @@ class Gridliner:
             collection_kwargs = {}
         collection_kwargs = collection_kwargs.copy()
         collection_kwargs['transform'] = transform
-        # XXX doesn't gracefully handle lw vs linewidth aliases...
-        collection_kwargs.setdefault('color', rc_params['grid.color'])
-        collection_kwargs.setdefault('linestyle', rc_params['grid.linestyle'])
-        collection_kwargs.setdefault('linewidth', rc_params['grid.linewidth'])
+        if not any(x in collection_kwargs.keys() for x in ['c', 'color']):
+            collection_kwargs.setdefault('color',
+                                         rc_params['grid.color'])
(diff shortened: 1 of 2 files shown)


muchojp:
In the CI test, the build for Ubuntu and macOS 3.9 has passed. 
However, the build for macOS 3.6 failed and I can't resolve it. 
In other pullrequests, this phenomenon happens. 
Can I remain it in this pullrequest ?

muchojp:
For some reason, plotting using GOOGLE is a little bit changed compared to /work1/nakamura_yuki/modules/cartopy/lib/cartopy/tests/mpl/baseline_images/mpl/test_images/web_tiles.png in the macos 3.6 build.
However, the difference is not essential.
We can increase the tolerance from 5.8 to 5.91, I think.

muchojp:
The differences. 
web_tiles
https://github.com/muchojp/cartopy/suites/3279494852/artifacts/76924227

lib/cartopy/mpl/gridliner.py
```diff
@@ -664,9 +664,15 @@ def _draw_gridliner(self, nx=None, ny=None, renderer=None):
         collection_kwargs = collection_kwargs.copy()
         collection_kwargs['transform'] = transform
         # XXX doesn't gracefully handle lw vs linewidth aliases...
```
greglucas, line 666: You do gracefully handle the aliases now, so the comment can be removed :)

greglucas (approved):
Looks like a good solution. Can you squash all of your commits into one descriptive commit?

muchojp:
I stashed my commits. Can you check it?

greglucas:
Looks good to me. Thanks @muchojp!

Issue #1659, opened 2020-09-16
Shapefile gives warning with GSHHS feature


MHBalsmeier:
Hi,
shapefile gives me a warning when I save a figure which has GSHHS features in it. Thought I'd ask it here:

MWE:

```python
import matplotlib.pyplot as plt;
import cartopy.crs as ccrs;
import cartopy;
fig = plt.figure();
proj = ccrs.Mercator(central_longitude = 0);
ax = plt.axes(projection = proj);
ax.add_feature(cartopy.feature.GSHHSFeature(levels = (1, 2), linewidth = 2.0));
print("before");
fig.savefig("test.png");
print("after");
```
output:

before
```
/home/max/.local/lib/python3.8/site-packages/shapefile.py:385: UserWarning: Shapefile shape has invalid polygon: no exterior rings found (must have clockwise orientation); interpreting holes as exteriors.
  warnings.warn('Shapefile shape has invalid polygon: no exterior rings found (must have clockwise orientation); interpreting holes as exteriors.')
```
after

What is the cause of this and how can it be resolved?

OS: Ubuntu 20.04
Cartopy: 0.18.0
pyshp: 2.1.2

Thanks in advance.

dopplershift:
Since the warning is coming from `/home/max/.local/lib/python3.8/site-packages/shapefile.py:385`, I think this is an issue between the GSHHS data and pyshp. It may be worth asking on the [pyshp issue tracker](https://github.com/GeospatialPython/pyshp/issues). Having said that, it looks like we are currently using an old copy of the GSHHS data (2.2.0 while 2.3.7 is the latest), so it's possible using a newer copy of the data would fix it--which is covered in #1353. Closing since the only thing CartoPy could do is update, and that's already covered in another issue.

MHBalsmeier:
thanks, updating the coastline data solved it

karimbahgat:
In short, the warnings are nothing to worry about. It's just a notification caused when trying to produce the GeoJSON representation of two weird slivers in the `GSHHS_c_L1.shp` shapefile (downloaded from the most recent cartopy GitHub repo), at shapefile index 95 and 471. They are both single-part polygons but have the ring orientation of an interior hole (i.e. holes without an exterior polygon). PyShp assumes they were meant to be exterior polygons and includes them in the GeoJSON, and notifies the user. I just made some commits that improved the somewhat cryptic warning messages, reports which shapes were affected, and allows users to disable the warning messages. Should be available in the next version.

Issue #1353, opened 2019-08-12
GSHHS not updated


enorazza:
I am using cartopy.io.shapereader.gshhs.
Noticed this output while processing:
```
Downloading: https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhs/oldversions
/version2.2.0/GSHHS_shp_2.2.0.zip
```

Version 2.2.0 is old (2013), recent version is 2.3.7 (June 15, 2017), available here:
https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhg/latest/

Can be updated to a more recent version if not the latest?

dopplershift:
I'm not sure what the ramifications of updating are or what it would take to do so, but I'd be open to a PR that does it.

greglucas:
This was updated in #1823

PR #1823, opened 2021-07-28
Coastlines update

MHBalsmeier:
## Rationale

The GSHHS coastlines that cartopy was automatically downloading were quite old, lots of newer versions came out. Several issues were raised about this (#1353, #1659). I updated the coastlines from 2.2.0 to 2.3.6.

## Implications

More up-to-date and better coastlines are used that include bugfixes and resolve some shape warnings that were frequently thrown.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -31,6 +31,7 @@ import glob
 import io
 import itertools
 import os
+from urllib.error import HTTPError
 
 import shapely.geometry as sgeom
 import shapefile
@@ -408,9 +409,12 @@ class GSHHSShpDownloader(Downloader):
     """
     FORMAT_KEYS = ('config', 'scale', 'level')
 
-    _GSHHS_URL_TEMPLATE = ('https://www.ngdc.noaa.gov/mgg/shorelines/data/'
-                           'gshhs/oldversions/version2.2.0/'
-                           'GSHHS_shp_2.2.0.zip')
+    gshhs_version = '2.3.7'
+
+    _GSHHS_URL_TEMPLATE = (
+        'https://www.ngdc.noaa.gov/mgg/shorelines/data/'
+        'gshhs/latest/gshhg-shp-' + gshhs_version + '.zip'
+    )
 
     def __init__(self,
                  url_template=_GSHHS_URL_TEMPLATE,
@@ -435,15 +439,43 @@ class GSHHSShpDownloader(Downloader):
 
         # Download archive.
         url = self.url(format_dict)
-        shapefile_online = self._urlopen(url)
+        try:
+            shapefile_online = self._urlopen(url)
+        # error handling:
+        except HTTPError:
+            try:
+                """
+                case if GSHHS has had an update
+                without changing the naming convention
+                """
+                url = (
+                    'https://www.ngdc.noaa.gov/mgg/shorelines/data/'
+                    'gshhs/oldversions/version' + self.gshhs_version + '/'
+                    'gshhg-shp-' + self.gshhs_version + '.zip'
+                )
+                shapefile_online = self._urlopen(url)
+            except HTTPError:
+                """
+                case if GSHHS has had an update
+                with changing the naming convention
+                """
+                url = (
+                    'https://www.ngdc.noaa.gov/mgg/shorelines/data/'
+                    'gshhs/oldversions/version2.3.6/'
+                    'gshhg-shp-2.3.6.zip'
+                )
+                shapefile_online = self._urlopen(url)
         zfh = ZipFile(io.BytesIO(shapefile_online.read()), 'r')
         shapefile_online.close()
 
         # Iterate through all scales and levels and extract relevant files.
         modified_format_dict = dict(format_dict)
         scales = ('c', 'l', 'i', 'h', 'f')
-        levels = (1, 2, 3, 4)
+        levels = (1, 2, 3, 4, 5, 6)
         for scale, level in itertools.product(scales, levels):
+            # the combination c4 does not occur for some reason
+            if scale == "c" and level == 4:
+                continue
             modified_format_dict.update({'scale': scale, 'level': level})
             target_path = self.target_path(modified_format_dict)
             target_dir = os.path.dirname(target_path)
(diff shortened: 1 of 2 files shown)


MHBalsmeier:
Undid 00f95f1 and used black to achieve correct code format.
I do not know why the Mac OS test fails though.
Sorry this PR look chaotic now but I had a hard time with this sticker-ci.

lib/cartopy/io/shapereader.py
```diff
@@ -442,20 +443,24 @@ def acquire_all_resources(self, format_dict):
...
-                os.makedirs(target_dir)
-
-            for member_path in self.zip_file_contents(modified_format_dict):
-                ext = os.path.splitext(member_path)[1]
-                target = os.path.splitext(target_path)[0] + ext
-                member = zfh.getinfo(member_path.replace(os.sep, '/'))
-                with open(target, 'wb') as fh:
-                    fh.write(zfh.open(member).read())
+            # the combination c4 does not occur for some reason
+            if scale != "c" or level != 4:
```
greglucas, line 449:
```suggestion
            if scale == "c" and level == 4:
                continue
```
I think it is more readable to be explicit about what you are skipping.

greglucas (commented):
I wonder if it is possible to get the `latest` version?
https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhg/latest/
That would be complicated by the version number being updated in the URL, but it would be nice to just always get the latest updates when possible.

Can you squash all of your commits? It is a bit complicated to follow when adding/removing content. You can safely ignore the Python3.6 failing test right now, it isn't related to your PR.

lib/cartopy/io/shapereader.py
MHBalsmeier: Yes thanks I did that.

MHBalsmeier:
The changing URL is indeed the reason for me not choosing the latest version. I agree this would be nice, but I cannot think of  a safe way to implement this. I could try something like `*shp*.zip`, but that is not safe either if the next version should be named differently.

I implemented your suggestion about the `c4` combination. Now I will try to squash the commits.

greglucas:
:+1: That may be overcomplicating things to parse the latest page to grab the url. I wonder if adding a version number to the arguments of the downloader would be useful though, so we could just update the version number when it is bumped? Then you could try both the latest location and old location to download that version number which would gracefully handle the move from latest to old. Something like:
```python
try:
    self._urlopen(latest_url)
except HTTPError:
    self._urlopen(oldversions_url)
```

I think what you have is a good update though, so don't feel obligated to add the suggestion, it can be handled in a follow-up PR as well.

MHBalsmeier:
Okay I will try this, let's do this in this PR.

lib/cartopy/io/shapereader.py
```diff
@@ -435,15 +438,29 @@ def acquire_all_resources(self, format_dict):
 
         # Download archive.
         url = self.url(format_dict)
-        shapefile_online = self._urlopen(url)
+        try:
+            shapefile_online = self._urlopen(url)
+        except HTTPError:
+            warnings.warn(
```
dopplershift, line 444: Does this mean that if ngdc updates to a new version, users are going to get a warning that they can't fix (without resorting to a warning filter)? I'm not a fan of issuing warnings when users are doing nothing wrong and can't fix it.

MHBalsmeier: Yes this is what it means, but ...

- it happens only once (and a Download warning also happens when the coastlines are downloaded anyway)
- it is better than having no warning but the bugs of an old coastline data set.

Alternative would be to either download the older version in the first place as before or leave the warning away.

greglucas: There is a third option, which is what I was trying to hint at, which is to put a version number in somewhere.

```python
_GSHHS_URL_TEMPLATE = f'https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhg/latest/gshhg-shp-{version}.zip'
```
Then on the except you would do:
```python
url = f'https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhs/oldversions/version{version}/gshhg-shp-{version}.zip'
```

I _assume_ that the 2.3.7 will be moved over to the archive at some point when a new one comes up, and this would catch that case assuming they keep their naming convention.

MHBalsmeier: Shouldn't we have something in the except that works even if the naming convention is changed?

greglucas: :+1: Add a fallback if both of those fail to the 2.3.6 version if both throw the URL not found error?

MHBalsmeier: Okay, here is my attempt

MHBalsmeier:
Okay, did squash everything into one again. Maybe someone can check.

greglucas (approved):
Worked for me locally! Just one minor request to update the docs.

lib/cartopy/io/shapereader.py
```diff
@@ -435,15 +439,43 @@ def acquire_all_resources(self, format_dict):
...
+                )
+                shapefile_online = self._urlopen(url)
         zfh = ZipFile(io.BytesIO(shapefile_online.read()), 'r')
         shapefile_online.close()
 
         # Iterate through all scales and levels and extract relevant files.
         modified_format_dict = dict(format_dict)
         scales = ('c', 'l', 'i', 'h', 'f')
-        levels = (1, 2, 3, 4)
+        levels = (1, 2, 3, 4, 5, 6)
```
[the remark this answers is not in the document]
MHBalsmeier: did that, will squash once more

MHBalsmeier:
sorry this review request was a mistake of mine

MHBalsmeier:
Looks like I did something wrong with the squash.

greglucas:
Keep going with this PR if you can. You're close on the rebase I think!

Try this:
```bash
git fetch upstream
git rebase -i upstream/master
```
Then delete the extra commit from your work and keep this one.

MHBalsmeier:
Okay, thanks. Let's see.

greglucas (approved):
@dopplershift, I think the warning you were concerned with is removed now. I'll leave open for a bit in case you have any comments on this.

MHBalsmeier:
> @dopplershift, I think the warning you were concerned with is removed now. I'll leave open for a bit in case you have any comments on this.

Yes I removed it, but the standard DownloadWarning still occurs as before.

greglucas:
Thanks for updating this, @MHBalsmeier!

Issue #1281, opened 2019-03-12
Configuration via environment variables


andreas-h:
### Description
I've just set up cartopy on our JupyterHub installation.  I would like to keep one central repository of all shapefiles needed / used by Cartopy.

I'm aware of the `cartopy.config` documentation, and the suggestion to put a `siteconfig.py` file into the Cartopy root directory.  This works perfectly.

However, I think it would be nice to have the option of setting these configuration variables via environment variables.  That way, I could `conda update` my environment, and wouldn't have to remember updating the `siteconfig.py` every time.

Instead, one could check in `__init__.py` if envrironment variables like `$CARTOPY_PRE_EXISTING_DATA_DIR` and the like are set, and if so, use those to overwrite the values coming from `siteconfig`.

<details>

### Operating system

Linux

### Cartopy version

0.17.0

### conda list

</details>

PR #1827, opened 2021-08-06
Allow site-wide configuration via environment variables for data directory containing standard data.

aerorahul:
## Rationale

Allow easier site-wide use of standard data (such as that from NaturalEarth).  
Addresses Issue #1281

Cartopy allows for site-wide customizations via `cartopy.config` and `siteconfig.py`, but that requires system admins to place this file into the installation directory.  Cartopy also allows for users to configuration via `cartopy_userconfig`, but most users are not savvy about such things.

## Implications

This will allow site-wide use of system installed/provided standard data as well as users to provide their own path to this data via an environment variable `CARTOPY_DATA_DIR`.

## Checklist

 * CLA -- Signed

 * Example:  
```
export CARTOPY_DATA_DIR="absolute path to a directory where standard data can be found"
```

 * Ensure there is a suitable item in the cartopy test suite for the change you are proposing -- N/A

### DIFF
--- a/lib/cartopy/__init__.py
+++ b/lib/cartopy/__init__.py
@@ -20,7 +20,7 @@ _data_dir = os.path.join(os.environ.get("XDG_DATA_HOME", _writable_dir),
                          'cartopy')
 _cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
 
-config = {'pre_existing_data_dir': '',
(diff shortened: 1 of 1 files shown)


greglucas (approved):
Seems reasonable to me and the env variable seems like a sensible name.

greglucas:
Thanks, @aerorahul!

PR #1829, opened 2021-08-20
Fix Geostationary boundary for ellipsoids

dopplershift:
This fixes the boundary for the `Geostationary` to handle arbitrary ellipsoids rather than only being correct for spheres. It actually ended up being a straightforward change--I just had to reteach myself the math I used last time. I drastically improved the naming and comments to hopefully make this easier to grok in the future if necessary. The important part is for our `eccentric_ellipse.py` example,  before it looked like:

and now:

Figure_1

You can also see a small improvement with Earth-based geostationary imagery. Before:

image

After:

image

If anyone wants to know more about parametric equations for ellipses, [here you are](https://math.stackexchange.com/questions/393215/how-to-write-a-point-on-an-ellipse-using-r-and-theta).


dopplershift:
I'll update this for when #1808 merges, I just wanted to get this posted since I cracked the problem.

dopplershift:
Ok, found one more fix for ellipsoids here (because the ellipsoid extends "elliptically" in the direction of the satellite (rather than as a circle) and thus some assumptions about tangent lines were wrong). This was only important for much smaller values of satellite height. (Like <=3500km).

QuLogic:
You can rebase now, I guess.

dopplershift:
Rebased. I also fixed the tests and restored the frame on the `eccentric_ellipse.py` example.

greglucas (approved):
Nice work on this. It looks nice and flexible now!

PR #1833, opened 2021-09-01
Update Natural Earth URL

rcomer:
This is my first Cartopy PR, so apologies if I've missed something!

## Rationale

Updated the URL for Natural Earth feature downloads, following the AWS template provided by @acarapetis at https://github.com/SciTools/cartopy/issues/1325#issuecomment-904343657.  @greglucas stated at https://github.com/SciTools/cartopy/issues/1325#issuecomment-904686853 that he thinks AWS is the official replacement for the https://naciscdn.org/ source (which has been rather intermittent in recent weeks).

## Implications

Hopefully the shapefile downloads will become a bit more reliable.

PR #1602, opened 2020-07-02
Rename feature_download.py

stefraynaud:
## Rationale

Rename ` feature_download.py`  to `cartopy_feature_download.py` and add it to setup.py to make it installable.

I prefixed it with `cartopy_` to make it clear that it is a cartopy tool.

## Implications

Once cartopy is installed, the user has a direct access to ` cartopy_feature_download.py`  to install new features, without having to search for this script.

PR #1837, opened 2021-09-03
enforce stable aws endpoint for cartopy_feature_download.py

### LINKED ISSUE #1834: Backport NaturalEarth url change in release 0.18 and 0.19 ?
It should be great, if possible, to backport the change of NaturalEarth url (done in this [commit](https://github.com/SciTools/cartopy/commit/af814c31829002acd7c1c49ba52cee530c6ac19a) ) in at least releases 0.18 (==> 0.18.1) and 0.19 (==> 0.19.1), and of course to release a new version with this change.

bjlittle:
This PR is related to #1833 and the ability to download `cartopy` feature resources from the now preferred stable AWS endpoint.

The now renamed `tools/cartopy_feature_download.py` (#1602) on `master` only requires a one line change to ensure that this script works with almost all older versions of `cartopy`, see [line +142](https://github.com/SciTools/cartopy/compare/master...bjlittle:cartopy-feature-download?expand=1#diff-b744d38ca10296c79eaa50b103cd221fb7bfc388caa7faa1a49ff08f1b3357ddR142). Thus removing the need to backport this fix. Clearly there is community appetite for this to happen given #1834.

We have a live use case on SciTools/iris, see [here](https://github.com/SciTools/iris/pull/4304/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR81), where we want the benefits of using #1833 to download and populate a `cartopy` cache for our CI. As it happens, the fix proposed here works with an installed `cartopy` v0.18. 

Initially, we worked around the issue by crafting the change solely within `iris`, but there is a bigger win for the `cartopy` community here rather than limiting the win within `iris` e.g., downloading this PR version of `cartopy_feature_download.py` will work for older immutable production versions of `cartopy` to populate the cache and avoid downloading resources on demand.

The benefits of this PR is also not dependant on https://github.com/conda-forge/cartopy-feedstock/pull/116, which only helps users of `cartopy_feature_download.py` with `cartopy` v0.19.0.post1 and onwards.

Closes #1834

### DIFF
--- a/tools/cartopy_feature_download.py
+++ b/tools/cartopy_feature_download.py
@@ -10,19 +10,23 @@ the data used by various Feature instances.
 
 For detail on how to use this tool, execute it with the `-h` option:
 
-    python download.py -h
+    python cartopy_feature_download.py -h
 
 """
 
 import argparse
+import pathlib
 
 from cartopy import config
 from cartopy.feature import Feature, GSHHSFeature, NaturalEarthFeature
-from cartopy.io import Downloader
+from cartopy.io import Downloader, DownloadWarning
 
 
 ALL_SCALES = ('110m', '50m', '10m')
 
+# See https://github.com/SciTools/cartopy/pull/1833
+URL_TEMPLATE = "https://naturalearth.s3.amazonaws.com/{resolution}_{category}/ne_{resolution}_{name}.zip"
+SHP_NE_SPEC = ("shapefiles", "natural_earth")
 
 FEATURE_DEFN_GROUPS = {
     # Only need one GSHHS resolution because they *all* get downloaded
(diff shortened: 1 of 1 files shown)


tools/cartopy_feature_download.py
```diff
@@ -114,11 +119,28 @@ def download_features(group_names, dry_run=True):
...
+    parser.add_argument('--no-warn',
+                        action='store_true',
+                        help='ignore cartopy "DownloadWarning" warnings')
     args = parser.parse_args()
 
     if args.output:
-        config['pre_existing_data_dir'] = args.output
-        config['data_dir'] = args.output
+        target_dir = pathlib.Path(args.output).expanduser().resolve()
+        target_dir.mkdir(parents=True, exist_ok=True)
```
bjlittle, line 129: The `exist_ok` flag was added in Python `3.5`, so I assume it's okay to use here...surely :stuck_out_tongue_winking_eye:

greglucas: I guess you'll just get an error if you want to try with an earlier Python version ;)

greglucas (approved):
Just tried locally with all the various options and an earlier version of Cartopy and it looks good to me. Thanks for the quick update @bjlittle!

bjlittle:
@greglucas awesome, thanks 👍

PR #1840, opened 2021-09-05
MNT: Update Feature Artist default zorder

greglucas:
This PR changes the default zorder to 1.5 for all features added to an axis. This places them above images and patches, but below lines and text.

## Rationale

Matplotlib 3.5 is changing the artist list draw order, which means patches will be drawn in the order they were added to the axis, and *not* last as we have been counting on inadvertently. Many users use `.coastlines()`, which would be drawn in a different order now and likely break a lot of workflows. This PR forces all "features" to have a default zorder of 1.5, which can be overridden when created or added to the axes.

ping @QuLogic, let me know if you have a better fix for this.

### DIFF
--- a/lib/cartopy/mpl/feature_artist.py
+++ b/lib/cartopy/mpl/feature_artist.py
@@ -112,21 +112,18 @@ class FeatureArtist(matplotlib.artist.Artist):
             color = self._kwargs.pop('color')
             self._kwargs['facecolor'] = self._kwargs['edgecolor'] = color
 
-        # Set default zorder so that features are drawn before
-        # lines e.g. contours but after images.
+        # Set default zorder so that features are drawn under
+        # lines e.g. contours but over images and filled patches.
         # Note that the zorder of Patch, PatchCollection and PathCollection
         # are all 1 by default. Assuming equal zorder drawing takes place in
-        # the following order: collections, patches, lines (default zorder=2),
-        # text (default zorder=3), then other artists e.g. FeatureArtist.
+        # the following order: collections, patches, FeatureArtist, lines,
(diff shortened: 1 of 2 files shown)


QuLogic:
I was actually aiming for 2 for lines, 1 for filled., but it's a bit complicated to determine, so this may be better. I'll open a PR for comparison's sake.

QuLogic:
See #1842 for comparison.

dopplershift (approved):
I'm ok with this approach.

lib/cartopy/mpl/feature_artist.py
```diff
@@ -112,21 +112,18 @@ def __init__(self, feature, **kwargs):
             color = self._kwargs.pop('color')
             self._kwargs['facecolor'] = self._kwargs['edgecolor'] = color
 
-        # Set default zorder so that features are drawn before
-        # lines e.g. contours but after images.
+        # Set default zorder so that features are drawn under
+        # lines e.g. contours but over images and filled patches.
         # Note that the zorder of Patch, PatchCollection and PathCollection
         # are all 1 by default. Assuming equal zorder drawing takes place in
```
QuLogic, line 118:
```suggestion
        # are all 1 by default. Assuming default zorder, drawing takes place in
```

greglucas:
I'm away from the keyboard for a few days, so I won't be able to update any of the PRs until I get back. (I tried to update this through the app, but not sure how well that works) anyone can feel free to push to my branches to update the PRs in the meantime.

Before 3.5 is released, we will also want to update the Quadmesh call in pcolormesh to remove a bunch of warnings.

PR #1842, opened 2021-09-08
Use higher default zorder for unfilled Features

QuLogic:
This matches the behaviour of contours in Matplotlib, where unfilled vs filled have different default zorder.

### DIFF
--- a/lib/cartopy/mpl/feature_artist.py
+++ b/lib/cartopy/mpl/feature_artist.py
@@ -17,6 +17,7 @@ import weakref
 import numpy as np
 import matplotlib.artist
 import matplotlib.collections
+import matplotlib.colors
 
 import cartopy.mpl.patch as cpatch
 from .style import merge as style_merge, finalize as style_finalize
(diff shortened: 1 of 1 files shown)


QuLogic:
Ah, alas, it breaks more than I expected because Matplotlib <3.5 still does the weird sorting.

greglucas (commented):
I think this may be a bit too fancy? Setting the zorder to 1.5 as a default is pretty simple and self-explanatory, and then the user can go and adjust it from there, IMO. They won't have to think about whether the Feature they added was a patch or a line.

I like this location of updating the zorder though, so I changed the other PR to update `self.set_zorder(1.5)` in the default case.

dopplershift:
Would it be better to wait for this until we drop matplotlib <3.5?

greglucas:
The reason for this change is that many of the coastlines disappear from the tests with MPL 3.5, which would likely mess up a lot of people's scripts. The change in MPL 3.5 would cause a difference in order depending on:
```python
ax.coastlines()
ax.contourf()
```
vs
```python
ax.contourf()
ax.coastlines()
```
because the default zorder of the two situations are the same. Previously, "artists" of the same zorder would still get sorted after everything else. That is no longer the case and all artists of the same zorder get sorted by when they were added to the list.

QuLogic:
This re-orders a bunch of things with older Matplotlib, so let's go with your PR instead.

PR #1843, opened 2021-09-08
Use packaging.version instead of distutils

### LINKED ISSUE #1836: DEP: distutils is deprecated
https://www.python.org/dev/peps/pep-0632/

QuLogic:
## Rationale

Makes us prepared for Python 3.12, which removes `distutils`.

Fixes #1836.

## Implications

New dependency.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,9 +21,9 @@ import os
 import sys
 
 import cartopy
-from distutils.version import LooseVersion
 import matplotlib
 import owslib
+from packaging.version import parse as parse_version
 from sphinx_gallery.sorting import ExampleTitleSortKey, ExplicitOrder
 
 # If extensions (or modules to document with autodoc) are in another directory,
@@ -84,7 +84,7 @@ release = cartopy.__version__
 
 
 if (hasattr(owslib, '__version__') and
-        LooseVersion(owslib.__version__) >= '0.19.2'):
+        parse_version(owslib.__version__) >= parse_version('0.19.2')):
     expected_failing_examples = []
 else:
     # OWSLib WMTS support is broken.
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -1,2 +1,3 @@
 flufl.lock
+packaging>=20
 pytest>=5.1.2
--- a/lib/cartopy/tests/mpl/__init__.py
+++ b/lib/cartopy/tests/mpl/__init__.py
@@ -5,7 +5,6 @@
 # licensing details.
 
 import base64
-from distutils.version import LooseVersion
 import os
 import glob
 import shutil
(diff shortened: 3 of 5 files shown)


docs/source/conf.py
```diff
@@ -84,7 +84,7 @@
 
 
 if (hasattr(owslib, '__version__') and
-        LooseVersion(owslib.__version__) >= '0.19.2'):
+        parse_version(owslib.__version__) >= parse_version('0.19.2')):
```
greglucas, line 87: This is slightly annoying that we have to wrap everything ourselves now, rather than "parse_version" being able to cast the string to `Version` internally during the comparison...

requirements/tests.txt
```diff
@@ -1,2 +1,3 @@
 flufl.lock
+packaging>=20
```
greglucas, line 2: setuptools has packaging distributed with it internally, do we want to use that instead of the new dependency? I haven't kept up with what the expected version comparison landscape is in Python lately, but a quick google seems like it is still quite dynamic.

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -126,7 +126,7 @@ def test_gridliner_specified_lines():
 
 # The tolerance on these tests are particularly high because of the high number
 # of text objects. A new testing strategy is needed for this kind of test.
-if MPL_VERSION < "3":
+if MPL_VERSION.major < 3:
```
greglucas, line 129: We have a minimum dependency of MPL 3.2, so I guess this can just be removed? There are quite a few checks in the tests that we should clean up other than this as well, so this can wait for another PR.

requirements/tests.txt
QuLogic: That's in `_vendor` and is private. `packaging` is the intended replacement for `distutils.version`.

lib/cartopy/tests/mpl/test_gridliner.py
QuLogic: Do we? The code doesn't seem to say that anywhere.

dopplershift: According to our CI config, it's matplotlib 3.1:

https://github.com/SciTools/cartopy/blob/c7af419dd452450f52e9987e00e764f6e18c8e42/.github/workflows/ci-testing.yml#L26-L33

😆 The docs currently say:
https://github.com/SciTools/cartopy/blob/c7af419dd452450f52e9987e00e764f6e18c8e42/INSTALL#L103-L105

dopplershift: Oh lovely, `requirements/plotting.txt` says 1.5.1 as well.

QuLogic: #1852

requirements/tests.txt
greglucas: Thanks for the clarification!

greglucas (approved):
This looks good to me. You have it as draft currently, feel free to self-merge if this is all you wanted to add here.

QuLogic:
Just waiting for the minimum version bumps, which will cause a lot of conflicts for things that are dropped.

greglucas (approved):

PR #1851, opened 2021-09-10
Bump minimum NumPy to 1.18

QuLogic:
## Rationale

Another step towards #1518. ~~Technically, since we were delayed for so long, we could bump minimum to 1.18, but we don't have any code that checks for anything that new, so it'd make no difference.~~

~~Maybe we should anyway though, for consistency with NEP29?~~

## Implications

Older installs would not work.


QuLogic:
Ah, I see our test minimum already did the bump to 1.18, so I'll just do that.

dopplershift (approved):

PR #1854, opened 2021-09-10
Bump minimum Proj to 8.0.0

QuLogic:
## Rationale

If I've understood #1808 correctly, we only support Proj 8+. This was changed in the documentation, but not actually in the coded requirements.

## Implications

We can also drop a bunch of conditional code for old versions.


greglucas (approved):
We are in a weird in-between state right now. pyproj has wheels that it uses for the transform calculations now. _But_, we still need the PROJ library headers for the Geodesic stuff. That header file hasn't changed since PROJ4.9, so it is technically the latest version that is required. But, the version number we are referencing is from pyproj, so all of the test checks here will never hit those paths.

I think this is OK and we live with PROJ8.0 being our minimum, it will make it much easier to explain to everyone. Hopefully, we can completely remove the PROJ requirement in a release or two.

dopplershift (approved):

Issue #1832, opened 2021-08-31
Port Geodesic routines to use pyproj


greglucas:
With #1808, we are using pyproj in trace.pyx, but it would be good to use pyproj for all PROJ interfaces. So, we will need to convert the Geodesic routines to use pyproj as well. (Some are in geodesic.pyx and others are in trace.pyx)

QuLogic:
See https://github.com/QuLogic/cartopy/commit/d109aac015d2e09e60dd9050dddeb1ed27363d1d for an attempt to convert `trace.pyx` to pyproj. 

The projection benchmarks increase by about 50% doing so (~2s to ~3s), which seems too large. I'm not sure if it's due to pyproj not supporting the `geod_line` API or because we have to flip from Cython to Python to C very often during the bisection.

greglucas:
We just took a big hit in speed in the linear case to move to pyproj. I actually think that 50% slower in the spherical interpolation is worth it if we can completely drop the PROJ installation dependency (relying on pyproj for everything). That commit looks appealing from my perspective.

dopplershift:
I'd want to know for sure the performance hit based on some real-world cases. In general, though reducing our maintenance and build overhead is worth quite a lot to me.

QuLogic:
This is based on our asv benchmarks:
![image](https://user-images.githubusercontent.com/302469/133180418-85e68ad3-e0d5-4f2d-bb0f-ebf3822ee6e3.png)
As you can see all projections are slower by roughly the same ratio. I only benchmarked the last 20 or so commits, so I don't know what effect the pyproj change had.

These tests all check `project_linear` on the `OCEAN` feature and 50m scale.

QuLogic:
You can see at https://qulogic.github.io/cartopy-bench/ in the geodesic benchmarks how the conversion to pyproj in #1857 slowed down geodesics. 

Additionally, the last commit here https://qulogic.github.io/cartopy-bench/#project_linear.Suite.time_project_linear shows what happens when `trace.pyx` is also converted to geodesics (the sixth- through second-last commit are for #805).

greglucas:
@QuLogic, I just checked out your branch and tried to profile a couple of examples that would use the geodesic routines (tissot and effects_of_the_ellipse) in a real example. These showed very minimal increases in overall runtime, maybe 10% or so (hard to tell on these fast examples though). There are a lot more calls to pyproj's geod.fwd due to the SphericalInterpolator, but even that isn't super slow.

[...]

PR #1857, opened 2021-09-11
Port Geodesic class to pyproj

QuLogic:
## Rationale

The class is quite similar to pyproj's `Geod` class and can be converted almost directly. I haven't converted `trace.pyx` yet, so it gets all the definition imports moved into it. The problem for the interpolator is that it sets up a line and repeatedly finds a point in between as it bisects, whereas pyproj only supports getting these (a single point or a bunch) in one go only.

## Implications

One less dependency on compiling against Proj. See #1832


QuLogic:
It might not be so difficult to port `trace.pyx` to pyproj as well, but I seem to have broken things in my first attempt.

dopplershift (approved):

dopplershift:
I'll leave it open for just a bit (unless you want to get the `trace.pyx` port here, but the tests pass and it seems straightforward enough.

greglucas (approved):
Looks good to me. Feel free to self-merge if you don't want to add anything else.

I don't think we can port trace.pyx over to pyproj quite yet. That will require some additions to pyproj, but we are close.

QuLogic:
It can be done, but benchmarks show that it will increase projection times by ~50%. I don't know if this is strictly because pyproj doesn't support the `geod_line` aPI or because of the increased transitions from C code back to pyproj's Python.

PR #1860, opened 2021-09-15
DOC: Remove private attributes from examples

greglucas:
TextPath vertices are changing to readonly, so we can't modify the private attribute (which we shouldn't have been doing anyways). Change to modifying the size and transforms of the paths instead.

Additionally, the old font wasn't being found in CI, so change that to the one it was falling back to.

### DIFF
--- a/examples/miscellanea/logo.py
+++ b/examples/miscellanea/logo.py
@@ -9,8 +9,8 @@ import cartopy.crs as ccrs
 import matplotlib.pyplot as plt
 import matplotlib.textpath
 import matplotlib.patches
+import matplotlib.transforms
 from matplotlib.font_manager import FontProperties
-import numpy as np
 
 
 def main():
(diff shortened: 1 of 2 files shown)


examples/miscellanea/logo.py
```diff
@@ -21,11 +21,12 @@ def main():
...
     # generate a matplotlib path representing the word "cartopy"
-    fp = FontProperties(family='Bitstream Vera Sans', weight='bold')
+    fp = FontProperties(family='DejaVu Sans', weight='bold')
     logo_path = matplotlib.textpath.TextPath((-175, -35), 'cartopy',
-                                             size=1, prop=fp)
+                                             size=80, prop=fp)
     # scale the letters up to sensible longitude and latitude sizes
-    logo_path._vertices *= np.array([80, 160])
+    transform = matplotlib.transforms.Affine2D().scale(1, 2).translate(0, 35)
+    logo_path = logo_path.transformed(transform)
```
QuLogic, line 29: Can you pass this as `transform=transform + ccrs.PlateCarree()` in `PathPatch` instead of transforming it here?

QuLogic:
Bitstream Vera Sans was the old (pre 2.0.0) Matplotlib default font.

examples/miscellanea/logo.py
greglucas: Sort-of, we can't add the ccrs.PlateCarree() directly, but have to add the MPL version of it. I also changed the image clip-path to just use the patch instead of a separate path as well.

examples/miscellanea/logo.py
```diff
@@ -21,24 +21,25 @@ def main():
...
+                                             size=80, prop=fp)
     # scale the letters up to sensible longitude and latitude sizes
-    logo_path._vertices *= np.array([80, 160])
+    transform = matplotlib.transforms.Affine2D().scale(1, 2).translate(0, 35)
 
     # add a background image
     im = ax.stock_img()
     # clip the image according to the logo_path. mpl v1.2.0 does not support
     # the transform API that cartopy makes use of, so we have to convert the
     # projection into a transform manually
```
QuLogic, line 34: We can drop this comment, which was about the clip path.

examples/miscellanea/logo.py
QuLogic: I wonder if we should implement `__add__` and `__radd__` to support that.

greglucas: I think that should be a follow-up if someone wants to implement that. The issue will be that the `_as_mpl_transform()` requires an **ax** argument to get the geoaxes currently. There may be a way around that, but it would be a little bit more effort than I was hoping for in this cleanup ;)

QuLogic: Ah, good point, maybe that won't work so automatically then.

QuLogic (approved):

PR #1861, opened 2021-09-15
Improve benchmarking

QuLogic:
## Rationale

Benchmarking stopped working because we've added `pyproj` and `setuptools-scm` as dependencies. 

I added more benchmarks for geodesics and grid lines. Additionally, I've folded the projection benchmarks into one parameterized test and also tested multiple resolutions (though 10m was too slow and timed out, so I've had to leave it out).

## Implications

You can see an example of the results for the last "100" commits here: https://qulogic.github.io/cartopy-bench/ (it's not actually 100 commits because Cartopy does not build with proj 8, so it's not actually any further than the pyproj merge.)

### DIFF
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -47,13 +47,21 @@
     //
     "matrix": {
         "numpy": [""],
-        "cython": [""],
         "matplotlib": [""],
-        "proj4": [""],
+        "pyproj": [],
         "pykdtree": [""],
         "scipy": [""],
         "fiona": [""]
     },
+
+    // Change the build command to pre-install build-time dependencies. We
+    // don't put this in matrix, because these are not things we will ever want
+    // to compare versions.
+    "build_command": [
+        "python -mpip install cython setuptools-scm",
+        "python setup.py build",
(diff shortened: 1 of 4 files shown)


greglucas:
This looks good to me. Is there a way to run these on CI and have them uploaded somewhere automatically? It would be nice to get a rough ballpark during PRs to make sure we aren't introducing any regressions. These may be too fast currently to get meaningful results from CI though based on shared resources, but I did run across this article https://labs.quansight.org/blog/2021/08/github-actions-benchmarks/ that had some potentially interesting ideas.

Are you running these with `PYPROJ_GLOBAL_CONTEXT=ON asv ...`? That provided a big speed-up on the pyproj PR.

QuLogic:
We've been wondering that for Matplotlib; I did see that blog post and maybe it'll be okay to try shared runners, but we'd have to try it out to be certain it's not too noisy.

I did not run with that env var set; I can try out to see how that affects things.

QuLogic:
I've updated the benchmark results with the environment variable set, and marked it with a different machine name (quorum-context) so you can compare them. It makes quite a large effect to runtimes.

greglucas (approved):

PR #1874, opened 2021-09-16
Dedent some gridline code

QuLogic:
## Rationale

Most of `_draw_gridliner` is within a single `if`, which can be inverted to drop an indent level. Additionally, most of the rest is within a double-for loop, which can be combined to reduce indent another level.

## Implications

This reduces several weird wrapping issues.

I do have some followup changes, but I thought I'd open this up separately for easier review. It's probably best to look at the diff with whitespace changes ignored.


greglucas (approved):

greglucas:
Good suggestion on ignoring the whitespace changes, that did make it much more straightforward to view the diff :)

PR #1875, opened 2021-09-17
DOC: Add an example using the transform_first keyword argument

greglucas:
The example demonstrates the use of the `transform_first` keyword
argument and the negative impact it can have with wrapped coordinates.

### DIFF
--- a/examples/scalar_data/contour_transforms.py
+++ b/examples/scalar_data/contour_transforms.py
@@ -1,0 +1,52 @@
+"""
+Contour transform options
+=========================
+
+This example demonstrates the difference between transforming
+the points before/after generating the contours. It uses the
+**transform_first** keyword argument to indicate that Cartopy should
+transform the points before calling the contouring algorithm,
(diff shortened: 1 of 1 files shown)


examples/scalar_data/contour_transforms.py
```diff
@@ -0,0 +1,52 @@
+"""
+Contour transform options
+=========================
+
+This example demonstrates the difference between transforming
+the points before/after drawing the contours. It uses the
```
QuLogic, line 6: Not drawing, but generating?

examples/scalar_data/contour_transforms.py
```diff
@@ -0,0 +1,52 @@
...
+**transform_first** keyword argument to indicate that you can
```
QuLogic, line 7:
```suggestion
**transform_first** keyword argument to indicate that Cartopy should
```

QuLogic (approved):

PR #1883, opened 2021-09-21
Remove use of distutils

QuLogic:
## Rationale

Python has deprecated `distutils` and will be removing it in 3.12 (supposedly).

## Implications

No warnings with Python 3.10.


greglucas (approved):
We should probably add an ignore clause for Stickler to ignore import warnings on `setup.py`

PR #1885, opened 2021-09-22
Use UTF8 degree symbol directly

QuLogic:
## Rationale

With Python 3, all source files are UTF8, so there's no need for a Unicode escape to encode the degree symbol.

## Implications

Docs are a tiny bit clearer.

### DIFF
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -22,9 +22,9 @@ class _PlateCarreeFormatter(Formatter):
 
     _target_projection = ccrs.PlateCarree()
 
-    def __init__(self, direction_label=True, degree_symbol='\u00B0',
+    def __init__(self, direction_label=True, degree_symbol='°',
                  number_format='g', transform_precision=1e-8, dms=False,
-                 minute_symbol="'", second_symbol="''",
+                 minute_symbol='′', second_symbol='″',
                  seconds_number_format='g',
                  auto_hide=True, decimal_point=None, cardinal_labels=None):
         """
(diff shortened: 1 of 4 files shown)


lib/cartopy/mpl/ticker.py
```diff
@@ -22,7 +22,7 @@ class _PlateCarreeFormatter(Formatter):
 
     _target_projection = ccrs.PlateCarree()
 
-    def __init__(self, direction_label=True, degree_symbol='\u00B0',
+    def __init__(self, direction_label=True, degree_symbol='°',
```
QuLogic, line 25: Hmm, maybe we should use real prime glyphs for minutes and seconds.

greglucas: I think it would make sense to change those to the proper symbols, but that may require some updates to images too so perhaps you want to save that for another PR?

greglucas (approved):

lib/cartopy/mpl/ticker.py
QuLogic: It fortunately appears that we don't need to update anything.

greglucas: Fortunately, or unfortunately... This could be because our tolerances are so large like you found in your last PR :)
Feel free to self-merge if you like this as-is.

QuLogic: Heh, yes, but I don't think any tests have ticks smaller than degrees. Fortunately, they are checked text-wise in `test_ticker.py`.

Issue #1880, opened 2021-09-18
Some CRS tests fail with Proj 8.1.1


QuLogic:
### Description

Currently, tests use conda-forge, which is on 8.0.1, but Fedora Rawhide is on 8.1.1. With that version of Proj, a few tests fail now.

#### Traceback 

```
___________________________ TestCRS.test_osgb[True] ____________________________

self = <cartopy.tests.test_crs.TestCRS object at 0x7f0fd2cd1d50>, approx = True

    @pytest.mark.parametrize('approx', [True, False])
    def test_osgb(self, approx):
>       self._check_osgb(ccrs.OSGB(approx=approx))

../../BUILDROOT/python-cartopy-0.20.0-1.fc36.x86_64/usr/lib64/python3.10/site-packages/cartopy/tests/test_crs.py:73: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cartopy.tests.test_crs.TestCRS object at 0x7f0fd2cd1d50>
osgb = <Projected CRS: +proj=tmerc +datum=OSGB36 +ellps=airy +lon_0=-2 +l ...>
Name: unknown
Axis Info [cartesian]:
- E[east]...ion:
- name: unknown
- method: Transverse Mercator
Datum: OSGB 1936
- Ellipsoid: Airy 1830
- Prime Meridian: Greenwich


    def _check_osgb(self, osgb):
        ll = ccrs.Geodetic()
    
        # results obtained by streetmap.co.uk.
        lat, lon = np.array([50.462023, -3.478831], dtype=np.double)
        east, north = np.array([295132.1,  63512.6], dtype=np.double)
    
        # note the handling of precision here...
>       assert_arr_almost_eq(np.array(osgb.transform_point(lon, lat, ll)),
                             np.array([east, north]),
                             1)
E       AssertionError: 
E       Arrays are not almost equal to 1 decimals
E       
E       Mismatched elements: 2 / 2 (100%)
E       Max absolute difference: 1.62307515
E       Max relative difference: 2.55551679e-05
E        x: array([295131.,  63511.])
E        y: array([295132.1,  63512.6])

../../BUILDROOT/python-cartopy-0.20.0-1.fc36.x86_64/usr/lib64/python3.10/site-packages/cartopy/tests/test_crs.py:56: AssertionError
[...]

PR #1891, opened 2021-09-24
Switch to pytest-mpl for image testing

QuLogic:
With `pytest-mpl`, we can throw away a bunch of code for image comparison and testing. I don't consider that public API, so didn't bother writing any deprecation.

To minimize changes, I returned `ax.figure` when we have an Axes, but no explicit Figure. Also, I added in a pytest hook to generate the baseline image directory in some manner that matches the old one.


QuLogic:
You can see that failed uploads work here where I removed all non-default tolerances: https://github.com/QuLogic/cartopy/actions/runs/1268110201

lib/cartopy/tests/mpl/conftest.py
```diff
@@ -19,3 +19,20 @@ def mpl_test_cleanup(request):
     finally:
         # Closes all open figures and switches backend back to original
         plt.switch_backend(orig_backend)
+
+
+def pytest_itemcollected(item):
+    mpl = item.get_closest_marker('mpl_image_compare')
```
greglucas, line 25: Perhaps a different name than `mpl`? This confused me thinking it was `import matplotlib as mpl` down below.

greglucas (commented):
Looks like we may need to think about the `PYPROJ_GLOBAL_CONTEXT` as there are locks on the PROJ database that occasionally cause the tests to fail.

greglucas (approved):

QuLogic:
There used to be a lock on all image tests, but they were only supposed to prevent the 1 or 2 tests that used the same test image from clashing. Since `pytest-mpl` saves the test images in different directories, this shouldn't be needed. But it seems that the slowdown of the locking was enough to prevent Proj contexts from stomping on each other.

I don't know if pre-downloading the required grids will help (I already know which ones are needed from #1880).

There have been several threading/multiprocess bugs in both Proj and pyproj, and I'm not sure if conda-forge has all the fixes, or if they are all released. Maybe @snowman2 knows?

snowman2:
> I'm not sure if conda-forge has all the fixes, or if they are all released

They are all released. You just can't use the global context if you need thread safety.

QuLogic:
> > I'm not sure if conda-forge has all the fixes, or if they are all released
> 
> They are all released.

But on conda-forge too? We only get proj 8.0.1, looks like.

> You just can't use the global context if you need thread safety.

pytest-xdist is process based, I believe, not threaded.

snowman2:
> pytest-xdist is process based, I believe, not threaded

I would be surprised if the global context was safe for processes.

snowman2:
> But on conda-forge too? We only get proj 8.0.1, looks like.

It depends on the dependencies you have. If you only install phone pyproj, you will get the latest. Other dependencies might have a migration PR pending.

snowman2:
I looked at the failure message and I think it is due to the fact that PROJ NETWORK is attempting to write to the cache database at the same time. A workaround is to download the grids and disable the network. Might also be worthwhile to open an issue in PROJ.

greglucas:
I'm not sure it is due the lockfile change in this PR either (I think this has happened sporadically on some of the other PRs and even for me locally too), so I think this PR is good, modulo my comment above about the naming of "mpl".

PR #1900, opened 2021-10-06
Parametrize inline gridliner tests

QuLogic:
## Rationale

Unlike the old `ImageTesting` decorator, pytest-mpl support parametrized tests. While this does add a bunch of new tests/images, they are all far smaller than the single test, and can be run in parallel.

## Implications

The overhead of all the individual images is only ~6% bigger than the two mega test images.

Previously, our 5 slowest gridliner tests were:
```
104.48s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline
79.11s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline_usa
13.77s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels
9.77s call     lib/cartopy/tests/mpl/test_gridliner.py::test_gridliner
8.42s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_tight
```
and now they are:
```
16.03s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[NorthPolarStereo]
16.01s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[SouthPolarStereo]
12.15s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels
11.54s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[Stereographic]
10.05s call     lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[Miller]
```

But instead of getting stuck on one single test for almost 2 minutes, we can run in 4-way parallel at about 70.17s.

Since they are all generated anew, I switched to the new mpl20 style. Besides the style, I believe this causes the following changes to gridline locations for inline:

- `EuroPP` labels are half the tick interval as before
- `OSNI` labels latitude every 0.5° instead of every 1°

and for inline USA:

- `InterruptedGoodeHomolosine` labels longitude every 10° instead of every 20°
- `LambertConformal` labels longitude every 10° instead of every 20°
- `NorthPolarStereo` labels latitude every 5° instead of every 10°
- `Robinson` labels longitude every 10° instead of every 20°
- `Stereographic` labels longitude every 10° instead of every 20°

The automatic display of labels has a few small differences in the inline results, probably due to different Axes size:

- `Geostationary` labels 20°W and 40°E instead of 0°
- `LambertAzimuthalEqualArea` also labels 30°N/30°S somehow
- `LambertConformal` also labels the inner 60°W
- `NearsidePerspective` labels 40°E instead of 60°E
- `Orthographic` also labels 0° longitude
- `OSGB` also labels 6°W and 2°W on the top, and 4°W on the bottom
- `Robinson` also labels 60°W

For the inline USA results, there are fewer differences:

- `AzimuthalEquidistant` also labels 60°N, but somewhat poorly
- `EuroPP`, `OSGB`, and `OSNI` are skipped since their extent is locked
- `Geostationary` also labels 60°W on the bottom
- `Gnomonic` also labels 68°W
- `NearsidePerspective` also labels 60°W on the bottom
- `Orthographic` also labels 70°W on the bottom

### DIFF
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -208,7 +208,7 @@ def test_grid_labels():
 @pytest.mark.skipif(geos_version == (3, 9, 0), reason="GEOS intersection bug")
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(filename='gridliner_labels_tight.png',
-                               tolerance=2.9)
+                               tolerance=2.92)
 def test_grid_labels_tight():
     # Ensure tight layout accounts for gridlines
     fig = plt.figure(figsize=(7, 5))


QuLogic:
All images generated with Matplotlib 3.5.0rc1, so we'll have to see if they need tolerance increases.

QuLogic:
Without titles, the new set of images appears to be smaller even than the two originals.

PR #1910, opened 2021-10-22
Create cartopy_gridspec.py

Daviology38:
Add an example showing how to use cartopy with the matplotlib gridspec layout

### DIFF
--- a/examples/miscellanea/cartopy_gridspec.py
+++ b/examples/miscellanea/cartopy_gridspec.py
@@ -48,7 +48,9 @@ def sample_data_3d(shape):
     return lons, lats, times, data
 
 
-def make_grid(rows: int = 0, cols: int = 0, proj=ccrs.PlateCarree()) -> list:
+def make_grid(
+    rows: int = 0, cols: int = 0, proj=ccrs.PlateCarree()
+) -> list:
 
     axlist = []
     fig2 = plt.figure(figsize=(20, 10))
@@ -58,7 +60,9 @@ def make_grid(rows: int = 0, cols: int = 0, proj=ccrs.PlateCarree()) -> list:
 
     for x in range(rows):
         for y in range(cols):
-            axlist.append(fig2.add_subplot(spec2[x, y], projection=proj))
+            axlist.append(
+                fig2.add_subplot(spec2[x, y], projection=proj)
+            )
 
     return axlist
 
@@ -80,7 +84,11 @@ def main():
         ax.yaxis.set_major_formatter(lat_formatter)
 
         p = ax.contourf(
-            lons, lats, data[i, ...], transform=projection, cmap="RdBu"
+            lons,
+            lats,
+            data[i, ...],
+            transform=projection,
+            cmap="RdBu",
         )
         ax.coastlines()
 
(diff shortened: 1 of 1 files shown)


examples/miscellanea/cartopy_gridspec.py
```diff
@@ -0,0 +1,100 @@
+"""
+Using Cartopy and AxesGrid toolkit
```
greglucas, line 2: This is from the other example still.

examples/miscellanea/cartopy_gridspec.py
```diff
@@ -0,0 +1,100 @@
...
+proj (cartopy.crs projection): cartopy.crs projection to be used
+
+Returns:
+axlist: List of axes for each individual figure in the grid.
+
+>>> make_grid(2,2,ccrs.PlateCarree())
+[<AxesSubplot:>, <AxesSubplot:>, <AxesSubplot:>, <AxesSubplot:>, <AxesSubplot:>]
+
+>>> make_grid(2,2,ccrs.PlateCarree())
+[<AxesSubplot:>, <AxesSubplot:>, <AxesSubplot:>, <AxesSubplot:>]
```
greglucas, line 20: This is all extraneous and incorrect. The first iteration you have 6, then 4 but with the same call? I'd suggest removing this entire chunk starting from "Creates a list of axes" and then expanding more upon why you want to use gridspec here and what you need to do  special (only pass in a projection parameter?).

examples/miscellanea/cartopy_gridspec.py
```diff
@@ -0,0 +1,100 @@
...
+
+    return lons, lats, times, data
+
+
+def make_grid(
+    rows: int = 0, cols: int = 0, proj=ccrs.PlateCarree()
+) -> list:
+
+    axlist = []
+    fig2 = plt.figure(figsize=(20, 10))
```
greglucas, line 56: We don't want a 20x10 figure, so just leave the default here.

examples/miscellanea/cartopy_gridspec.py
```diff
@@ -0,0 +1,100 @@
...
+
+    return axlist
+
+
+def main():
+    projection = ccrs.PlateCarree()
+
+    lons, lats, times, data = sample_data_3d((6, 73, 145))
+
+    axlist = make_grid(rows=3, cols=2, proj=projection)
```
greglucas, line 75: I think we can delete the make_grid function here. I also don't think this is currently highlighting the benefits of using gridspec, it really shines when you start including slicing, so give us an example of that.

```python
fig = plt.figure()
spec = gs.GridSpec(nrows=3, ncols=3, figure=fig)
spec.update(wspace=0.2, hspace=0.5)

axlist = []
# A 1x3 plot across the entire first row
axlist.append(fig.add_subplot(spec[0, :], projection=proj))

# A 1x1 plot in the bottom left
axlist.append(fig.add_subplot(spec[0, 2], projection=proj))

# A 2x2 plot in the bottom right
axlist.append(fig.add_subplot(spec[1:, 1:], projection=proj))
```

rcomer:
I'm going to close this one because it has stalled for some time, and I agree with @greglucas that it doesn't highlight benefits of `gridspec`.  Now that [subplot_mosaic](https://matplotlib.org/stable/users/explain/axes/mosaic.html) is well-established, I'm not actually sure in what situation it would be better to use `gridspec` directly.

@Daviology38 thank you for your work on this and please let us know if you want to pick it up again.  We can always re-open.

PR #1918, opened 2021-10-31
MNT: Add a cache to the Transformer creation in transform_points

greglucas:
This adds a cache mechanism to creating Transformer objects, which
can be slow when called frequently. Additionally, this requires
adding attributes to the Mercator creation as the x/y limits are
created in the initializer which uses the cache/hash function.

closes #1915

PR #1931, opened 2021-11-13
Cache transformer crs for v0.20.x

greglucas:
Same as #1918, but pointing at v0.20.x.

I think this would be good to get in soon to contribute to a patch release as it is fixing something that I missed in the cache the first go around and should only improve performance for users.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -12,6 +12,7 @@ between them.
 
 from abc import ABCMeta
 from collections import OrderedDict
+from functools import lru_cache
 import io
 import json
 import math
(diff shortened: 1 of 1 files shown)


dopplershift (approved):

dopplershift (approved):

QuLogic:
>  [@greglucas](/greglucas) [greglucas](/greglucas) added this to the [0.20](/SciTools/cartopy/milestone/29) milestone [2 days ago](#event-5613451177)

This milestone was closed with 0.20.0; this should go on a new milestone for 0.20.2

PR #1934, opened 2021-11-14
Add pre-commit hooks and fixes

greglucas:
Added a pre-commit configuration file and applied the auto-linters to the codebase fixing up whitespace and docstrings. Not sure if we want to enable https://pre-commit.ci/ or not?


lib/cartopy/__init__.py
```diff
@@ -3,29 +3,6 @@
...
-_cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
-
-config = {'pre_existing_data_dir': os.environ.get('CARTOPY_DATA_DIR', ''),
-          'data_dir': _data_dir,
-          'cache_dir': _cache_dir,
-          'repo_data_dir': os.path.join(os.path.dirname(__file__), 'data'),
-          'downloaders': {},
-          }
 """
 The config dictionary stores global configuration values for cartopy.
```
QuLogic, line 7: This is not a module docstring; it's a docstring for the `config` dictionary. I'm not sure it actually _works_ as a docstring though.

lib/cartopy/feature/__init__.py
```diff
@@ -19,27 +19,20 @@
 import cartopy.crs
 
 
+# A dictionary of colors useful for drawing Features.
+# The named keys in this dictionary represent the "type" of
+# feature being plotted.
 COLORS = {'land': np.array((240, 240, 220)) / 256.,
```
QuLogic, line 25: I don't think this is going to become the `COLORS` doctstring.

lib/cartopy/feature/__init__.py
```diff
@@ -19,27 +19,20 @@
...
-A dictionary of colors useful for drawing Features.
-
-The named keys in this dictionary represent the "type" of
-feature being plotted.
 
-"""
+# Caches a mapping between (name, category, scale) and a tuple of the
+# resulting geometries.
 
+# Provides a significant performance benefit (when combined with object id
```
QuLogic, line 32: Goes together as one comment.
```suggestion
# resulting geometries.
#
# Provides a significant performance benefit (when combined with object id
```

setup.py
```diff
@@ -34,10 +34,7 @@
 
 from setuptools import Command, Extension, convert_path, setup
 
-"""
-Distribution definition for Cartopy.
-
-"""
+# Distribution definition for Cartopy.
```
QuLogic, line 37: Meh, just delete this one.

QuLogic (commented):
I'm not sure about some of these docstring changes, that may or may not have been working before.

greglucas:
I agree, there are some things that could be tidied up a bit still. I think before I fix up these little ones we should decide if the auto-formatting is something we want more generically or not?

After this PR I also tested out black, which surprisingly wasn't as bad as I expected it to be... https://github.com/SciTools/cartopy/compare/main...greglucas:black
So, I may even prefer to go that direction and just be done with any formatting questions. Some of the examples have obnoxious list expansion that we could ignore, but most of the updates seem pretty reasonable. My suggestion would be to swap autoflake8 to black in the pre-commit integration here if others are also interested in that swap.

rcomer:
If it helps, several other SciTools projects have adopted `black`, and I think we’re pretty happy about not having to think about the formatting any more.

bjlittle:
@greglucas As a workflow, I'd prefer to see `cartopy` use the `pre-commit.ci` service to run all registered pre-commit git-hooks as part of the pull-request workflow.

IMHO this is a light-weight way to enforce compliance and ensure code quality across the board.

bjlittle:
@greglucas Could you also adopt the use of [.git-blame-ignore-revs](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) in this PR, so that you are not blamed for other authors contributions.

You can also decorate the commit sha with a comment to record some provenance to what the commit sha is, what PR it's associated with and the change that was introcuded e.g., something along the lines of...

`.git-blame-ignore-revs`:
```
# style: flake8 (#3755)
7c86bc0168684345dc475457b1a77dadc77ce9bb
```

bjlittle:
@greglucas Are you happy for me to enable `pre-commit.ci` for `cartopy`? I've got the ownership perms to do that...

lib/cartopy/__init__.py
greglucas: OK, I agree. It doesn't do anything in the code i.e. `help(cartopy.config)` just tells me about a dict, but the documentation generated does use this. I've reverted these changes and removed that check from the pre-commit configuration.

greglucas:
I would be in favor of using `pre-commit.ci`.  One thing I have run into on other projects is that if you enable it and we don't go with it or this PR sits for a while, it will fail on the main branch annoyingly because the configuration isn't present on that branch.

bjlittle:
> I would be in favor of using `pre-commit.ci`. One thing I have run into on other projects is that if you enable it and we don't go with it or this PR sits for a while, it will fail on the main branch annoyingly because the configuration isn't present on that branch.

Well, I'm happy to merge this PR now and enable `pre-commit.ci`

I guess that might force other PR authors to rebase once this PR lands on `main`, but I'm seeing the advantage of this PR far out-weighing the rebase hassle factor atm.

bjlittle:
Note to self: use a `merge commit` and not `squash and merge`

greglucas:
> I guess that might force other PR authors to rebase once this PR lands on main

We have a lot of PRs that have been sitting for several years now, so I'm guessing rebasing is expected at this point 😄 

I'm fine with rolling with it and making updates incrementally as we go if this does cause any issues.

bjlittle:
@greglucas Tis a beautiful thing...
image
Nice one mate! 🍻

bjlittle:
@greglucas I'll push a PR to add the `pre-commit.ci` badge to the `README.md`... inbound now.

PR #1938, opened 2021-11-16
TST: Update test tolerances for MPL 3.5

greglucas:
Matplotlib 3.5 caused some images to change slightly, so increase the tolerances on those tests since they were minor.

* The three pcolormesh tests are all using the same image and looks to be very minor with a new tolerance update acceptable.

* The `multiple_projections_520` test appears to have had the entire axes stretched up (or down, I can't tell which image is which here) ever so slightly. So, the boundaries and line all are shifted a pixel or two in the new result image. A tolerance increase is probably acceptable again.

* The contours is a little more troubling. It looks like the contours now follow the boundary to close on each other and get smeared on the far right edge, whereas that wasn't the case before. @QuLogic, I know you had to update some of the contour open/closed paths on the Matplotlib side, so perhaps you have some insight on this one.
contour-result

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Latest packages
         if: steps.minimum-packages.conclusion == 'skipped'
         run: |
-          echo "PACKAGES=cython fiona matplotlib-base numpy pyproj pykdtree scipy shapely" >> $GITHUB_ENV
+          echo "PACKAGES=cython fiona matplotlib-base numpy pyproj 'proj>=8' pykdtree scipy shapely" >> $GITHUB_ENV
 
       - name: Coverage packages
         id: coverage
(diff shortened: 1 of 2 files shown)


greglucas:
Do not merge yet, I pushed this up to see if CI was able to run with 3.5 yet or not. (Apparently the answer is no)

dopplershift:
Yeah, I'm currently working on getting matplotlib 3.5 builds on conda-forge (just waiting on some new fonttools packages across the architectures since matplotlib added that dependency). Then I'll work on getting the testing builds for 3.5 up.

greglucas:
OK, I think the failures now are due to conda inconsistencies with packages. Specifically having to do with installing the `unicodedata2` package. The unicode `\N{}` is correct here and works fine normally, so that package is just my guess as to the failure.

dopplershift:
@greglucas It turns out to be an [old bug]() in `charset_normalizer`, and right now conda-forge's requests 2.26.0 package is pinning down a bugged version. This should be resolved as soon as [my PR](https://github.com/conda-forge/requests-feedstock/pull/48) lands on conda-forge's recipe.

greglucas:
Thanks for looking into that @dopplershift! It looks like that did it :)

dopplershift (approved):

Issue #1907, opened 2021-10-10
cartopy tiles in Orthographic projection miss some datas


ludwigVonKoopa:
### Description

When using tiles in Orthographic projection, i noticed missing data.

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
from cartopy.io.img_tiles import Stamen

fig = plt.figure(figsize=(5,5), dpi=120)
ax = fig.add_subplot(1, 1, 1, projection=ccrs.Orthographic(-10, 60))
tiler = Stamen('terrain-background')

ax.add_image(tiler, 3)
```

![image](https://user-images.githubusercontent.com/49512274/136713451-6bed5746-0b10-46d5-866b-5ff216d9faeb.png)

I noticed white data at the top of the globe. It doesn't appear with precision < 3 

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
from cartopy.io.img_tiles import Stamen

tiler = Stamen('terrain-background')
fig = plt.figure(figsize=(5,5), dpi=120)
for x in range(1,5):
    # print(x)
    ax = fig.add_subplot(2, 2, x, projection=ccrs.Orthographic(-10, 60))

    ax.add_image(tiler, x)
    ax.set_title(f"tile pre={x}")
```
![image](https://user-images.githubusercontent.com/49512274/136713554-ca0d4028-077c-4db9-b737-867a55b1fae7.png)

Is that something wanted ? 

Thanks

### Cartopy version

```python
import cartopy
cartopy.__version__
```
0.20.1

greglucas:
It looks like the stamen-terrain doesn't go fully up to the pole. There is a band near the pole in PlateCarree as well.
```
ax = fig.add_subplot(1, 1, 1, projection=ccrs.PlateCarree())
ax.set_global()
```

I think this is unactionable on Cartopy's side.

QuLogic:
I don't think this is referring to the pole, but the top of the circle; one tile that only shows a few pixels is not added.

ludwigVonKoopa:
I checked this issue, i found the "problem".

When cartopy load the image, it load only the tiles who will be seen in the final images (i.e. skip tiles which are behind the earth).
This is the PlateCarree representation of the projected tiles loaded, it is more understandable : 

<details><summary>code to create figure</summary>

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
from cartopy.io.img_tiles import Stamen
[...]

PR #1951, opened 2021-11-28
Orthographic threshold reduction

ludwigVonKoopa:
Related issue : #1907 

This PR changes Orthographic projection threshold to solve tiles loading error.
With actual threshold value, when using tiles with enough precision, the projection boundaries are eroded too much, and result in some tiles data not loaded : 

```python
import cartopy.crs as ccrs
from cartopy.io.img_tiles import Stamen, GoogleWTS, GoogleTiles
import matplotlib.patches as mpatches

tiler = Stamen(style="terrain-background", cache="/tmp/tiles/")

fig = plt.figure(figsize=(12,7), dpi=120)

proj1 = ccrs.Orthographic(0, 45)
proj2 = ccrs.Orthographic(0, 45)
proj2.threshold /= 20
titles = [f"threshold = {p.threshold:,.2f}" for p in (proj1, proj2)]
for p, proj, title in zip(range(2), [proj1, proj2], titles):

    ax_pla = fig.add_subplot(2, 2, 1+p, projection=ccrs.PlateCarree(), frameon=False)
    ax_ort = fig.add_axes([0.5*p, 0.1, 0.5, 0.5], projection=proj)

    bounds_proj = sgeom.Polygon(proj.boundary)
    geom = ax_ort._get_extent_geom(tiler.crs)
    img, extent, origin = tiler.image_for_domain(geom, 4)

    for ax in [ax_pla, ax_ort]:
        ax.imshow(img, extent=extent, origin=origin, transform=tiler.crs, regrid_shape=500)
        c_geom = ax.add_geometries([geom], crs=tiler.crs, color="red", alpha=0.3, zorder=30)
        c_bounds = ax.add_geometries([bounds_proj], crs=proj, color="green", alpha=0.3, zorder=40)
    ax_pla.set_title(title)
    
geo2 = mpatches.Patch(color='green', label="ortho boundaries")
geo1 = mpatches.Patch(color='red', label='eroded boundaries')
ax_pla.legend(handles=[geo1, geo2], loc=(-0.3,-0.5))
```

image

I did not add tests with this PR, because i don't see how i could easily tests this.

I could add a test which 
* create a black image
* paint tiles which pass through correct extent
* project this tiles on orthographic
* and check in the final image if there is black.

But this may be too much. What do you think ?

Thanks

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -2071,7 +2071,7 @@ class Orthographic(Projection):
         maxs = np.max(coords, axis=1)
         self._x_limits = mins[0], maxs[0]
         self._y_limits = mins[1], maxs[1]
-        self.threshold = np.diff(self._x_limits)[0] * 0.02
+        self.threshold = np.diff(self._x_limits)[0] * 0.001
 
     @property
     def boundary(self):


greglucas:
Looks like you'll have to update some images that are causing the test failures. I think all of the new images look like improvements to me, which is good :)

For the multiple projections one, we may want to just increase the tolerance a bit rather than updating that image, since it is only a small difference for one image test. (We should really break this out into separate images later on)

QuLogic:
> We should really break this out into separate images later on

Oh, wait, have I not PR'd this already? I should.

greglucas:
@ludwigVonKoopa, the images were just updated, so you're going to have to rebase this and likely need to update the Orthographic test image. Can you also remove the `0.9999*a` in the projection definition while you're at it?

greglucas:
@ludwigVonKoopa, I still think this looks like a good update if you're interested in getting it in. I'm guessing you'll need to update the Orthographic image tests already present in the test suite. One additional request would be to put a comment above, saying something about why this should be the threshold so someone else doesn't come and change it later unknowingly.

PR #1956, opened 2021-12-04
Significantly improve tight layout performance in presence of cartopy axes

lukelbd:
## Rationale

This was a surprising discovery....

Matplotlib's `get_tightbbox()` automatically skips artists that are clipped by the subplot edge to speed things up. Without [these lines](https://github.com/matplotlib/matplotlib/blob/9765379ce6e7343070e815afc0988874041b98e2/lib/matplotlib/axes/_base.py#L4646-L4657), in the presence of lots of plotted content, every single draw will take about twice as long (a pre-draw to get the bounding box extents, and then the actual draw to render the figure).

Apparently, those `get_tightbbox()` lines do not remove clipped cartopy artists. That means that cartopy quad meshes, contour sets, etc. are needlessly slowing down the tight layout algorithm.

The problem is that cartopy artists are only clipped with `clip_path`, not `clip_box` (used by the [`_get_clipped_bbox`](https://github.com/matplotlib/matplotlib/blob/fc4f0ccb831596ff57cc43ea3c38da0d922a4c77/lib/matplotlib/artist.py#L323-L338) method to bypass artists in `get_tightbbox`). I presume cartopy does not set `clip_box` because the subplot bounds are usually "curved" rather than rectangluar boxes.

I implemented a solution in lukelbd/proplot#306 (copying here) that additionally skips artists whose `clip_path` is a `TransformedPatchPath` derived from the axes patch (i.e. artists clipped by the map edge, as is the default behavior). It also bypasses this check for the three objects that matplotlib (evidently) *intends* to ignore by removing `clip_box`: children axes, x and y axis objects, and spine objects. I also verified that this block is removing the artists I think it is (try adding `print(set(super().get_default_bbox_extra_artists() - set(artists))` before the return statement).

Note this could also be accomplished by directly comparing the path coordinates... but that could be very labor-intensive, and this solution is valid all the way back to [matplotlib 2.1](https://github.com/matplotlib/matplotlib/commit/8357131482fcdfff307a9846128ebbb5e9df7c03) when `TransformedPatchPath` was first added.

This speedup from this PR is very gratifying. Looking forward to your thoughts. Since this seemed fairly cartopy-specific, I've submitted here, but let me know if I should submit to matplotlib instead.

## Implications

Faster plots and happier scientists.

## Tests

Some sample code:

```python
import numpy as np
import cartopy.crs as ccrs
fig, ax = plt.subplots(subplot_kw={'projection': ccrs.Robinson()})
N = 5000  # large dataset
lon = np.linspace(-180, 180, N)
lat = np.linspace(-90, 90, N)
data = np.random.rand(N, N)
m = ax.pcolormesh(lon, lat, data, transform=ccrs.PlateCarree())
print(m.get_clip_box() is None)  # returns True
%timeit fig.tight_layout()
```

Performance before:

```
638 ms ± 67.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Performance after:

```
2.37 ms ± 60.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

And of course the performance difference is larger the larger the dataset.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -21,6 +21,7 @@ import weakref
 import matplotlib as mpl
 import matplotlib.artist
 import matplotlib.axes
+import matplotlib.axis
 import matplotlib.contour
 from matplotlib.image import imread
 import matplotlib.transforms as mtransforms
@@ -489,6 +490,23 @@ class GeoAxes(matplotlib.axes.Axes):
         self.apply_aspect()
         for gl in self._gridliners:
             gl._draw_gridliner(renderer=renderer)
+
+    def get_default_bbox_extra_artists(self):
+        """
+        Return a default list of artists that are used for the bounding box
+        calculation. Artists are excluded by not being visible, by being clipped
+        by the axes patch path, or by ``artist.set_in_layout(False)``.
+        """
+        artists_keep = (
+            matplotlib.axes.Axes, matplotlib.axis.Axis, matplotlib.spines.Spine
+        )
+        artists = [
+            artist for artist in super().get_default_bbox_extra_artists()
+            if isinstance(artist, artists_keep)
+            or not isinstance(artist.get_clip_path(), mtransforms.TransformedPatchPath)
+            or artist.get_clip_path()._patch is not self.patch
+        ]
+        return artists
 
     def get_tightbbox(self, renderer, *args, **kwargs):
         """
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -494,16 +494,17 @@ class GeoAxes(matplotlib.axes.Axes):
     def get_default_bbox_extra_artists(self):
         """
         Return a default list of artists that are used for the bounding box
-        calculation. Artists are excluded by not being visible, by being clipped
-        by the axes patch path, or by ``artist.set_in_layout(False)``.
+        calculation. Artists are excluded by not being visible, by being
+        clipped by the axes patch path, or by ``artist.set_in_layout(False)``.
         """
         artists_keep = (
             matplotlib.axes.Axes, matplotlib.axis.Axis, matplotlib.spines.Spine
         )
+        transform = mtransforms.TransformedPatchPath
         artists = [
             artist for artist in super().get_default_bbox_extra_artists()
             if isinstance(artist, artists_keep)
-            or not isinstance(artist.get_clip_path(), mtransforms.TransformedPatchPath)
+            or not isinstance(artist.get_clip_path(), transform)
             or artist.get_clip_path()._patch is not self.patch
         ]
         return artists


lukelbd:
This is also one of those PRs that might require a thousand test image replacements, because I'm guessing "clipped" artists were probably effecting the tight layout by a pixel or two...... ugh.

dopplershift:
This sounds like a nice improvement! I'm split on whether this belongs here or in matplotlib, since this isn't using anything specific to us, it's just a case that's particularly important here.

@tacaswell @jklymak thoughts?

jklymak:
`TransformedPatchPath` seems a pretty general check.  Is there not an axes object that can be tested instead?

lukelbd:
Just to be clear I'm checking the `ax.patch` (of type `geoaxes._ViewClippedPatch`) from which the `clip_path` (of type `TransformedPatchPath`) is derived (see [the `set_clip_path` code](https://github.com/matplotlib/matplotlib/blob/c0fc78160a97264792dbb2eba964ad6975074614/lib/matplotlib/artist.py#L759-L819)). When `TransformedPatchPath` is instantiated, the parent patch is stored as a `_patch` attribute so that it can be automatically updated (see [the `TransformedPatchPath` code](https://github.com/matplotlib/matplotlib/blob/c0fc78160a97264792dbb2eba964ad6975074614/lib/matplotlib/transforms.py#L2799-L2835)). Note that the private `_patch` attribute [dates back to the origin](https://github.com/matplotlib/matplotlib/commit/8357131482fcdfff307a9846128ebbb5e9df7c03) of `TransformedPatchPath` in matplotlib 2.1. 

Re: Putting this in matplotlib vs. cartopy, on reflection, I think there are a few advantages to putting it here:

* It looks like [`set_clip_path`](https://github.com/matplotlib/matplotlib/blob/c0fc78160a97264792dbb2eba964ad6975074614/lib/matplotlib/artist.py#L759-L819) accepts a variety of input types -- not just `Patch` instances. A PR submitted to matplotlib might incur requests for a less "obvious", more general solution appopriate for all input types.
* Matplotlib's [`get_tightbbox`](https://github.com/matplotlib/matplotlib/blob/9765379ce6e7343070e815afc0988874041b98e2/lib/matplotlib/axes/_base.py#L4646-L4657) bypasses artists when the `Bbox.extents` of its `clip_box` are identical to `ax.bbox.extents`. Again, a PR submitted to matplotlib might incur requests for a similarly general solution comparing the actual `clip_path` coordinates to `TransformedPatchPath(ax.patch)` rather than just checking the `_patch` identity... which could involve more design choices and possibly reduce performance gains.
* Most importantly, I'm not sure whether "the axes patch edge defines the subplot edge" is standard procedure across all matplotlib-dependent projects. Another library might use larger or smaller axes patches that are not necessarily synonymous with the subplot edge. This solution is "stable" for cartopy because cartopy does (?) use the axes patch in this way, but I'm guessing the choice isn't canon.

A last point (of course) is that submitting here makes the improvement backwards-compatible with older matplotlib versions. Which seems like a practical advantage.

One thing that bothers me is I have no idea *where* matplotlib artists get clipped by `clip_box`, or why cartopy artists do not. I recursive grep inside matplotlib shows nothing obvious in the python code:

<details>
<summary>Grep results</summary>

```python
$ grep -r 'set_clip_box' .
../legend_handler.py:        legend_handle.set_clip_box(None)
../legend_handler.py:        legend_handle.set_clip_box(None)
../axes/_base.py:            _title.set_clip_box(None)
../tests/test_image.py:    im.set_clip_box(axs[0].bbox)
../tests/test_polar.py:    el.set_clip_box(ax.bbox)
../artist.py:    def set_clip_box(self, clipbox):
../text.py:    def set_clip_box(self, clipbox):
../text.py:        super().set_clip_box(clipbox)
../contour.py:        label.set_clip_box(self.axes.bbox)
../patheffects.py:        self.patch.set_clip_box(gc.get_clip_rectangle())
```

</details>

By contrast, a recursive grep of `set_clip_path` indicates that *all* artists (not just cartopy artists) are clipped by `ax.patch`:

<details>
<summary>Grep results</summary>

```python
$ grep -r 'set_clip_path' .
../legend_handler.py:        legend_handle.set_clip_path(None)
../legend_handler.py:        legend_handle.set_clip_path(None)
../axis.py:    def set_clip_path(self, clippath, transform=None):
../axis.py:        super().set_clip_path(clippath, transform)
../axis.py:        self.gridline.set_clip_path(clippath, transform)
../axis.py:            self.set_clip_path(self.axes.patch)
../axis.py:    def set_clip_path(self, clippath, transform=None):
../axis.py:        super().set_clip_path(clippath, transform)
../axis.py:            child.set_clip_path(clippath, transform)
../backend_bases.py:    def set_clip_path(self, path):
../axes/_axes.py:        t.set_clip_path(self.patch)
../axes/_axes.py:            a.set_clip_path(self.patch)
../axes/_axes.py:            line.set_clip_path(self.patch)
../axes/_axes.py:            im.set_clip_path(self.patch)
../axes/_axes.py:            ret.set_clip_path(self.patch)
../axes/_base.py:        self.xaxis.set_clip_path(self.patch)
../axes/_base.py:        self.yaxis.set_clip_path(self.patch)
../axes/_base.py:        a.set_clip_path(self.patch)
../axes/_base.py:            collection.set_clip_path(self.patch)
../axes/_base.py:            line.set_clip_path(self.patch)
../axes/_base.py:            p.set_clip_path(self.patch)
../axes/_base.py:        tab.set_clip_path(self.patch)
../colorbar.py:        col.set_clip_path(mpath.Path(xy, closed=True),
../backends/backend_cairo.py:    def set_clip_path(self, path):
../tests/test_artist.py:    col.set_clip_path(clip_path, ax1.transData)
../tests/test_artist.py:    patch.set_clip_path(clip_path, ax2.transData)
../tests/test_artist.py:    l.set_clip_path(p)
../tests/test_image.py:    im.set_clip_path(patch)
../tests/test_image.py:    im.set_clip_path(rect)
../tests/test_image.py:    im.set_clip_path(None)
../tests/test_bbox_tight.py:    patch.set_clip_path(path, transform=ax.transAxes)
../figure.py:            artist.set_clip_path(self.patch)
../offsetbox.py:                c.set_clip_path(tpath)
../projections/polar.py:            self.yaxis.set_clip_path(self.patch)
../artist.py:    def set_clip_path(self, path, transform=None):
../artist.py:                "Invalid arguments to set_clip_path, of type {} and {}"
../artist.py:            gc.set_clip_path(self._clippath)
../artist.py:            gc.set_clip_path(None)
../text.py:    def set_clip_path(self, path, transform=None):
../text.py:        super().set_clip_path(path, transform)
../image.py:                    gc.set_clip_path(parent.get_clip_path())
../patheffects.py:            self.patch.set_clip_path(*clip_path)
```

</details>

So, this PR "works" because cartopy artists are not clipped by `clip_box`, not because *only* cartopy artists are clipped by `ax.patch`. Curious if anyone knows why/where `clip_box` is skipped for cartopy artists (though maybe it's not critical).

greglucas:
Does this only help with pcolormesh? My initial guess is that most of the artists that are causing the slowdown are the pcolor cells involved in the wrap, which can be found with `mesh._wrapped_collection_fix`, so I wonder if we aren't adding those to the axes properly or need to do some additional work there?

https://github.com/SciTools/cartopy/blob/d35c910cbd1ee8a38e8395e0a00ff06963435448/lib/cartopy/mpl/geoaxes.py#L1900-L1920

If we remove these wrapped cells from the bbox calculation I'm wondering if we would start to shrink the limits when the cells are wrapped by not accounting for the pcolor anymore. Basically what I'm getting at is that I _think_ the slowdown is due to slow trans_non_affine calculations from the pcolor cell bounds calculation and we are just removing that slow calculation here, but at the loss of auto-scaling bounds calculations.

greglucas:
After thinking more about what I just wrote above, I wonder why we need to calculate the bounds for those pcolor cells at all? We **know** they are guaranteed to be the width of the plot, so the bounds for those cells should be much easier to compute than a slow non-affine transform. Perhaps we should override the get_datalim() calculation for that collection...

lukelbd:
No the slowdown is present in almost all situations -- every artist drawn in a `GeoAxes` has the same issue that `clip_path` is set while `clip_box` is not. I believe the issue simply derives from matplotlib drawing the figure "twice" in all situations (once for the bounding box, once to actually render the image). I'm guessing the `GeoAxes.pcolormesh` algorithm could be independently improved, but I think that's a separate point. Here's an example with `contourf` (and a smaller dataset):

```python
import numpy as np
import cartopy.crs as ccrs
fig, ax = plt.subplots(subplot_kw={'projection': ccrs.Robinson()})
N = 100  # medium dataset
lon = np.linspace(-180, 180, N)
lat = np.linspace(-90, 90, N)
data = np.random.rand(N, N)
m = ax.contourf(lon, lat, data, transform=ccrs.PlateCarree())
```

Performance before:

```
>>> %timeit fig.tight_layout()
18 ms ± 1.53 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Performance after:

```
>>> %timeit fig.tight_layout()
2.72 ms ± 237 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

greglucas:
Thanks for clarifying, @lukelbd! Sorry for the noise, I opened a separate issue for that comment.

tacaswell:
If I am understanding this right, there is a short cut that says "if you are clipped to your Axes bounding box, then I can ignore you and just ask your containing Axes how big it is (as I trust it to clip you to it)" and the generalization here is  "... to your *transformed* Axes bounding box ..."? 

While I agree with @jklymak that there is probably a better way to do this, I think this is a reasonable extra fast-path to pick up upstream.

lukelbd:
> If I am understanding this right, there is a short cut that says "if you are clipped to your Axes bounding box, then I can ignore you and just ask your containing Axes how big it is (as I trust it to clip you to it)" and the generalization here is "... to your transformed Axes bounding box ..."?

That's about right, although not sure what you mean by "transformed bounding box" -- I think it's more like "if you are clipped to your Axes background patch" (an instance of `Patch`; the `grep` I posted above suggests this is almost always the case) in addition to "... clipped to your Axes bounding box" (an instance of `BBox`). The `Bbox` itself is always square, and is currently the only thing matplotlib's `get_tightbbox` uses to bypass artists.

lukelbd:
Just wanted to ping @jklymak and @dopplershift (not sure if Github sends notifications when I edit comments). Do you agree with @tacaswell that I should try a matplotlib PR instead?

Note that in a matplotlib PR, I think I'd make a couple changes:

1. For consistency with the identity check `TransformedPatchPath._patch is ax.patch`, I would propose replacing [these lines](https://github.com/matplotlib/matplotlib/blob/9765379ce6e7343070e815afc0988874041b98e2/lib/matplotlib/axes/_base.py#L4646-L4657) with a similar identity check `clip_box is ax.bbox`. I guess that's not necessary... but seems kind of inelegant to compare the `clip_box` boundary based on coordinates and the `clip_path` boundary based on identity.
2. As I mentioned above, it's possible that "the axes patch edge defines the subplot edge" is not standard procedure across all projects... i.e. the correct tight layout calculation could *require* artists clipped by `ax.patch`. The workaround would be to add `ax.patch` to the `get_tightbbox()` algorithm, which would implicitly cover artists clipped by `ax.patch` (note `get_tightbbox()` does not currently include `ax.patch`).

jklymak:
I think a PR would be fine.  If there are artists kicking around that are clipped, if they have a clip_path that is fully inside the axes, they should also be removed from the draw list.  This just seems an oversight in the fast-path logic, not a deliberate design choice.

lukelbd:
Cool -- then I think I'll close this and submit a matplotlib PR later today or tomorrow.

lukelbd:
Please see matplotlib/matplotlib#21935 for further updates.

lukelbd:
Thanks for the help everyone! matplotlib/matplotlib#21935 was merged.

Turns out the speedup is specific to the case where artist coordinates are converted between coordinate systems (generally longitude-latitude to map coordinates). This conversion has to happen in both `draw` and `get_window_extent`, and it is often the limiting step for hi-res artists.

PR #1646, opened 2020-08-29
Pcolormesh update for interpolation with wrapped coordinates

greglucas:
Fixes #1638 

## Rationale
The new shading keyword argument in MPL 3.3 causes an interpolation that creates new edges. With coordinates that could potentially be wrapped, as in map projections, this could cause new erroneous interpolation points to be inserted into the coordinates.

## Implications

A big refactor of pcolormesh to avoid calling private MPL functions. We still have to use the private `_coordinates` Quadmesh attribute to get the coordinates, I couldn't find a way around that for now.

We also now calculate some edges in Cartopy, meaning more code to maintain.

PR #1958, opened 2021-12-05
FIX: Calculate bbox for pcolormesh

### LINKED ISSUE #1947: Global map cropped with pcolormesh when lats > 90
Here is a mwe:

```python
import matplotlib
import numpy as np 
import cartopy
import cartopy.crs as ccrs  

print('matplotlib', matplotlib.__version__)
print('cartopy', cartopy.__version__)

lon = np.linspace(-179.625, 179.625, 480)
lat = np.linspace(90, -90, 241)
data, _ = np.meshgrid(lon, lat)

plt.figure()
ax = plt.axes(projection=ccrs.EqualEarth())
plt.pcolormesh(lon, lat, data, transform=ccrs.PlateCarree(), shading='nearest')
ax.coastlines(); ax.gridlines();
```

Which produces:
image

Strangely enough, this does not occur with other coordinates (here in steps of 1):

```python
lon = np.linspace(-180, 180, 361)
lat = np.linspace(90, -90, 181)
data, _ = np.meshgrid(lon, lat)

plt.figure()
ax = plt.axes(projection=ccrs.EqualEarth())
plt.pcolormesh(lon, lat, data, transform=ccrs.PlateCarree(), shading='nearest')
ax.coastlines(); ax.gridlines();
```

image

This is:

```
matplotlib 3.4.3
cartopy 0.20.1
```

greglucas:
This was missed in the pcolormesh refactor causing some incorrect bounds calculations when plotting the data. It was typically covered up by the pcolor wrapped addition updating the limits there, so it wasn't caught earlier.
Commit: https://github.com/SciTools/cartopy/commit/00400f3a9aa12c16e3e6d7c979c6a3c50ed51df9
PR: https://github.com/SciTools/cartopy/pull/1646

Added some tests to try and check these data limits in the future.

Fixes #1947

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1794,10 +1794,6 @@ class GeoAxes(matplotlib.axes.Axes):
         cross the boundary of the projection.
         """
         t = kwargs.get('transform', None)
-        if not (getattr(t, '_wrappable', False) and
-                getattr(self.projection, '_wrappable', False)):
-            # Nothing to do
-            return collection
 
         # Get the quadmesh data coordinates
         coords = collection._coordinates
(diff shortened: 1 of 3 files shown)


dopplershift (approved):
Seems good to me. Any idea about the macos 3.7 failure?

greglucas:
I have no idea on that failure... Looks like the geos libraries are struggling with linking, so it could be something related to the shapely/cartopy geos library imports got mixed up somewhere along the line in a recent conda build of shapely or geos...

QuLogic:
Well, a rebuild made it worse... Seems like one of the shapefiles was updated and now the result is a bit different, so all builds fail. The original failure is still broken too.

PR #1963, opened 2021-12-11
CI: Fix min dependency build by switching to Linux

greglucas:
There is some conda-forge issue with Shapely/geos so switch over to Linux for the min dependency tests. This will fix the macos 3.7 min version failures. Note this will still have failures that will be fixed by #1962.

I tried removing DYLD fallback library path and a few other things that didn't work, so this is the simplest solution for our test suite.

xref: 
https://github.com/shapely/shapely/issues/394
https://github.com/conda-forge/shapely-feedstock/issues/53

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -25,13 +25,12 @@ jobs:
           channels: conda-forge/label/testing,conda-forge
 
       - name: Minimum packages
-        # Only run on macos for now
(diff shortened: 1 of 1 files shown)


QuLogic:
I think you can rebase this.

QuLogic:
Or actually re-opening might be enough.

QuLogic (approved):

Issue #1932, opened 2021-11-13
Rename the "master" branch to "main"


greglucas:
I propose to rename the master branch to main. I'm not sure when the best time to do this is, but we don't have a major amount of contributors right now that would be affected by this. GitHub automatically takes care of the repointing of currently open PRs.

dopplershift:
That's fine with me. I don't think it should kick off too many problems.

PR #1976, opened 2021-12-24
Rename master branch to main

QuLogic:
## Rationale

Matplotlib renamed the default branch, and it's been annoying to switch back and forth.
There fortunately seem to be very few mentions of the default branch in the source code.

## Implications

Nothing right now until we toggle the branch in settings.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -28,7 +28,7 @@ installed using the pip installer::
 
     pip install cartopy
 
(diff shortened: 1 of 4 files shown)


greglucas:
I agree that this would be nice.
Closes #1932

Feel free to self-merge and switch the default branch after that.

greglucas (approved):

PR #1977, opened 2021-12-24
CI: Avoid newer openssl version due to WMTS tests

greglucas:
The negotiation with a tile provider needed more up-to-date TLS/SSL protocols, so for now we just pin to a lower openssl
on the runners.

I am able to reproduce locally if I install openssl 3.0.0. I think this is a server-side issue on NASA's end, so we will have to wait for them to update to more secure protocols before unpinning.


QuLogic (commented):
Seems good for now.

greglucas:
I reached out to NASA and they responded saying it would be in early 2022 (Jan/Feb time frame). So, we'll have to wait a month or two before unpinning.

PR #1980, opened 2022-01-05
MNT: Change map_projection to projection for Axes creation

### LINKED ISSUE #1979: Instantiating from class cartopy.mpl.geoaxes.GeoAxes(*args, **kwargs)[source]
### Description
Using
`vAxes = cartopy.mpl.geoaxes.GeoAxes(projection=ccrs.Mercator())`

throws `KeyError`:

```
File "/usr/lib64/python3.10/site-packages/cartopy/mpl/geoaxes.py", line 410, in __init__
self.projection = kwargs.pop('map_projection')
KeyError: 'map_projection'
```

Note that the following _does_ work:
```
vAxes = matplotlib.pyplot.axes(projection=ccrs.Mercator())

```
Although this throws a warning `QSocketNotifier: Can only be used with threads started with QThread` - which is a different issue?

#### Code to reproduce

```
import cartopy.crs as ccrs
import cartopy.mpl.geoaxes as cga

vAxes = cga.GeoAxes(projection=ccrs.Mercator())
```

#### Traceback 

```
File "/usr/lib64/python3.10/site-packages/cartopy/mpl/geoaxes.py", line 410, in __init__
self.projection = kwargs.pop('map_projection')
KeyError: 'map_projection'
```

<details>

### Operating system
Fedora 35

### Cartopy version
0.20.1

</details>

greglucas:
The GeoAxes class now requires the `projection` keyword argument rather than `map_projection` to be more consistent with how a normal Axes is created. In particular, this makes working with AxesGrid and other toolkits easier by enabling a consistent keyword when passing `subplot_kw`, `axes_class`, etc...

closes #1979

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -399,16 +399,22 @@ class GeoAxes(matplotlib.axes.Axes):
 
         Parameters
         ----------
-        map_projection: optional
-            The target :class:`~cartopy.crs.Projection` of this Axes object.
-
-
-        All other args and keywords are passed through to
-        :class:`matplotlib.axes.Axes`.
-
+        projection : cartopy.crs.Projection
+            The target projection of this Axes.
         """
-        self.projection = kwargs.pop('map_projection')
-        """The :class:`cartopy.crs.Projection` of this GeoAxes."""
+        if "map_projection" in kwargs:
+            warnings.warn("The `map_projection` keyword argument is "
+                          "deprecated, use `projection` to instantiate a "
+                          "GeoAxes instead.")
+            projection = kwargs.pop("map_projection")
+        else:
+            projection = kwargs.pop("projection")
+
+        # The :class:`cartopy.crs.Projection` of this GeoAxes.
+        if not isinstance(projection, ccrs.Projection):
+            raise ValueError("A GeoAxes can only be created with a "
+                             "projection of type cartopy.crs.Projection")
+        self.projection = projection
 
         super().__init__(*args, **kwargs)
         self._gridliners = []
--- a/examples/miscellanea/axes_grid_basic.py
+++ b/examples/miscellanea/axes_grid_basic.py
@@ -41,7 +41,7 @@ def sample_data_3d(shape):
 def main():
     projection = ccrs.PlateCarree()
     axes_class = (GeoAxes,
-                  dict(map_projection=projection))
+                  dict(projection=projection))
 
     lons, lats, times, data = sample_data_3d((6, 73, 145))
 
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -775,7 +775,7 @@ class Projection(CRS, metaclass=ABCMeta):
 
     def _as_mpl_axes(self):
         import cartopy.mpl.geoaxes as geoaxes
-        return geoaxes.GeoAxes, {'map_projection': self}
+        return geoaxes.GeoAxes, {'projection': self}
 
     def project_geometry(self, geometry, src_crs=None):
         """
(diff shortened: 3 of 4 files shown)


dopplershift:
The change seems sensible. It's unclear to me whether the test failure is related.

greglucas:
Hm... I am pretty sure it isn't related to this because this would have complained about bad "map_projection" or similar errors.

Perhaps the numpy 1.22 floating point difference from AVX512... The difference is just over the tolerance now and only on the Linux runners.

greglucas:
@dopplershift, I _think_ the failing tests were due to the floating point issues with AVX512 after all. I added a commit to increase the floating point dtype size https://github.com/SciTools/cartopy/pull/1980/commits/7c71977b659c402cd1846dd9ab2ef9de7c146b68, which appears to have fixed the failing CI. I can break that out to another PR if it would help too.

greglucas:
@dopplershift, I _think_ the commit to address that other failed test example is in this PR... Maybe I should have broken it out :)

lib/cartopy/mpl/geoaxes.py
```diff
@@ -392,23 +392,27 @@ class GeoAxes(matplotlib.axes.Axes):
...
-            The target :class:`~cartopy.crs.Projection` of this Axes object.
-
-
-        All other args and keywords are passed through to
-        :class:`matplotlib.axes.Axes`.
-
-        """
-        self.projection = kwargs.pop('map_projection')
-        """The :class:`cartopy.crs.Projection` of this GeoAxes."""
+        projection : cartopy.crs.Projection, required
```
dopplershift, line 402: `projection` has a default argument so is still optional:
```suggestion
        projection : cartopy.crs.Projection, optional
```

dopplershift (changes requested):
Minor doc adjustment and this can go in.

lib/cartopy/mpl/geoaxes.py
greglucas: Yep, fixed that.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -392,23 +392,27 @@ class GeoAxes(matplotlib.axes.Axes):
...
+            The target projection of this Axes.
+        """
+        if "map_projection" in kwargs:
+            warnings.warn("The `map_projection` keyword argument is "
+                          "deprecated, use `projection` to instantiate a "
+                          "GeoAxes instead.")
+            projection = kwargs.pop("map_projection")
+
+        # The :class:`cartopy.crs.Projection` of this GeoAxes.
+        if not isinstance(projection, ccrs.Projection):
```
dopplershift, line 412: So I missed this last time. If `projection` is left as its default of `None`, this check fails. Previously, it would have failed on `kwargs.pop('map_projection')`. Was there any reason you made it "optional"?

dopplershift (changes requested):

lib/cartopy/mpl/geoaxes.py
greglucas: Oh good catch. I can't say I gave this a ton of thought... I was more hoping for consistency in downstream calls like AxesGrid, rather than looking at the specifics.

I guess this should actually be `required` above. It was previously required as you say, but gave a different error if it wasn't included... I'm inclined to update the docstring to `required` and not change it to raising a `KeyError`.

dopplershift: Fine with me. Just make sure to also remove the default to `None`.

greglucas: actually, I have no idea why I did that way in the first place... Let's just pop the new keyword in the else branch and then the same error will be thrown. But, we also have the additional check later on to see if it is a Projection.

dopplershift (approved):

Issue #1984, opened 2022-01-14
Shapely deprecation warning from ShapelyFeature class


chowington:
### Description
The `ShapelyFeature` class causes a deprecation warning from Shapely to appear.

#### Code to reproduce

Instantiate a `ShapelyFeature`.

#### Traceback 

```
/home/chowingt/miniconda3/envs/GOS-env/lib/python3.9/site-packages/cartopy/feature/__init__.py:217: ShapelyDeprecationWarning: Iteration over multi-part geometries is deprecated and will be removed in Shapely 2.0. Use the `geoms` property to access the constituent parts of a multi-part geometry.
  self._geoms = tuple(geometries)
/home/chowingt/miniconda3/envs/GOS-env/lib/python3.9/site-packages/cartopy/feature/__init__.py:217: ShapelyDeprecationWarning: __len__ for multi-part geometries is deprecated and will be removed in Shapely 2.0. Check the length of the `geoms` property instead to get the  number of parts of a multi-part geometry.
  self._geoms = tuple(geometries)
```

<details>
 <summary>Full environment definition</summary>

### Operating system
Ubuntu 20.04.3

### Cartopy version
0.20.2

### conda list

```
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
alabaster                 0.7.12                     py_0    conda-forge
argon2-cffi               20.1.0           py39h27cfd23_1
async_generator           1.10               pyhd3eb1b0_0
attrs                     21.2.0             pyhd3eb1b0_0
babel                     2.9.1              pyh44b312d_0    conda-forge
backcall                  0.2.0              pyhd3eb1b0_0
blas                      1.0                         mkl
bleach                    4.1.0              pyhd3eb1b0_0
bottleneck                1.3.2            py39hdd57654_1
brotli                    1.0.9                he6710b0_2
brotlipy                  0.7.0           py39h3811e60_1003    conda-forge
bzip2                     1.0.8                h7f98852_4    conda-forge
c-ares                    1.18.1               h7f98852_0    conda-forge
ca-certificates           2021.10.26           h06a4308_2
cartopy                   0.20.2           py39h8394fad_0    conda-forge
certifi                   2021.10.8        py39h06a4308_2
cffi                      1.14.6           py39h400218f_0
charset-normalizer        2.0.10             pyhd8ed1ab_0    conda-forge
[...]

PR #1930, opened 2021-11-12
Address Shapely 2.0 Deprecations

akrherz:
#1914 commits cherry-picked to v0.20.x branch, refs #1913

Issue #1303, opened 2019-05-04
Colouring countries


Especuloide:
### Description
Cartopy stop working when you try to color some countries, like Austria (AUT), Albania among others in Europe as far as I could see. My maps heve been working fine until couple last update on april.

#### Code to reproduce
```python
import matplotlib.pyplot as plt
import cartopy
import cartopy.io.shapereader as shpreader
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import cartopy.feature as cfeature


def area(ax, iso, clr) :
  
    shp = shpreader.natural_earth(resolution='10m',category='cultural',name='admin_0_countries')
    reader = shpreader.Reader(shp)
    countries = reader.records()
  
    for n in countries :
        if n.attributes['ADM0_A3'] == iso: 
            ax.add_geometries(n.geometry, ccrs.PlateCarree(),
                              facecolor=clr, alpha = 1.00, linewidth =0.15, edgecolor = "black",
                              label=n.attributes['ADM0_A3']) 
          
    return ax

iso3 = ['USA','CAN','RUS','GBR','ISL','FRA','ITA','AUT'] # works fine with this codes
#iso3 = ['USA','CAN','RUS','GBR','ISL','FRA','ITA','CHN','AUT'] # error message - due to  "AUT" - Austria 

def main():
    ax = plt.axes(projection=ccrs.Miller())
    
    states_provinces = cfeature.NaturalEarthFeature(
        category='cultural',
        name='admin_1_states_provinces_lines',
        scale='50m',
        facecolor='none')

    ax.add_feature(cfeature.LAND)
    ax.add_feature(cfeature.COASTLINE)
    ax.add_feature(states_provinces, edgecolor='gray')

    for n in iso3 :
        area(ax,n,"red")         
             
    plt.show()


if __name__ == '__main__':
    main()
```

#### Traceback

```python-traceback

File "C:\ProgramData\Anaconda3\lib\site-packages\cartopy\mpl\geoaxes.py", line 588, in add_geometries
  feature = cartopy.feature.ShapelyFeature(geoms, crs, **kwargs)

File "C:\ProgramData\Anaconda3\lib\site-packages\cartopy\feature\__init__.py", line 231, in __init__
  self._geoms = tuple(geometries)

TypeError: 'Polygon' object is not iterable
```

<details>
 <summary>Full environment definition</summary>

### Operating system
Windows 10

### Cartopy version
0.17.0

### conda list
[...]

PR #1985, opened 2022-01-14
Use geometries.geoms to fix deprecation warning

chowington:
## Rationale

Fix deprecation warning from Shapely, in the vein of https://github.com/SciTools/cartopy/pull/1930.

## Implications

None.

Fixes https://github.com/SciTools/cartopy/issues/1984.

### DIFF
--- a/lib/cartopy/feature/__init__.py
+++ b/lib/cartopy/feature/__init__.py
@@ -214,7 +214,7 @@ class ShapelyFeature(Feature):
 
         """
         super().__init__(crs, **kwargs)
-        self._geoms = tuple(geometries)
+        self._geoms = tuple(geometries.geoms)
 
     def geometries(self):
         return iter(self._geoms)


akrherz:
Thank you, sorry I missed one :(

chowington:
No prob!

Unfortunately it looks like some tests are failing. Maybe the fix isn't as simple as I'd thought. I'm not actually that familiar with this package, so someone else my need to look into it

rcomer:
According to the docstring, the class expects a collection of geometries, so the tests are passing tuples and lists.  It seems reasonable to me that a user might want to pass a single geometry e.g. via `GeoAxes.add_geometries`.  One option might be to check whether `geometries` is a shapely geometry object and turn it into a single element tuple before the line that is throwing the warning.

Disclaimer:  I am not a Cartopy developer so have no authority here.  Just making a suggestion!

greglucas:
I think @rcomer is on the right track. I think that suggestion would also help some other cases too with the pluralization confusion. See here for a possible starting implementation: https://github.com/SciTools/cartopy/issues/1303#issuecomment-619641832

greglucas:
I believe this was superseded by https://github.com/SciTools/cartopy/pull/1999 which should allow single geometries to be passed through now. Feel free to re-open if that is not the case.

PR #1987, opened 2022-01-15
avoid an edge case with base=0 in Gridliner._round

syrte:
## Rationale

The round function does not work when base=0.
The existing code for testing nan might be enough to justify testing 0.


syrte:
Sorry for my inconsiderate pull request. Raising error should be better than working smoothly with wrong output.

PR #1990, opened 2022-01-22
Minimum Python 3.8

greglucas:
This bumps the minimum supported version to Python 3.8 for the next release and also adds Python 3.10 to the tests.

https://numpy.org/neps/nep-0029-deprecation_policy.html

### DIFF
--- a/lib/cartopy/tests/mpl/test_mpl_integration.py
+++ b/lib/cartopy/tests/mpl/test_mpl_integration.py
@@ -8,7 +8,6 @@ import re
 
 import numpy as np
 import matplotlib.pyplot as plt
-from packaging.version import parse as parse_version
 import pytest
 
 import cartopy.crs as ccrs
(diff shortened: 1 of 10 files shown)


QuLogic (approved):

greglucas:
I just removed the last commit to bump the test tolerances. I'm hoping those are fixed by https://github.com/SciTools/cartopy/pull/1994 instead, which won't need a tolerance bump then. So, if that goes in, then the 3.9/3.10 test_examples should be fine here too.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -112,7 +111,7 @@ def test_global_scatter_wrap_no_transform():
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(
     filename='global_hexbin_wrap.png',
-    tolerance=2 if MPL_VERSION < parse_version('3.2') else 0.5)
+    tolerance=0.5)
```
QuLogic, line 114: The default is 0.5, so this can be dropped.

greglucas: Thanks, and apologies for the CI thrashing here. The conda-forge testing labels were not in all of the releases, so had to pin to a specific version and make sure `matplotlib-base` is from the testing label too.

dopplershift: @greglucas Apologies on that. Probably those releases pre-date cartopy using them. It should be much easier keeping testing up-to-date nowadays.

greglucas: No problem at all, that actually wasn't too bad to diagnose compared to some of the other flaky failures we seem to have here ;)

PR #1994, opened 2022-01-25
TST: Increase dtype of sample data calculations

greglucas:
This helps avoid some floating point differences in the image tests.

Breaking this out into a separate PR, rather than burying it in #1980


QuLogic (approved):
I don't love having to do this, but it's only a test.

greglucas:
Agreed, it isn't ideal. I think this numpy instruction set update caused some unfortunate side effects. Although I think that is a pretty good testament to just how stable numpy has been up until now with floating point calcs!

Issue #2013, opened 2022-03-10
API: inconsistent projection arguments in img_transform.regrid


ellequelle:
While helping someone with their code I noticed this inconsistency in the `img_transform.regrid` API and docstring. The projection parameters are `source_cs` and `target_proj`, but in the docstring they're listed as `source_cs` and `target_cs`:

<details><summary>regrid excerpt</summary>

https://github.com/SciTools/cartopy/blob/22cdafca511744143c5b673598a1889f169c1e3d/lib/cartopy/img_transform.py#L222-L241

</details>

Could this be changed to match `img_transform.wrap_array`, with `source_proj` and `target_proj`?

<details><summary>wrap_array excerpt</summary>

https://github.com/SciTools/cartopy/blob/22cdafca511744143c5b673598a1889f169c1e3d/lib/cartopy/img_transform.py#L124-L141

</details>

greglucas:
I prefer consistency, so I would say yes that sounds like a great idea. A PR for the change would be great! There are probably other locations of mismatch as well, so if there are others you find don't hesitate to make PRs for them.

dopplershift:
If we do that, it's technically an API change (since people passing via kwargs would be broken). That's fine, but we need to make sure it gets noted in the release notes.

ellequelle:
I have two questions:
1. Does cartopy have a standard way of dealing with API deprecations?
1. Is there a whatsnew file or someplace to make note of the API change?

dopplershift:
@ellequelle In some cases we'll warn for a change. I think this is minimally disruptive, and given our 0.x versioning, I feel fine just making the change. We generally don't have an on-going what's new file, but the release manager writes up the release notes from the PRs included in the release milestone. The easiest way to deal with this might be to just add an "api change" label.

Issue #2012, opened 2022-03-09
pyshp-2.2.0 test failure


manisandro:
I'm building proj9 for fedora, and rebuilding cartopy against it yielded the following test failure:

    =================================== FAILURES ===================================
    ___________________________ TestLakes.test_geometry ____________________________
    [gw0] linux -- Python 3.10.2 /usr/bin/python3

    self = <cartopy.tests.test_shapereader.TestLakes object at 0x7f7df66ec190>

        def test_geometry(self):
            lake_geometry = self.test_lake_geometry
            assert lake_geometry.type == 'Polygon'
        
            polygon = lake_geometry
        
            expected = np.array([(-84.85548682324658, 11.147898667846633),
                                (-85.29013729525353, 11.176165676310276),
                                (-85.79132117383625, 11.509737046754324),
                                (-85.8851655748783, 11.900100816287136),
                                (-85.5653401354239, 11.940330918826362),
                                (-85.03684526237491, 11.5216484643976),
                                (-84.85548682324658, 11.147898667846633),
                                (-84.85548682324658, 11.147898667846633)])
        
    >       assert_array_almost_equal(expected, polygon.exterior.coords)
    E       AssertionError: 
    E       Arrays are not almost equal to 6 decimals
    E       
    E       Mismatched elements: 12 / 16 (75%)
    E       Max absolute difference: 0.75447591
    E       Max relative difference: 0.00887234
    E        x: array([[-84.855487,  11.147899],
    E              [-85.290137,  11.176166],
    E              [-85.791321,  11.509737],...
    E        y: array([[-84.855487,  11.147899],
    E              [-84.855487,  11.147899],
    E              [-85.036845,  11.521648],...

    ../../BUILDROOT/python-cartopy-0.20.1-1.fc37.x86_64/usr/lib64/python3.10/site-packages/cartopy/tests/test_shapereader.py:45: AssertionError
    ---------------------------- Captured stdout setup -----------------------------

manisandro:
Actually this looks like a bug in pyshp-2.2.0. With pyshp-2.1.3 the test passes.

manisandro:
[...]

PR #2014, opened 2022-03-10
API: GH #2013 make regrid arguments, docstring consistent

ellequelle:
## Rationale
Addresses #2013 

#### This is a change to the API for `img_transform.regrid`

I changed the name of the 4th argument from `source_cs` to `source_proj` to be consistent with other argument names. It's probably uncommon for source projection object to be passed as a keyword argument since it's the fourth of six positional arguments.

#### Old function signature:
https://github.com/SciTools/cartopy/blob/22cdafca511744143c5b673598a1889f169c1e3d/lib/cartopy/img_transform.py#L222-L223

#### New function signature:
```python
def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
           target_x_points, target_y_points, mask_extrapolated=False)
```

### DIFF
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -325,7 +325,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
                                                       desired_nx)
 
         bounds = _determine_bounds(source_x_coords, source_y_coords,
-                                    source_proj)
+                                   source_proj)
 
         outside_source_domain = ((target_in_source_y >= bounds['y'][1]) |
                                  (target_in_source_y <= bounds['y'][0]))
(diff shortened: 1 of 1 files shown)


greglucas (approved):
:+1: I think this has little chance of impacting people downstream and isn't worth doing a lot of deprecation dance around.

dopplershift:
I agree, my only concern is the test failures, which I've not had a chance to dig into to see if they're related.

ellequelle:
I've been confused by this one, but I get an identical error (cartopy.tests.test_shapereader.TestLakes) when testing the head branch on my computer, so it doesn't seem to be due to any of these changes.

greglucas:
Yes, that looks like the same error as here: https://github.com/SciTools/cartopy/issues/2012
which is caused by pyshp, and our minimum dependencies is the only location where we are using pyshp.

greglucas:
Thanks, @ellequelle!

Issue #2017, opened 2022-03-22
Missing default natural earth shapefiles


yaswant:
### Description
The cartopy geoaxes `coastlines()` method attempts to download (and store them in local cartopy cache dir) 110m natural earth shapefiles.  This is not ideal, particularly when a stack is deployed on a machine which is not internet facing and the user has not manually updated the cache folder.

For example the following code will hang if there is no internet connection and the only way out is via keyboard interruption (eg `Ctrl+C`)

#### Code to reproduce
For example, on a Linux OS, remove `~/.local/share/cartopy` and disconnect from the internet and run:

```python
#!/usr/bin/env python
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

proj = ccrs.PlateCarree()
fig = plt.figure(figsize=(6, 4))
ax = fig.add_subplot(111, projection=proj)
ax.set_global()
ax.coastlines()
plt.show()
```

#### Traceback 
```
.../lib/python3.6/site-packages/cartopy/io/__init__.py:260: DownloadWarning: Downloading: https://naciscdn.org/naturalearth/110m/physical/ne_110m_coastline.zip
  warnings.warn('Downloading: {}'.format(url), DownloadWarning)

^CException in Tkinter callback
Traceback (most recent call last):
  File ".../lib/python3.6/tkinter/__init__.py", line 1702, in __call__
    return self.func(*args)
  File ".../lib/python3.6/tkinter/__init__.py", line 746, in callit
    func(*args)
  File ".../lib/python3.6/site-packages/matplotlib/backends/_backend_tk.py", line 253, in idle_draw
    self.draw()
  File ".../lib/python3.6/site-packages/matplotlib/backends/backend_tkagg.py", line 9, in draw
    super(FigureCanvasTkAgg, self).draw()
  File ".../lib/python3.6/site-packages/matplotlib/backends/backend_agg.py", line 407, in draw
    self.figure.draw(self.renderer)
  File ".../lib/python3.6/site-packages/matplotlib/artist.py", line 41, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File ".../lib/python3.6/site-packages/matplotlib/figure.py", line 1864, in draw
    renderer, self, artists, self.suppressComposite)
  File ".../lib/python3.6/site-packages/matplotlib/image.py", line 131, in _draw_list_compositing_images
    a.draw(renderer)
[...]

Issue #1325, opened 2019-06-14
Clarifying the 'under the hood' data download (and other GIS stuff)


jypeter:
I have just used `ax_plot.add_feature(cfeature.LAND)` for the first time, which triggered an **external data download**
```
/home/share/unix_files/cdat/miniconda3/envs/cdatm_py2/lib/python2.7/site-packages/cartopy/io/__init__.py:260: DownloadWarning: Downloading: http://naciscdn.org/naturalearth/110m/physical/ne_110m_land.zip
  warnings.warn('Downloading: {}'.format(url), DownloadWarning)
```
This went quite well, but it reminded me that such a **download recently failed when using cartopy on a supercomputer** with no outside world access. I was in a hurry and did not write down the error message, so I can't say if it was friendly and informative or not (I figured out what it was about, and a workaround, but you want to make this easy for all users)

Anyway, I think it would be nice to add a page somewhere about:
- which data is available by default
- what might be downloaded and where (in my case, the data ended up in `cartopy.config['data_dir']'/shapefiles/natural_earth/physical'`, more precisely in a **hidden directory** `~/.local/share/cartopy/shapefiles/natural_earth/physical`)
- some recipe to pre-download the data on a computer with network access, and copy it to the appropriate directory on a supercomputer
- a short summary (or links to appropriate external web sites) of how this shapefiles/features/layers/WMS stuff work, to help modellers that are not familiar with the _GIS stuff_

And you could then add a link to this new page from [The cartopy Feature interface](https://scitools.org.uk/cartopy/docs/latest/matplotlib/feature_interface.html) page and other relevant places

joernu76:
I second this. We plan to use cartopy for a flight planning tool used in remote locations with bad internet connection and want to automatically install it and the necessary data fully before going abroad.

cpatrizio88:
@jypeter I just encountered the same issue working on a supercomputer. Did you happen to figure out a workaround for this?

trexfeathers:
+1 for this. I recently tried to guide a supercomputer user through this issue by debugging and picking apart the various functions. I missed some steps, confusing the situation

So a thorough guide would be very helpful

jypeter:
I had completely forgotten about that question until you guys asked about it yesterday, and I also had to plot a map on my laptop during a conference and it triggered the download on my laptop (WiFi was fortunately working). I have just tried to find out what was downloaded:

- nothing new was installed in the anaconda Python installation (even though I had write access to it)
[...]

PR #2018, opened 2022-03-22
V0.20.x

yaswant:
## Rationale

addresses issue #2017 

## Implications
None


dopplershift:
I'm still on the fence whether we should ship these files, but I'm a definite -1 on including it in the git repository (they do occasionally change). If we do ship them, my preference would be to add the download as a build-time step for generating the wheels and the source distribution in `setup.py`.

greglucas (changes requested):
Yes, I am against putting these into the git repository as well.

I think documentation could go a long way here. There are some good suggestions for how to handle the offline data access in this issue: https://github.com/SciTools/cartopy/issues/1325 that could be brought out into a documentation page to explain the options clearly.

Another thought would be to add a timeout to the download so it doesn't just hang on a system without internet access, but then someone on a poor internet connection will probably complain because the timeout isn't long enough...

Also as a note here is that this only includes the 110m resolution coastlines, but the `ax.coastlines()` call actually adds an auto-scaling feature, so if you're on a limited-extent region it would try to grab `10m` or `50m` coastline features and you'd be back in the same predicament...

yaswant:
@dopplershift , @greglucas  thanks for your comments.  I am not in favour or against including the files in git repo.  Cartopy does include an older version of GSHHS files though, and once can use `GeoAxes.add_feature` method to use them.  So why is it okay to have GSHHS files?  I can see the point when this can still fail (limited extent) - thanks for highlighting that.

We have software stacks deployed on production and hpc systems which are not connected to the world.  There, the coastlines method fails out-of-the-box, which I personally think is not desirable.  I'll take a look at #1325

dopplershift:
@yaswant Well I wasn't around when those data were included, but would have argued against it since it's an external, updated dataset. For instance, the copy in the repo is from 2009, while the official source has renamed to [gshhg](https://www.soest.hawaii.edu/pwessel/gshhg/) and was last updated in 2017.

I'm definitely game to better support use of cartopy in environments without network use. I just think better options are either making it easier to let users dump them in the right spot (or point cartopy to where they already exist), or to only grab them when building packages.

yaswant:
@dopplershift that was my point too.  Recognizing this inconsistence and in anticipation of a reasonable fix, I am closing this PR, but keeping the issue open.

PR #2022, opened 2022-03-25
TST: Skip lakes test for pyshp 2.2

greglucas:
pyshp 2.2 reverses the order of the lakes shapefile, which is not what
we want for this test. They are discussing reverting that change, so
just skip this version for now.

xref: #2012

### DIFF
--- a/lib/cartopy/tests/test_shapereader.py
+++ b/lib/cartopy/tests/test_shapereader.py
@@ -27,6 +27,8 @@ class TestLakes:
             list(self.reader.geometries())[self.lake_index]
         self.test_lake_record = list(self.reader.records())[self.lake_index]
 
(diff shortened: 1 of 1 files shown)


QuLogic:
I was thinking we'd just reverse the `expected` for that version. In #2012, @karimbahgat suggested sorting the coordinates, but that doesn't seem quite right to me as sorting is not holistic. Perhaps we can check `.is_ccw`?

greglucas:
I was just trying to get the tests passing again, and because this seems to be a potentially one-off situation I decided to skip it for this version instead of altering any tests. I'd be fine with a different approach too if you want to PR something.

QuLogic:
I merged #2028, so we can close this.

PR #2023, opened 2022-03-25
TST: Removing unnecessary cleanup decorator

### LINKED ISSUE #2020: Vendor test decorator from MPL?
### Description

Matplotlib may deprecate the `@cleanup` test decorator https://github.com/matplotlib/matplotlib/pull/22697 as it is no longer used in the Matplotlib code base. It was noted that Cartopy uses it, so my question now is where it should be vendored. To me it seems like creating a new file either `testing.py` or `decorators.py` may be options, possibly in the `mpl` directory. Or just add it to `utils.py`.

Any preferences?

(I can provide a PR, just in case.)

greglucas:
This removes the Matplotlib cleanup decorator as we are already handling the cleanup with the autouse fixture.
https://github.com/SciTools/cartopy/blob/22cdafca511744143c5b673598a1889f169c1e3d/lib/cartopy/tests/mpl/conftest.py#L10-L21

closes #2020

### DIFF
--- a/lib/cartopy/tests/mpl/test_axes.py
+++ b/lib/cartopy/tests/mpl/test_axes.py
@@ -6,7 +6,6 @@
 
 from unittest import mock
 
-from matplotlib.testing.decorators import cleanup
 import matplotlib.path as mpath
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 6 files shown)


QuLogic (approved):

greglucas:
Test failures are all unrelated.

PR #2037, opened 2022-04-25
Revert "CI: Avoid newer openssl version due to WMTS tests"

greglucas:
NASA reached out to me indicating they have updated their servers, which means we should be able to use newer openssl versions. Let's test it on CI and see.

This reverts commit 0ecf4d2be92a72f778d03b5bb3073912c3bec38f.

### DIFF
--- a/lib/cartopy/tests/test_shapereader.py
+++ b/lib/cartopy/tests/test_shapereader.py
@@ -9,6 +9,7 @@ import os.path
 import numpy as np
 from numpy.testing import assert_array_almost_equal
 import pytest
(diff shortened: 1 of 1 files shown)


greglucas:
Looks like it works now, openssl 3 is installed in the environment and tests are passing. The failing test is the Lakes shapefile, which has a few PRs open to try and resolve that.

greglucas:
Hm... I didn't close this or merge it, and I don't see this in the history of main at all. I guess I'll push it to a new branch and try again.

PR #2044, opened 2022-05-10
examples: define 'sample_data' for all examples

phil-blain:
Ever since 297e1cb0 (Moving sample_data imports out of cartopy
namespace., 2020-07-23), some examples are not self-contained since they
want to import 'sample_data' from a non-existing module like 'waves' or
'arrows'.

Add the required 'sample_data' function to make all examples
self-contained. For 'contour_labels.py' and 'contour_transforms.py', the
function is copied from 'waves.py'. For 'barbs.py' and 'streamplot.py',
the function is copied from 'arrows.py'.

## Rationale

Make all examples self-contained.

## Implications

Examples can be downloaded and ran individually.

### DIFF
--- a/examples/scalar_data/contour_labels.py
+++ b/examples/scalar_data/contour_labels.py
@@ -8,7 +8,23 @@ An example of adding contour labels to matplotlib contours.
 import cartopy.crs as ccrs
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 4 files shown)


greglucas (approved):
Looks good, thanks for updating these!

PR #2048, opened 2022-05-20
Additional support for LINZ Maps Tiles

aluthfian:
This is to add support for plotting NZ Map tiles, which is in EPSG:3857 (https://epsg.io/3857) already and in png format. Layer ID can be read in every layer's "About" tab, and the "subscription key" parameter is actually an API key given to the registered users. Tested in Python 3.8 and runs well.

## Rationale

To add support to plotting LINZ Map Tiles, which are used widely within the Australia-New Zealand region.

## Implications

More cartopy applicability for Australia-New Zealand users.

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -682,3 +682,19 @@ class AzureMapsTiles(GoogleWTS):
             f'https://atlas.microsoft.com/map/tile?'
             f'api-version={self.api_version}&tilesetId={self.tileset_id}&'
             f'x={x}&y={y}&zoom={z}&subscription-key={self.subscription_key}')
+
+class LINZMapsTiles(GoogleWTS):
+    
+    def __init__(self, subscription_key, layer_id, api_version="v4", desired_tile_form='RGB', cache=False):
+        
+        super().__init__(desired_tile_form=desired_tile_form, cache=cache)
+        self.subscription_key = subscription_key
+        self.layer_id = layer_id
+        self.api_version = api_version
+        
+    def _image_url(self, tile):
+        x, y, z = tile
+        return (
+            f'https://tiles-a.koordinates.com/services;'
+            f'key={self.subscription_key}/tiles/{self.api_version}/'
+            f'layer={self.layer_id}/EPSG:3857/{z}/{x}/{y}.png') 
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -699,4 +699,4 @@ class LINZMapsTiles(GoogleWTS):
         return (
             f'https://tiles-a.koordinates.com/services;'
             f'key={self.subscription_key}/tiles/{self.api_version}/'
-            f'layer={self.layer_id}/EPSG:3857/{z}/{x}/{y}.png') 
+            f'layer={self.layer_id}/EPSG:3857/{z}/{x}/{y}.png')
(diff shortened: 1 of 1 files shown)


lib/cartopy/io/img_tiles.py
```diff
@@ -682,3 +682,19 @@ def _image_url(self, tile):
             f'https://atlas.microsoft.com/map/tile?'
             f'api-version={self.api_version}&tilesetId={self.tileset_id}&'
             f'x={x}&y={y}&zoom={z}&subscription-key={self.subscription_key}')
+
+class LINZMapsTiles(GoogleWTS):
```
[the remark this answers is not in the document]
aluthfian: `
class LINZMapsTiles(GoogleWTS):
    
    def __init__(self, subscription_key, layer_id, api_version="v4",
                 desired_tile_form='RGB', cache=False):
        
        super().__init__(desired_tile_form=desired_tile_form, cache=cache)
        self.subscription_key = subscription_key
        self.layer_id = layer_id
        self.api_version = api_version
        
    def _image_url(self, tile):
        x, y, z = tile
        return (
            f'https://tiles-a.koordinates.com/services;'
            f'key={self.subscription_key}/tiles/{self.api_version}/'
            f'layer={self.layer_id}/EPSG:3857/{z}/{x}/{y}.png')
`

lib/cartopy/io/img_tiles.py
```diff
@@ -682,3 +682,21 @@ def _image_url(self, tile):
             f'https://atlas.microsoft.com/map/tile?'
             f'api-version={self.api_version}&tilesetId={self.tileset_id}&'
             f'x={x}&y={y}&zoom={z}&subscription-key={self.subscription_key}')
+
+
+class LINZMapsTiles(GoogleWTS):
+    
+    def __init__(self, subscription_key, layer_id, api_version="v4",
```
greglucas, line 689: Could you add a description of these and how to obtain a `subscription_key` / link to the main landing page?

greglucas (commented):
Is there a way to add a test for this if we have an environment variable for the `subscription_key` set? See here for a similar situation: https://github.com/SciTools/cartopy/blob/ba25aee6c7868ee376ae7d0a52afea9a315ea623/lib/cartopy/tests/test_img_tiles.py#L287-L295

lib/cartopy/io/img_tiles.py
aluthfian: Hi Greg,

Obtaining a subscription key/API key can be achieved by signing up at https://id.koordinates.com/signup/
(it is not a subscription service per se, just a unique API key for each registered user).
Layer ID is in the "Technical details" section of the "About" tab for each data in the LINZ data service (e.g. https://data.linz.govt.nz/layer/51769-auckland-05m-rural-aerial-photos-2010-2012/ , scroll down for details).
API version is currently v4 and determined manually from the WMTS descriptor.

Fian

greglucas: Thanks! I meant to add that information to the docstring so others can know about how to use this too :)

As a note: Someone with admin access on Cartopy should sign up for some of these services if possible so we can add the keys as secrets to the runners...

aluthfian: Ahh, I see...

I am a bit of a newbie on this Github business, can just let me know if I need to comment on the .py file :-D
All links I have given here are open and do not need any login, however, the contents of WMTS GetCapabilities etc. are for registered users only.

aluthfian:
Yup, done! Thanks for reviewing!

Fian

greglucas (approved):

greglucas:
Thanks, @aluthfian! Sorry for not checking on this in quite a while!

Issue #2049, opened 2022-05-23
Cartopy failling when using DataArray directly 


axnsantana:
### Description
Hi,

Trying to plot some data using xarray's DataArray structure, cartopy fails because it expects (implicitly) the input data to be a numpy array. There is a similar discussion/issue [here](https://github.com/pydata/xarray/issues/2371), [here](https://github.com/numpy/numpy/issues/11772), and a fixing [here](https://github.com/numpy/numpy/pull/11777).

We could follow the Numpy solution and apply it to Cartopy, calling asanyarray on the input data.

#### Code to reproduce

```
import xarray as xr
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import cartopy

w = 3
Y, X = np.mgrid[-w:w:100j, -w:w:100j]
U = -1 - X**2 + Y
V = 1 + X - Y**2
speed = np.sqrt(U**2 + V**2)

lons = np.linspace(0, 5, np.shape(Y)[0])
lats = np.linspace(37, 42, np.shape(Y)[0])

xarray_field = xr.Dataset(
    {
        "u_velocity": (["latitude", "longitude"], U),
        "v_velocity": (["latitude", "longitude"], V),
        "speed": (["latitude", "longitude"], speed)
    },
    coords={
        "longitude": lons,
        "latitude": lats
    })

fig, ax = plt.subplots(nrows=1, ncols=1,
                       subplot_kw={'projection': ccrs.PlateCarree()})
xarray_field.speed.plot(ax=ax)
sp = ax.streamplot(xarray_field["longitude"], xarray_field["latitude"], xarray_field["u_velocity"], xarray_field["v_velocity"], color="w")

plt.show()

```

#### Traceback 

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [5], in <cell line: 4>()
      1 fig, ax = plt.subplots(nrows=1, ncols=1,
      2                        subplot_kw={'projection': ccrs.PlateCarree()})
      3 xarray_field.speed.plot(ax=ax)
----> 4 sp = ax.streamplot(xarray_field["longitude"], xarray_field["latitude"], xarray_field["u_velocity"], xarray_field["v_velocity"], color="w")
      6 plt.show()

File ~/miniconda3/lib/python3.8/site-packages/cartopy/mpl/geoaxes.py:318, in _add_transform.<locals>.wrapper(self, *args, **kwargs)
[...]

PR #2050, opened 2022-05-23
Explicitly calling numpy asanyarray over the inputs.

axnsantana:
Applying asanyarray to x, y, u, v, and all possible scalars to allow the 
use of objects that implements a `__array__`, like `xarray.DataArray`.

## Rationale

Cartopy was falling when non-ndarray inputs like `xarray.DataArray` is used.
More details on issue #2049


greglucas (approved):
For a test, does passing lists to this fail currently with trying to call `[1, 2, 3].ravel()`? That might be a way to test and make sure this handles that case too.

Looks good though, thanks!

greglucas:
Thanks, @axnsantana, hope to see some more contributions from you!

PR #2058, opened 2022-06-28
Fix image tests

dopplershift:
This is fixing failing image tests in CI.

* Two images seem to have updated due to a change in how the dashes in the gridlines align. Not sure why, but it's not really our problem either.
~* We apparently had some missing images in the parametrized tests (#1900) that pytest-mpl is now complaining about. I have no idea if these are actually correct, but they're the current behavior.~
* Force `pytest-mpl!=0.16.0`

Alternatively for the second problem, if anyone knows if the projections make absolutely no sense for the US domain, we could just remove them from the list for that test.


greglucas (changes requested):
The `inline_usa` images are supposed to skipped. They hit the `pytest.skip()` call, but pytest-mpl still compares the images even though it was skipped. I think this is likely a recent update on the pytest-mpl side. I guess I'd suggest we don't add the images and raise an issue upstream for them to fix that use-case.

dopplershift:
Ah, I see it now. I'll force `pytest!=0.16.0`.

greglucas (approved):
Looks good. This needs a rebase now that the other PR went in first. Feel free to self-merge once the conflicts are fixed.

QuLogic (approved):

Issue #1720, opened 2021-01-27
annotate


shaharkadmiel:
It is a bit cumbersome and unintuitive to use the `annotate` method with `GeoAxes` (see [this discussion](https://stackoverflow.com/a/25421922/1579211) on stackoverflow).

To make the `ax.annotate` method more like the other `GeoAxes` methods (like plot, scatter, etc.), I suggest this thin wrapper:

```python
class GeoAxes():
    ...

    def annotate(self, *args, **kwargs):
        crs = kwargs.pop('transform', self.projection)
        transform = crs._as_mpl_transform(self)
        kwargs['xycoords'] = transform
        xy = kwargs['xy']
        return self.annotate(*args, **kwargs)
```

This way the `transform` keyword is used to properly construct the `xycoords` transform.

Example:

```python
ax = plt.axes(projection=ccrs.PlateCarree())
ax.stock_img()

ny_lon, ny_lat = -75, 43
delhi_lon, delhi_lat = 77.23, 28.61

ax.plot(
    [ny_lon, delhi_lon], [ny_lat, delhi_lat], '-bo', transform=ccrs.Geodetic(),
)

ax.annotate('New York', xy=(ny_lon, ny_lat), xytext=(-30, +15),
         textcoords='offset points', ha='center', va='center',
         arrowprops=dict(arrowstyle='->'), 
         transform=ccrs.Geodetic())

ax.annotate('Delhi', xy=(delhi_lon, delhi_lat), xytext=(+30, -15),
         textcoords='offset points', ha='center', va='center',
         arrowprops=dict(arrowstyle='->'), 
         transform=ccrs.Geodetic())
```

![image](https://user-images.githubusercontent.com/6872529/106035050-ee75de00-60d3-11eb-8103-e29e01f3f323.png)

QuLogic:
Seems pretty reasonable, though you have to be careful to only do so if the coords are really for data space; would you like to open a PR?

shaharkadmiel:
Thanks @QuLogic , I can make a pull request.
Indeed, I will have to first check that ``xycoords`` is either  'data' or unset (the default is 'data') and only then assign the transform to it.

PR #2065, opened 2022-08-04
Add functionality for correct transforms in ax.annotate

abrammer:
fixes https://github.com/SciTools/cartopy/issues/1720

## Rationale
Annotate has been "broken" for quite some time (https://stackoverflow.com/questions/25416600/why-the-annotate-worked-unexpected-here-in-cartopy),  and while the work around is somewhat simple, fixing it to be more intuitive also seems fairly straight forward.  Also recently raised in the open issue https://github.com/SciTools/cartopy/issues/1720 which spurred me on to open this PR. 

I'm not particularly aware of underlying stuff here or if this is the "correct" way to fix it, but it seems like a relatively simple fix.  

## Implications
`ax.annotate` will function like most other GeoAxes.methods, but previous work arounds will continue to work.  

## Examples
added a couple simple tests which exemplify the various annotate transform configurations.

Example annotated maps:
annotate_homolosine
annotate_mercator


lib/cartopy/mpl/geoaxes.py
```diff
@@ -1685,6 +1685,46 @@ def scatter(self, *args, **kwargs):
...
+
+        """
+        transform = kwargs.pop('transform', None)
+        is_transform_crs = isinstance(transform, ccrs.CRS)
+
+        # convert CRS to mpl transform for default 'data' setup
+        if is_transform_crs and xycoords == 'data':
+            xycoords = transform._as_mpl_transform(self)
+
+        # textcoords = xycoords be default but complains if xytext is empty
```
greglucas, line 1707:
```suggestion
        # textcoords = xycoords by default but complains if xytext is empty
```

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,131 @@ def test_streamplot():
     ax.streamplot(x, y, u, v, transform=ccrs.PlateCarree(),
                   density=(1.5, 2), color=mag, linewidth=2 * mag)
     return fig
+
+
+@pytest.mark.natural_earth
+@pytest.mark.mpl_image_compare(filename='annotate_mercator.png', style='mpl20')
```
greglucas, line 862:
```suggestion
@pytest.mark.mpl_image_compare(filename='annotate_mercator.png')
```
Do we need the old style for some reason here? Since this is a new test, I'd say just stick with the defaults.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,131 @@ def test_streamplot():
...
+def test_annotate_backwardscompat():
+    """ Work around for annotate to work from
+    https://stackoverflow.com/questions/25416600/why-the-annotate-worked-unexpected-here-in-cartopy/25421922#25421922
+    - check that an mpl_transform is passed through without issue.
```
greglucas, line 866:
```suggestion
    """ Test that annotate can handle an mpl_transform
    https://stackoverflow.com/questions/25416600/why-the-annotate-worked-unexpected-here-in-cartopy/25421922#25421922
    - check that an mpl_transform is passed through without issue.
```

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,131 @@ def test_streamplot():
...
+    """
+    fig = plt.figure(figsize=(10, 5))
+    ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mercator())
+
+    ax.set_extent([65, 125, 5, 40])
+    ax.coastlines()
+    ax.plot(116.4, 39.95, 'ob', transform=ccrs.PlateCarree())
```
greglucas, line 873: I don't think we want to test plot() lines, but rather the annotations. So, I think we can drop all the `plot` and `scatter` calls in the tests.

greglucas (commented):
I think this looks like a reasonable approach.

For the tests, I would suggest combining them all into one with the different cases pointing to different positions, or parameterizing over the transforms with `pytest.mark.parametrize()`, I'm not sure we need a lot of projections to be tested, more so the points/transforms being moved around.

One thing I don't see is the explicit "data" case? If I'm following correctly, that should be the same as the transform passed into the annotate call, and if there is none, default to the map projection.

It might be good to also test that supplying different transforms for `xycoords` and `textcoords` works as expected too, and we aren't accidentally only using one of them.

lib/cartopy/tests/mpl/test_mpl_integration.py
abrammer: Ah, I had just copied the test above, have removed that in the process of consolidating the tests.

lib/cartopy/tests/mpl/test_mpl_integration.py
abrammer: yeah, good call.  Removed the calls that aren't annotate in the consolidated test, except coastlines so we know it's all located with correctly with respect to 1 other feature.

abrammer:
@greglucas Thanks for the input.  Consolidated the tests and added a couple permutations of providing different CRS transforms to `xycoords` and `textcoords`.  Parameterized a couple base projections but not sure it's that's really needed but it's also pretty simple.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,70 @@ def test_streamplot():
     ax.streamplot(x, y, u, v, transform=ccrs.PlateCarree(),
                   density=(1.5, 2), color=mag, linewidth=2 * mag)
     return fig
+
+
+@pytest.mark.natural_earth
+@pytest.mark.parametrize(
+    'map_projection',
+    [ccrs.Mercator(), ccrs.InterruptedGoodeHomolosine(), ccrs.Robinson()]
```
greglucas, line 864:
```suggestion
    [ccrs.Mercator(), ccrs.InterruptedGoodeHomolosine(), ccrs.Robinson()]
```
I guess I don't see a lot of benefit from multiple projections here and the downside is it grows the repo size, so I'd personally vote to keep it at just one of your choice here.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,70 @@ def test_streamplot():
...
+    )
+@pytest.mark.mpl_image_compare()
+def test_annotate(map_projection):
+    """ test a variety of annotate options on mulitple projections
+
+    Annotate defaults to coords passed as if in map projection space.
+    `transform`, `xycoords` or `textcoords` controls location and text offset
+        location. `transform` is a cartopy kwarg so expects a CRS,
+        `xycoords` and `textcoords` accept CRS or matplotlib args.
```
greglucas, line 873: Indent these all to the same amount.

greglucas (commented):
Looking good. Just a few minor comments. I like the combinations you chose for the tests.

lib/cartopy/tests/mpl/test_mpl_integration.py
abrammer: fair enough.  I added homolosine originally to verify that annotations bridged the gaps and extended outside of the map areas correctly.  So, I'll leave that one in, even thought it's a little ridiculous.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,72 @@ def test_streamplot():
     ax.streamplot(x, y, u, v, transform=ccrs.PlateCarree(),
                   density=(1.5, 2), color=mag, linewidth=2 * mag)
     return fig
+
+
+@pytest.mark.natural_earth
+@pytest.mark.parametrize(
+    'map_projection',
+    [ccrs.InterruptedGoodeHomolosine(), ]
+    )
```
greglucas, line 865:
```suggestion
```

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,72 @@ def test_streamplot():
...
+@pytest.mark.mpl_image_compare()
+def test_annotate(map_projection):
```
greglucas, line 867:
```suggestion
def test_annotate():
```

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -856,3 +856,72 @@ def test_streamplot():
...
+    `transform` or `xycoords` & `textcoords` control the marker and text offset
+    through shared or independent projections or coordinates.
+    `transform` is a cartopy kwarg so expects a CRS,
+    `xycoords` and `textcoords` accept CRS or matplotlib args.
+
+    The various annotations below test a variety of the different combinations.
+    """
+
+    fig = plt.figure(figsize=(10, 5))
+    ax = fig.add_subplot(1, 1, 1, projection=map_projection)
```
greglucas, line 880:
```suggestion
    ax = fig.add_subplot(1, 1, 1, projection=ccrs.InterruptedGoodeHomolosine())
```

greglucas (approved):
Looks good to me, just one minor comment left about removing the `parametrize` now that we're only using one projection (sorry about sending you down that path!).

lib/cartopy/tests/mpl/test_mpl_integration.py
abrammer: left the projection as a variable (but not a parameterized one) for the map coord transform further down the test.

abrammer:
yeah, no worries.  It was a simple addition and a simple removal.

greglucas:
Thank you for contributing this, @abrammer! Hope to see you around again.

Issue #1174, opened 2018-11-07
Build wheels and upload to pypi


ngoldbaum:
Currently installing cartopy via pip is somewhat painful, e.g. #1035. This could be ameliorated if cartopy had wheels uploaded on pypi for stable releases.

I think adding the pyproject.toml file will separately mitigate #1035, however uploading wheels will allow many more users, who do not necessarily have compilers set up on their machine, to use cartopy without having to setup a build environment.

I've been pretty happy using https://github.com/matthew-brett/multibuild/ to automate building wheels for yt. I'd be happy to give advice about getting that setup for cartopy.

Note that this came up in the context of us adding catopy as an optional dependency of yt, see https://github.com/yt-project/yt/pull/1966.

dopplershift:
We'd love to. The problem is that CartoPy depends on Shapely, which links to GEOS, and CartoPy directly links to GEOS.  (See #805) If the CartoPy wheels and the Shapely wheels are not linked to the same binary compatible version of GEOS, **BOOM**. (This is not a problem for conda/conda-forge.)

Until this is resolved somehow, providing brittle wheels seems worse than providing none at all.

barronh:
I am curious if there is a way. Shapely Wheels are created using a docker script. Could a docker build have the same docker build as a prerequisite?

greglucas:
wheels are now built and uploaded to PyPI as of v0.22

PR #2069, opened 2022-08-21
Pyproj geodesic trace

greglucas:
This is @QuLogic's work in the first commit, and removes the last remaining remnants of PROJ required for installation. I removed all the mentions of the library and requirements in the second commit.

This will likely be slower (only with geodesic calculations), similar to the previous move to pyproj in v0.20, but I think it will greatly simplify the installation, making up for the loss in speed and moving us one step closer to building wheels...

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -30,12 +30,12 @@ jobs:
         if: matrix.python-version == '3.8' && matrix.os == 'ubuntu-latest'
         id: minimum-packages
         run: |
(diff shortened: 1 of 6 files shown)


dopplershift (approved):
I'm happy to see this go in regardless of performance just to eliminate our PROJ install problems.

Any idea how good our test coverage is for our use of the geodesic stuff? Or more directly, how likely is this to break stuff?

greglucas:
Halfway descent: https://github.com/SciTools/cartopy/blob/main/lib/cartopy/tests/test_geodesic.py

I am also in-favor of this :) and when I was playing around with wheel building, I believe geos was much easier to build than proj, so we could probably even give the wheels a shot before removing the geos dependency as well.

dopplershift:
Wheels are a great idea idea in theory, but we still have the problem of needing to make sure that at all times CartoPy is built against the same version of GEOS as Shapely. #1174 

The next step is to eliminate compiling against GEOS and just rely on Shapely.

PR #2071, opened 2022-08-21
Fix Axes clearing with Matplotlib 3.6+

QuLogic:
## Rationale

Matplotlib 3.6 expired the `cla`/`clear` alias, so `GeoAxes` were not being correctly cleared with Matplotlib 3.6.

## Implications

Plots don't get spectacularly broken.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -575,9 +575,8 @@ class GeoAxes(matplotlib.axes.Axes):
     def __str__(self):
         return '< GeoAxes: %s >' % self.projection
 
-    def cla(self):
-        """Clear the current axes and adds boundary lines."""
-        result = super().cla()
+    def __clear(self):
+        """Clear the current axes and add boundary lines."""
         self.xaxis.set_visible(False)
         self.yaxis.set_visible(False)
         # Enable tight autoscaling.
@@ -593,7 +592,18 @@ class GeoAxes(matplotlib.axes.Axes):
         self.dataLim.intervalx = self.projection.x_limits
         self.dataLim.intervaly = self.projection.y_limits
 
-        return result
+    if mpl.__version__ >= '3.6':
+        def clear(self):
+            """Clear the current Axes and add boundary lines."""
+            result = super().clear()
+            self.__clear()
+            return result
+    else:
+        def cla(self):
+            """Clear the current Axes and add boundary lines."""
+            result = super().cla()
+            self.__clear()
+            return result
 
     def format_coord(self, x, y):
         """
--- a/lib/cartopy/tests/mpl/test_images.py
+++ b/lib/cartopy/tests/mpl/test_images.py
@@ -9,7 +9,6 @@ import types
 
 import numpy as np
 import matplotlib.pyplot as plt
-import matplotlib.cm as cm
 import matplotlib.colors as colors
 from PIL import Image
 import pytest
@@ -143,7 +142,7 @@ def test_imshow_rgba():
     # tests that the alpha of a RGBA array passed to imshow is set to 0
     # instead of masked
     z = np.full((100, 100), 0.5)
-    cmap = cm.get_cmap()
+    cmap = plt.get_cmap()
     norm = colors.Normalize(vmin=0, vmax=1)
     z1 = cmap(norm(z))
     plt_crs = ccrs.LambertAzimuthalEqualArea()
--- a/lib/cartopy/tests/mpl/test_pseudo_color.py
+++ b/lib/cartopy/tests/mpl/test_pseudo_color.py
@@ -92,7 +92,7 @@ def test_pcolormesh_datalim():
     # Z with the same shape as X/Y to force the interpolation
     z = np.zeros(xs.shape)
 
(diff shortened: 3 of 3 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -587,7 +586,18 @@ def cla(self):
...
+            """Clear the current Axes and add boundary lines."""
+            result = super().clear()
+            self._clear()
+            return result
+    else:
+        def cla(self):
+            """Clear the current Axes and add boundary lines."""
+            result = super().cla()
+            self._clear()
+            return result
```
greglucas, line 600: Rather than explicit version gating, what about just defining clear ourselves even for older versions of MPL.

```suggestion
    def clear(self):
        """Clear the current Axes and add boundary lines."""
        result = getattr(super(), "clear", super().cla)()
        self._clear()
        return result

    def cla(self):
        """Clear the current Axes and add boundary lines."""
        return self.clear()
```

tacaswell:
This should be fixed upstream.

dopplershift:
@tacaswell Are you saying upstream needs to put in a fix? Or it already has? I can't find a relevant PR.

tacaswell:
That is a prescriptive "should" as I think this is a release blocker for Matplotlib.

I do not fully understand the issue yet, but I do not think we intended to remove support for either `clear` or `cla`, just make the aliases and always do the same thing!

dopplershift:
@tacaswell That's what I thought, but wasn't sure.

So I just finally made sense of  it too. The problem is that matplotlib itself changed all calls to use `clear()`, and changed `cla()` itself to call `clear()`. That breaks Cartopy because now there's no matplotlib code that will ever call `Geoaxes.cla()`, it's going to look for `Geoaxes.clear()`, which falls back to `Axes.clear()`, which avoids all of Cartopy's code.

tacaswell:
Ah, so we need check in `clear` to see if there is a sub-class defined `cla`, call that and then warn?

dopplershift:
@tacaswell That makes sense to me. @QuLogic ?

Would definitely affect how we proceed here.

QuLogic:
I've put in https://github.com/matplotlib/matplotlib/pull/23735 upstream. I'm not sure that we actually want to warn about it.

However, I _think_ we still want to move forward this this PR, so that we (eventually when Matplotlib 3.6 is the minimum required) are using the new term.

lib/cartopy/mpl/geoaxes.py
QuLogic: `Axes.clear` always existed though, so I don't think that `getattr` fallback will work.

greglucas: Yep I agree, I don't think I realized that it was the coming back down calls that were being missed. It might be helpful to put a quick comment about removing this once we only require MPL 3.6.

greglucas (approved):

dopplershift:
I definitely agree we should go ahead and future proof, I just wasn't sure if the form of the fix here would change based on what's merged for matplotlib 3.6.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -569,9 +569,8 @@ def _update_title_position(self, renderer):
     def __str__(self):
         return '< GeoAxes: %s >' % self.projection
 
-    def cla(self):
-        """Clear the current axes and adds boundary lines."""
-        result = super().cla()
+    def _clear(self):
```
greglucas, line 572:
```suggestion
    def _cartopy_clear(self):
```

and update the calls below.

greglucas (commented):
I just checked out both development branches to test these updates out, and this is definitely close. On Cartopy main I get the expected `PendingDeprecationWarning`, but on this branch there are a bunch of failures. This is due to some more naming conflicts now that Matplotlib added a private `_clear()` method that is called and you're also redefining that here. The easiest fix is probably to update the name here to something explicit and only for Cartopy.

lib/cartopy/mpl/geoaxes.py
tacaswell: we changed to `__clear` upstream to get the name mangling, but this is probably a good idea anyway!

lib/cartopy/mpl/geoaxes.py
```diff
@@ -575,9 +575,8 @@ def _update_title_position(self, renderer):
     def __str__(self):
         return '< GeoAxes: %s >' % self.projection
 
-    def cla(self):
-        """Clear the current axes and adds boundary lines."""
-        result = super().cla()
+    def __clear(self):
```
greglucas, line 578: I don't think this can be the same as mpl's private version. So I _think_ you can go back to `_clear` here now that double  under is in MPL.

QuLogic: Double underscore is class-private; it's mangled automatically to `_GeoAxes__clear`, so it won't clash with Matplotlib's (which is mangled to `_AxesBase__clear`).

greglucas: Ahh, yes of course. This looks good to me now.

greglucas:
I just tested this out locally with rc2 and all worked well for me.

PR #2074, opened 2022-08-31
DOC: What's new for v0.21 release

greglucas:
Preparing for a new v0.21 release by adding a What's New page and documenting what was removed and deprecated in this version.

It doesn't look like we ever settled on moving the docs somewhere else (xref: https://github.com/SciTools/cartopy/issues/1765) , so we will likely have to manually upload them again.

NOTE: Date still needs to be added to the top of the page.

### DIFF
--- a/docs/source/whatsnew/v0.21.rst
+++ b/docs/source/whatsnew/v0.21.rst
@@ -1,0 +1,133 @@
+Version 0.21 (September XXX, 2022)
+==================================
+
+For a full list of included Pull Requests and closed Issues, please see the
+`0.21 milestone <https://github.com/SciTools/cartopy/milestone/30>`_.
+
+Features
+--------
+
+* The requirement to install with a local PROJ installation has been removed.
+  The previous C PROJ library calls have been replaced by pyproj.
+  (:pull:`2069`)
+
+* Many test improvements, including moving to pytest-mpl for image comparisons
+  and parametrizing many tests where possible.
+  (:pull:`1887`, :pull:`1891`, :pull:`1900`)
+
+* The UTF8 degree symbol is now used for latitude and longitude labels.
+  (:pull:`1885`)
+
+* Clément fixed an issue that would ignore the alpha channel when
+  reprojecting RGBA images. (:pull:`1906`)
+
+    .. plot::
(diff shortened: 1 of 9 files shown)


greglucas:
I think this is good to go now, just need to update the release date with whenever we decide to tag a new release.

When building the docs, I noticed that we do get some warnings on the SRTM module, but apparently the docs still build. The warning I think is related to trying to document the inherited class attribute, which seems to be a bug upstream:
https://github.com/sphinx-doc/sphinx/issues/9884

```bash
docs/source/reference/generated/cartopy.io.srtm.SRTM1Source.rst:29: WARNING: autosummary: failed to import SRTM1Source.crs
```
These are also present on the CI builds, so we should probably raise warnings to errors on CI to catch these sooner. It looks like we can just ignore these for now, but we should probably try and figure out a better fix or document them without autosummary.

docs/source/whatsnew/v0.21.rst
```diff
@@ -0,0 +1,133 @@
...
+  now defaults to False. (:pull:`1957`)
+
+* ``geoaxes.add_geometries()`` now accepts both a list of geometries and a
+  single geometry. (:pull:`1999`)
+
+* Better handling of non-ndarray inputs like ``xarray.DataArray``.
+  (:pull:`2050`)
+
+* Alan Brammer added the ability to pass CRS's for the text and xy coordinates
+  used in ``ax.annotate()``.
```
dopplershift, line 55: Add a reference to the PR?

greglucas: Yep, good call.

dopplershift (approved):
Just leaving here until we're pencils down.

greglucas:
@dopplershift, @QuLogic do either of you have any preference for when we release or if anything else needs to go in?

dopplershift:
I don't have anything else that needs to go in. #1951 was mentioned, but I'm ok if that gets punted off.

greglucas:
@dopplershift , I think I got all of the locations with shapely versions and pinned those <2.

I put the release date for tomorrow September 9th. We can move that around still.

dopplershift (approved):

PR #2079, opened 2022-09-11
V0.20.x merge into main

greglucas:
Merge v0.20.x into main. Not sure if this is the best way to do it by creating a separate branch, so let me know if I should do it a different way, or if we even need to do this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,6 @@ Make sure you have pip >= 9.0.1.
 
 
 import os
-import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


dopplershift:
I'm surprised there's any change to merge in, but since there's one that looks useful, I'll merge.

greglucas:
There were plenty of merge conflicts, so it may have been me trying to clean one up that was useful, but wasn't actually truly there...

PR #2080, opened 2022-09-12
Compatibility with shapely 2.0 (remove usage of lgeos)

jorisvandenbossche:
## Rationale

See https://github.com/SciTools/cartopy/issues/2076. 

It is actually relatively easy to remove the usage of `shapely.geos.lgeos`, which is removed in Shapely 2.0, as a first step to make cartopy compatible with Shapely 2.0 (without making use of new features).

I think this should actually be a safe change to include in a 0.21.1 bug-fix release, in case that would be planned.

Now, while this is a necessary change to run the test suite with Shapely 2.0, there are still some other failures, but for things that potentially can be fixed on Shapely's side (see https://github.com/SciTools/cartopy/issues/2076#issuecomment-1243294605 for details)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ deps-run: &deps-install
         pyshp \
         scipy \
         setuptools_scm \
-        'shapely<2' \
+        shapely \
         $EXTRA_PACKAGES \
         --file docs/doc-requirements.txt
     conda list -n test-environment
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Install Shapely dev
         if: matrix.shapely-dev
         run: |
-          python -m pip install git+https://github.com/jorisvandenbossche/shapely.git@compat-geom-factory
+          python -m pip install git+https://github.com/shapely/shapely.git@main
 
       - name: Install Cartopy
         id: install
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Latest packages
         if: steps.minimum-packages.conclusion == 'skipped' && !matrix.shapely-dev
         run: |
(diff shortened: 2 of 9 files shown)


greglucas:
@jorisvandenbossche, thanks for the patch! I agree that this seems reasonable for a patch release. Can you revert all the pins throughout the documentation?
https://github.com/SciTools/cartopy/commit/ad1ec1a4b901556a2111d6f87dbd09d9b4f920a2

Cartopy was not compatible with MPL 3.6 or Shapely 2.0, so we were trying to get ahead of the MPL 3.6 release. I am not opposed to turning this around for a quick v0.21.1 tag to remove the pin on Shapely though so more people can test.

jorisvandenbossche:
Reverted the pins (and all required PRs on the Shapely side are now fixed)

jorisvandenbossche:
The doc build failure seems unrelated?

felixdivo:
The Shapely team just published version [2.0b1](https://github.com/shapely/shapely/releases/tag/2.0b1), so this might be a good point in time to publish this Cartopy improvement and (let the community) test them together. 😄

greglucas (approved):

jorisvandenbossche:
Yes, shapely 2.0b1 should contain all fixes to have this working. 

You could replace the `pip install git+https://github.com/shapely/shapely.git@main` with `pip install --pre`

felixdivo:
Is there an estimate on when the next version of Cartopy will be released? :)

greglucas:
Not yet. We could make a patch release, but I think it'd be good if we settled on https://github.com/SciTools/cartopy/pull/2083 first to try and add that to the release.

felixdivo:
That sounds reasonable. And as long as shapely 2.0 is not stable, this won't cause too many people to ask for cartopy to support it.

jorisvandenbossche:
Just a heads up that I am planning to release a final shapely 2.0 in a few days.

So it would be nice to either have a 0.21patch release with this fix included, or a to too distant timeline for 0.22

jorisvandenbossche:
> Would you mind cherry-picking this commit over and targeting the v0.21.x branch? Then I can merge that and make a new patch release

-> opened https://github.com/SciTools/cartopy/pull/2110

PR #2082, opened 2022-09-14
Remove Google Group link

rcomer:
## Rationale

The Google Group is not much used any more, and over in Iris we have been discussing retiring it (https://github.com/SciTools/iris/issues/3860).  Do you have any objections to that?  Iris have decided to use GitHub Discussions instead.  Possibly Cartopy doesn't need anything to replace it as there seems to be quite a lot of Cartopy activity on StackOverflow.

## Implications

One less place for users to ask questions, but also one less place that support people need to keep on top of.

### DIFF
--- a/.github/SUPPORT.md
+++ b/.github/SUPPORT.md
@@ -1,3 +1,5 @@
  * Ask usage questions on [StackOverflow](https://stackoverflow.com/questions/tagged/cartopy).
+ * For less well defined questions, ideas, general discussion or announcements of related projects use the
+  [Cartopy category on Matplotlib's Discourse](https://discourse.matplotlib.org/c/3rdparty/cartopy/19).
  * Report bugs, suggest features or view the source code on [GitHub](https://github.com/SciTools/cartopy).
  * To chat with developers and other users you can use the [Gitter Chat](https://gitter.im/SciTools/cartopy).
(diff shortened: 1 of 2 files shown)


greglucas (approved):
If Iris is moving away from it, seems like a good idea to remove it from Cartopy's docs.

lbdreyer:
It looks like google groups are also mentioned: https://github.com/SciTools/cartopy/blob/main/.github/SUPPORT.md so that may be worth updating at the same time?

rcomer:
Nice spot @lbdreyer!  I have removed that link and also copied the gitter link from the README, as I assume they should be consistent.

dopplershift:
That's fine with me.

QuLogic:
We should maybe point to the Matplotlib Discourse which has a Cartopy category: https://discourse.matplotlib.org/c/3rdparty/cartopy/19

rcomer:
Since the [scope of the 3rd Party section](https://discourse.matplotlib.org/t/what-are-3rd-party-packages/20666) on Matplotlib's Discourse seems pretty similar to what was previously written for the Google Group, I just reused that text.

QuLogic (approved):

rcomer:
Hi @SciTools/cartopy-devs, is there any more I should do for this one?  The Google Group has now been retired.

dopplershift (approved):

PR #2083, opened 2022-09-14
Remove GEOS dependency in favor of Shapely

greglucas:
This removes the GEOS dependency and pushes the geometry handling into Shapely. It currently takes ~2x longer, so opening this for discussion purposes.

ping: @QuLogic, @jorisvandenbossche

For simple benchmarking I've been making the `project_linear` benchmark executable for quick runs by adding this to the bottom:
```python
if __name__ == "__main__":
    s = Suite()
    p, r = "PlateCarree", "50m"
    s.setup(p, r)
    s.time_project_linear(s, r)
```

Then profiling
```bash
python -m cProfile -o nogeos.prof benchmarks/cases/project_linear.py
snakeviz nogeos.prof
```

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -113,10 +113,6 @@ Further information about the required dependencies can be found here:
     Python package for 2D plotting. Python package required for any
     graphical capabilities.
 
-**GEOS** 3.7.2 or later (https://trac.osgeo.org/geos/)
-    GEOS is an API of spatial predicates and functions for processing geometry
-    written in C++.
-
 **Shapely** 1.7.1 or later (https://github.com/Toblerity/Shapely)
     Python package for the manipulation and analysis of planar geometric
     objects.
--- a/lib/cartopy/tests/test_linear_ring.py
+++ b/lib/cartopy/tests/test_linear_ring.py
@@ -23,25 +23,21 @@ class TestBoundary:
         assert len(multi_line_string.geoms) > 1
         assert not rings
 
-        def assert_intersection_with_boundary(segment_coords):
-            # Double the length of the segment.
-            start = segment_coords[0]
-            end = segment_coords[1]
-            end = [end[i] + 2 * (end[i] - start[i]) for i in (0, 1)]
-            extended_segment = sgeom.LineString([start, end])
-            # And see if it crosses the boundary.
-            intersection = extended_segment.intersection(projection.boundary)
-            assert not intersection.is_empty, 'Bad topology near boundary'
+        def assert_close_to_boundary(xy):
+            # Are we close to the boundary, which we are considering within
+            # a fraction of the x domain limits
+            limit = (projection.x_limits[1] - projection.x_limits[0]) / 1e4
+            assert sgeom.Point(*xy).distance(projection.boundary) < limit, \
+                'Bad topology near boundary'
 
-        # Each line resulting from the split should start and end with a
-        # segment that crosses the boundary when extended to double length.
+        # Each line resulting from the split should be close to the boundary.
         # (This is important when considering polygon rings which need to be
         # attached to the boundary.)
         for line_string in multi_line_string.geoms:
             coords = list(line_string.coords)
             assert len(coords) >= 2
-            assert_intersection_with_boundary(coords[1::-1])
-            assert_intersection_with_boundary(coords[-2:])
+            assert_close_to_boundary(coords[0])
+            assert_close_to_boundary(coords[-1])
 
     def test_out_of_bounds(self):
         # Check that a ring that is completely out of the map boundary
--- a/lib/cartopy/trace.pyx
+++ b/lib/cartopy/trace.pyx
@@ -18,51 +18,20 @@ from functools import lru_cache
 
 cimport cython
 from libc.math cimport HUGE_VAL, sqrt
-from libc.stdint cimport uintptr_t as ptr
+from numpy.math cimport isfinite, isnan
 from libcpp cimport bool
 from libcpp.list cimport list
-from libcpp.vector cimport vector
-from numpy.math cimport isfinite, isnan
-
 
 cdef bool DEBUG = False
-
-cdef extern from "geos_c.h":
-    ctypedef void *GEOSContextHandle_t
-    ctypedef struct GEOSGeometry:
-        pass
-    ctypedef struct GEOSCoordSequence
-    GEOSCoordSequence *GEOSCoordSeq_create_r(GEOSContextHandle_t, unsigned int, unsigned int) nogil
-    GEOSGeometry *GEOSGeom_createLineString_r(GEOSContextHandle_t, GEOSCoordSequence *) nogil
-    GEOSGeometry *GEOSGeom_createCollection_r(GEOSContextHandle_t, int, GEOSGeometry **, unsigned int) nogil
-    GEOSGeometry *GEOSGeom_createEmptyCollection_r(GEOSContextHandle_t, int) nogil
-    GEOSCoordSequence *GEOSGeom_getCoordSeq_r(GEOSContextHandle_t, GEOSGeometry *) nogil
-    int GEOSCoordSeq_getSize_r(GEOSContextHandle_t handle, const GEOSCoordSequence* s, unsigned int *size) nogil
-    int GEOSCoordSeq_getX_r(GEOSContextHandle_t, GEOSCoordSequence *, int, double *) nogil
-    int GEOSCoordSeq_getY_r(GEOSContextHandle_t, GEOSCoordSequence *, int, double *) nogil
-    int GEOSCoordSeq_setX_r(GEOSContextHandle_t, GEOSCoordSequence *, int, double) nogil
-    int GEOSCoordSeq_setY_r(GEOSContextHandle_t, GEOSCoordSequence *, int, double) nogil
-    cdef int GEOS_MULTILINESTRING
 
 import re
 import warnings
 
 import shapely.geometry as sgeom
 import shapely.prepared as sprep
-from shapely.geos import lgeos
-from pyproj import Geod, Transformer
+from pyproj import Geod, Transformer, proj_version_str
 from pyproj.exceptions import ProjError
 import shapely.geometry as sgeom
-
-
-cdef GEOSContextHandle_t get_geos_context_handle():
-    cdef GEOSContextHandle_t handle = GEOS_init_r()
-    return handle
-
-
-cdef shapely_from_geos(GEOSGeometry *geom):
-    """Turn the given GEOS geometry pointer into a shapely geometry."""
-    return sgeom.base.geom_factory(<ptr>geom)
 
 
 ctypedef struct Point:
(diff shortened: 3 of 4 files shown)


greglucas:
> > 2. The most time spent in Shapely seems to be in getting coordinates,

> It might be easier to comment inline on a draft PR, but one idea to make this more efficient is to do src_coords = np.asarray(geometry.coords) (inside project_linear) and then further access coordinates in that numpy array, instead of accessing the coordinates from the coords object (since this has a __getitem__ in python that has some overhead).

Good idea! That reduced it by ~20%. But, we are still creating a lot of Points and Linestrings to do the covers test further on, so we need to think about creating the array of all geometries as soon as possible if we can. This would also greatly help the pyproj transforms too...

dopplershift:
I'd be willing to pay a significant speed penalty (with the hopes of getting it back later) in order to remove the headache that is depending on Shapely while simultaneously linking to GEOS.

lib/cartopy/trace.pyx
```diff
@@ -249,22 +196,14 @@ cdef enum State:
...
-        GEOSCoordSeq_setX_r(handle, coords, 0, point.x)
-        GEOSCoordSeq_setY_r(handle, coords, 0, point.y)
-        g_point = GEOSGeom_createPoint_r(handle, coords)
-        state = (POINT_IN
-                 if GEOSPreparedCovers_r(handle, gp_domain, g_point)
-                 else POINT_OUT)
-        GEOSGeom_destroy_r(handle, g_point)
+        g_point = sgeom.Point((point.x, point.y))
+        state = POINT_IN if gp_domain.covers(g_point) else POINT_OUT
+        del g_point
```
jorisvandenbossche, line 206: To avoid this create-destroy Point overhead, shapely could have some predicates specialized for taking x, y coordinates instead of a Point. 
Coincidentally, GEOS recently added this to their C API as well, so I experimented with this a bit in shapely: https://github.com/shapely/shapely/pull/1548 (and this improvement wouldn't be tied to the GEOS version).

With that, if the user has shapely 2.0 installed, the above could be replaced with:

```
state = POINT_IN if shapely.intersects_xy(gp_domain, point.x, point.y) else POINT_OUT
```

(intersects and covers should be the same in case of polygon/point predicate)

Testing this, this seems to give around 15-20% speedup on the `time_project_linear` benchmark (on my laptop, going from around 6s to 5s)

varchasgopalaswamy:
If it means I don't have to build GEOS, I would take a 2x slowdown any day :)

lib/cartopy/trace.pyx
dopplershift: That'd be nice to have.

greglucas:
After a long delay, I finally got around to pushing up some commits I've had around for a while.

I added a new commit that transforms all of the source points in one call immediately to remove some of the repetitive `transform(single_point)` calls. However, this still only improves the performance by about 20% or so. Reducing the number of calls from 1.2M to 900K, indicating many of the calls are from within the `bisect()` routine itself.

If we want to be more like basemap and transform everything immediately and not interpolate between the transformed points, we can get a factor of 2x improvement. See this quick commit I was using for testing: https://github.com/greglucas/cartopy/commit/c052f63687355147dcfcaee6be51e9e8872192be

My `project_linear` timings were:
Removing GEOS and only using Shapely: 9 seconds
Transforming the points as early as possible: 8 seconds
Removing interpolation/bisection between points: 4 seconds

On the final one where we've removed all interpolation/bisection, the slowest functions are all within Shapely, so we could focus effort in refactoring the shapely geometry handling then.

dopplershift:
I think the bisection/interpolation was an intentional feature of Cartopy, so we shouldn't just blindly remove it. How hard would it be to add an option (somewhere) to disable it?

The 20% sounds gain sounds great, and honestly the win in terms of making this project more sustainable, simplifying packaging, and eliminating a solid 50% of the support questions is worth the current performance impact.

greglucas:
:100: agree on all fronts. I'm hopeful that this can be a short-term hit in performance, but lead to a more sustainable project and getting people to be able to contribute.

> I think the bisection/interpolation was an intentional feature of Cartopy, so we shouldn't just blindly remove it. How hard would it be to add an option (somewhere) to disable it?

I agree we should not remove it (I have some ideas for how to do it better still), that commit is just a demonstration of how it could work and the speed we can gain. Thinking more about it, it is analogous to the `transform_first` keyword argument we added to the contour functions. So, I can see adding that decorator to more functions, or even bringing that up a level and adding a property to `Projection` that would signal us to transform things immediately for the user.

dopplershift:
That seems reasonable.

greglucas:
Any other thoughts from people on this PR and the trade-off of the speed decrease vs ease of installation?

@QuLogic this was mostly your work, so did you have a preference one way or the other on it?

dopplershift:
Just a ping to any interested parties for comment (@QuLogic ) since the release of Shapely 2.0 is imminent.

lib/cartopy/trace.pyx
jorisvandenbossche: So this actually made it into shapely 2.0, and thus could be used (conditionally on the shapely version, if you are OK with adding some different code branches depending on the shapely version)

jorisvandenbossche:
As another speed-up idea, I am wondering: how many of the projections / CRS classes have a `domain` that is rectangular (bounding box) ? Because if a significant portion is limited to that case, I think it would be relatively straightforward to have a fast path for that case for some of the shapely interactions (eg checking if a point is in a bbox is easy and could be done without shapely, avoiding the overhead of shapely in this fast path)

dopplershift:
If someone has plans to immediately implement some speed-ups, I'd entertain waiting. Otherwise, while I'm sure we'll catch some flack for being even slower, I think finally being able to upload wheels and eliminating whole classes of installation issue outweighs that. Maybe it being released slow will motivate some contributions. 😉

greglucas:
> Maybe it being released slow will motivate some contributions

One can dream 😄 (that is my hope too)

ocefpaf:
> I think finally being able to upload wheels

We, conda uses, underestimate the value of this. Cartopy really need wheels ASAP!

greglucas:
Latest commit adds the version gate to avoid the create/destroy Point geometry as @jorisvandenbossche suggested. I wasn't able to see much of a speed-up locally with that though.

There is another LineString create/destroy for the covers()/disjoint() calls that is likely still the slow path.

greglucas:
I've taken a stab at upstreaming some performance updates to pyproj, which would be able to more than offset this (i.e. faster than current main). We are definitely taking a performance hit here, but I still think the benefits are worth it and then we can start whittling away on other performance improvements.

Anyone willing to review/approve and hit the merge button ;) ?

lib/cartopy/tests/test_linear_ring.py
```diff
@@ -23,25 +23,21 @@ def test_cuts(self):
...
-            end = segment_coords[1]
-            end = [end[i] + 2 * (end[i] - start[i]) for i in (0, 1)]
-            extended_segment = sgeom.LineString([start, end])
-            # And see if it crosses the boundary.
-            intersection = extended_segment.intersection(projection.boundary)
-            assert not intersection.is_empty, 'Bad topology near boundary'
-
-        # Each line resulting from the split should start and end with a
-        # segment that crosses the boundary when extended to double length.
+        def assert_close_to_boundary(xy):
```
greglucas, line 26: The latest commit changes this. As far as I can tell we don't extend a line segment by 2x the distance to attach to a boundary anywhere, rather we just project along the boundary to find the closest point. So really we just want to be "close" and not be in the middle of the domain. This is relevant because some segments may be very close to the boundary, but when extended they don't go very far because the points are too closely spaced.

greglucas:
@bjlittle does IRIS have any issues with removing GEOS? Does this slow things down too much, and are there any additional failures that this introduces for all of you?

bjlittle:
> @bjlittle does IRIS have any issues with removing GEOS? Does this slow things down too much, and are there any additional failures that this introduces for all of you?

We've got no direct need for `geos` on `SciTools/iris` :+1:

bjlittle:
> upload wheels

I'm happy to take that on... as long as I wouldn't be stepping on anyone's toes?

bjlittle:
> @bjlittle does IRIS have any issues with removing GEOS? Does this slow things down too much, and are there any additional failures that this introduces for all of you?

@greglucas We also have `asv` benchmarking enabled on `SciTools/iris`, so we'll soon see whether there are any regressions.

@trexfeathers Would it be possible to benchmark against the PR pre-merge? That might be quite informative. Fancy making that happen, if you have capacity?

INSTALL
```diff
@@ -113,9 +113,9 @@ Further information about the required dependencies can be found here:
     Python package for 2D plotting. Python package required for any
     graphical capabilities.
 
-**GEOS** 3.7.2 or later (https://trac.osgeo.org/geos/)
-    GEOS is an API of spatial predicates and functions for processing geometry
-    written in C++.
+**Matplotlib** 3.2 or later (https://matplotlib.org/)
+    Python package for 2D plotting. Python package required for any
+    graphical capabilities.
```
pmav99, line 118: `matplotlib` is duplicated. This changeset can be removed.

greglucas: Yep, thanks for catching that!

varchasgopalaswamy:
As a user I would say < 2x performance hit is more than acceptable given the ease of installation. Also, once it's out in the field it'll be easier to notice where the performance hits are really mattering, right?

dopplershift:
> Does anyone have other thoughts on this PR? What performance penalty is the maximum we would take for easier installations, or what else we would need to consider to keep this moving forward?

In all honesty, given the resources we have for maintaining this project, I think the only requirement I have is that it still works. This is a huge win for packaging and maintainability.

greglucas:
OK, some updated numbers for everyone then.

Running the benchmark from the initial comment (project_linear), I get 1.85s on `upstream/main` and 2.35s on this branch. So, much less than a factor of 2 even. Curiously, it is a bit slower on shapely 2.0 compared to 1.8.5, which probably means there are even more speed-ups to be had if we look at refactoring some of the array coordinate usage throughout the library.

Running the test suite locally I get within a second of each other. On CI it looks like there is ~50% difference between this branch and the most recent merge to main. (Look at the testing time explicitly because the installation time takes forever)
`main`: https://github.com/SciTools/cartopy/actions/runs/5205628895/jobs/9391295722
`this branch`: https://github.com/SciTools/cartopy/actions/runs/5217703945/jobs/9417803995

varchasgopalaswamy:
Sounds more than acceptable to me !

greglucas:
> In all honesty, given the resources we have for maintaining this project, I think the only requirement I have is that it still works. This is a huge win for packaging and maintainability.

It looks like I do have the ability to merge without any approvals. With the lack of people and time to review this PR I think we should push forward and merge this since we have new numpy, scipy, and matplotlib releases recently that it would be nice to push an update out for.

I'll give this 2 more weeks and someone can throw a request changes to block me from self-merging if they have strong opinions against that, otherwise since no one has spoken up adamantly against this I think it is time to move forward with getting this in.

PR #2097, opened 2022-10-22
Add antarctic_ice_shelf_polys to downloaded data

xylar:
## Rationale

These data are quite useful for folks like me doing Antarctic research.  We already include these data in the `cartopy_offlinedata` package on conda-forge.

## Implications

Slightly more downloaded data for folks that don't work on Antarctica.

### DIFF
--- a/tools/cartopy_feature_download.py
+++ b/tools/cartopy_feature_download.py
@@ -42,7 +42,8 @@ FEATURE_DEFN_GROUPS = {
         ('physical', 'geography_regions_polys', ALL_SCALES),
         ('physical', 'geography_regions_points', ALL_SCALES),
         ('physical', 'geography_marine_polys', ALL_SCALES),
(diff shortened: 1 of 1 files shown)


xylar:
## Testing

I made sure this works by running:
```
./tools/cartopy_feature_download.py -o ~/Desktop/cartopy_test gshhs physical cultural cultural-extra
```
As expected, I get the `ne_10m_antarctic_ice_shelves_polys.*` and  `ne_50m_antarctic_ice_shelves_polys.*` files, along with the rest.

xylar:
Here's an example usage:
https://github.com/MPAS-Dev/MPAS-Analysis/blob/develop/mpas_analysis/shared/plot/climatology_map.py#L778-L781
and a typical resulting image (from a very coarse resolution simulation):
image

xylar:
Thanks @dopplershift!

PR #2110, opened 2022-12-07
Compatibility with shapely 2.0 (remove usage of lgeos)

jorisvandenbossche:
The changes of https://github.com/SciTools/cartopy/pull/2080 squashed and applied to v0.21.x branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ deps-run: &deps-install
         pyshp \
         scipy \
         setuptools_scm \
-        'shapely<2' \
+        shapely \
         $EXTRA_PACKAGES \
         --file docs/doc-requirements.txt
     conda list -n test-environment
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -12,6 +12,11 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: [3.8, 3.9, '3.10']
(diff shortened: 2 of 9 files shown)


jorisvandenbossche:
@greglucas any idea if the CI failures are related?

greglucas (approved):
Those test failures are unrelated. I am getting them locally as well. Looks like someone must have changed the XML content or parsing somewhere along the chain with those remote tests. We should really get that XML content and monkeypatch URL to not go to a remote resource when possible... But, lets not have that hold us up here.

Thanks, @jorisvandenbossche!

greglucas:
Actually, it looks like this wasn't enough. I just checked out shapely 2.0 with the Cartopy v0.21.x branch and there are a few failures still. I am not sure it is worth the effort of doing all the backporting on v0.21.x versus just making a new minor release from main.

jorisvandenbossche:
@greglucas which failure do you see? CI here has a build with shapely 2.0rc3, and AFAIK the only 2 failures are the unrelated ones that also happen in the other builds

greglucas:
Uggh, I must have been looking at other Actions and didn't realize this did bring the CI over as well, sorry for the noise! I created a new environment locally and now all is working as expected (with the two unrelated tests failing).

I just pushed a release from this branch, so people should be able to use the new Cartopy and Shapely releases together now.

jorisvandenbossche:
Great! Thanks a lot!

PR #2114, opened 2022-12-13
Address test failures and remove some xfails

greglucas:
* Add an xfail when parsing bad XML for the two recent tests that are failing now.
* owslib 0.20 was released over two years ago now and fixed some of the reasons for test failures so we can remove some xfails that should always pass.

### DIFF
--- a/lib/cartopy/tests/test_img_nest.py
+++ b/lib/cartopy/tests/test_img_nest.py
@@ -234,7 +234,7 @@ def test_nest(nest_from_config):
         assert ('aerial z1 test', img) in nest_z0_z1._ancestry[key]
 
     x1_y0_z1, = (img for img in z1.images
-                 if img.filename.endswith('z_1/x_1_y_0.png'))
+                 if img.filename.endswith('x_1_y_0.png'))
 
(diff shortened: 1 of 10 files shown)


dopplershift (approved):

dopplershift:
It seems our circleCI build is fundamentally broken?

greglucas:
> It seems our circleCI build is fundamentally broken?

I'm not sure what happened there... I think it is back now.

lib/cartopy/tests/test_img_nest.py
```diff
@@ -191,7 +191,6 @@ def world_file_extent(*args, **kwargs):
         return extent, pix_size
 
 
-@pytest.mark.xfail(reason='MapQuest is unavailable')
```
dopplershift, line 194: I think this is the test failing on AppVeyor?

greglucas: Ahh, yes, thanks for catching that! I don't think this is the right xfail. I think it was failing on all Windows installs due to the path separator in the path. Just pushed a new commit to see if that does it.

QuLogic (approved):

PR #2115, opened 2022-12-13
MNT: Remove mentions of GDAL

greglucas:
It isn't used anywhere within Cartopy and can be handled by other libraries.


dopplershift (approved):

QuLogic:
What about `requirements/plotting.txt`?

greglucas:
Oops, I had changed it, but forgot to save the file before committing. Thanks for catching that!

dopplershift (approved):

dopplershift:
I'm not finding any more with grep now.

QuLogic (approved):

amotl:
Hi there,

we appreciate the removal of the GDAL dependency from Cartopy very much. Because we are currently blocked at https://github.com/earthobservations/wetterdienst/pull/878 and https://github.com/earthobservations/wetterdienst/pull/879, it would be so lovely to rebase this work on top of a new Cartopy release, which will not pull in GDAL. 🌻 

With kind regards,
Andreas.

greglucas:
This PR should have been mostly a "documentation" change and not a code/requirement change. GDAL only comes in via `pip install -r requirements/plotting.txt` or through the `environment.yaml`, so my guess is that the GDAL dependency comes from somewhere else for you (Fiona?).

amotl:
Hi Greg,

thank you very much for clarifying. You are right, GDAL does not get pulled in by Cartopy, but by `wradlib-1.18.0` instead, despite it is being listed as _optional_ [^1].

Sorry for the noise and with kind regards,
Andreas.

[^1]: https://github.com/wradlib/wradlib/blob/1.18.0/requirements_optional.txt#L3

PR #2116, opened 2022-12-14
MNT: Use pathlib throughout the codebase

greglucas:
Change `os.path` usages to `pathlib.Path()`

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -17,7 +17,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 from datetime import datetime
-import os
+from pathlib import Path
 import sys
 
 import cartopy
@@ -25,9 +25,8 @@ import matplotlib
 from sphinx_gallery.sorting import ExampleTitleSortKey, ExplicitOrder
 
 # If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+# add these directories to sys.path here.
+sys.path.insert(0, str(Path(__file__).parent.resolve()))
 
 # -- General configuration -----------------------------------------------------
 
--- a/lib/cartopy/__init__.py
+++ b/lib/cartopy/__init__.py
@@ -3,28 +3,27 @@
 # This file is part of Cartopy and is released under the LGPL license.
 # See COPYING and COPYING.LESSER in the root of the repository for full
 # licensing details.
+import os
+from pathlib import Path
 
 from ._version import version as __version__  # noqa: F401
 import tempfile
 
 __document_these__ = ['config']
 
-# Configuration
-import os.path
-
 
 # for the writable data directory (i.e. the one where new data goes), follow
 # the XDG guidelines found at
 # https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
-_writable_dir = os.path.join(os.path.expanduser('~'), '.local', 'share')
-_data_dir = os.path.join(os.environ.get("XDG_DATA_HOME", _writable_dir),
-                         'cartopy')
-_cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
+_writable_dir = Path.home() / '.local' / 'share'
+_data_dir = Path(os.environ.get("XDG_DATA_HOME", _writable_dir)) / 'cartopy'
+_cache_dir = Path(tempfile.gettempdir()) / 'cartopy_cache_dir'
 
-config = {'pre_existing_data_dir': os.environ.get('CARTOPY_DATA_DIR', ''),
+config = {'pre_existing_data_dir': Path(os.environ.get('CARTOPY_DATA_DIR',
+                                        '')),
           'data_dir': _data_dir,
           'cache_dir': _cache_dir,
-          'repo_data_dir': os.path.join(os.path.dirname(__file__), 'data'),
+          'repo_data_dir': Path(__file__).parent / 'data',
           'downloaders': {},
           }
 """
--- a/lib/cartopy/io/__init__.py
+++ b/lib/cartopy/io/__init__.py
@@ -11,7 +11,7 @@ various data formats.
 """
 
 import collections
-import os
+from pathlib import Path
 import string
 from urllib.request import urlopen
 import warnings
@@ -150,8 +150,8 @@ class Downloader:
             expected as a minimum in their ``FORMAT_KEYS`` class attribute.
 
         """
-        return self._formatter.format(self.target_path_template,
-                                      **format_dict)
+        return Path(self._formatter.format(self.target_path_template,
+                                           **format_dict))
 
     def pre_downloaded_path(self, format_dict):
         """
@@ -167,8 +167,9 @@ class Downloader:
             expected as a minimum in their ``FORMAT_KEYS`` class attribute.
 
         """
-        return self._formatter.format(self.pre_downloaded_path_template,
-                                      **format_dict)
+        p = self._formatter.format(self.pre_downloaded_path_template,
+                                   **format_dict)
+        return None if p == '' else Path(p)
 
     def path(self, format_dict):
         """
@@ -193,10 +194,9 @@ class Downloader:
         """
         pre_downloaded_path = self.pre_downloaded_path(format_dict)
         target_path = self.target_path(format_dict)
-        if (pre_downloaded_path is not None and
-                os.path.exists(pre_downloaded_path)):
+        if pre_downloaded_path is not None and pre_downloaded_path.exists():
             result_path = pre_downloaded_path
-        elif os.path.exists(target_path):
+        elif target_path.exists():
             result_path = target_path
         else:
             # we need to download the file
@@ -217,9 +217,9 @@ class Downloader:
             expected as a minimum in their ``FORMAT_KEYS`` class attribute.
 
         """
-        target_dir = os.path.dirname(target_path)
-        if not os.path.isdir(target_dir):
-            os.makedirs(target_dir)
+        target_path = Path(target_path)
+        target_dir = target_path.parent
+        target_dir.mkdir(parents=True, exist_ok=True)
 
         url = self.url(format_dict)
 
--- a/lib/cartopy/io/img_nest.py
+++ b/lib/cartopy/io/img_nest.py
@@ -6,8 +6,7 @@
 
 
 import collections
-import glob
-import os.path
+from pathlib import Path
 
 import numpy as np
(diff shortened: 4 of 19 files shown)


setup.py
```diff
@@ -24,6 +24,7 @@
     sys.exit(error)
 
 
+from pathlib import Path
```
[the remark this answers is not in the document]
greglucas: Do we want to drop the Python 2 crutch and allow the imports to be at the top?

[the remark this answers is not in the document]
QuLogic: I think it's be nice to remove, but maybe we should wait until we have wheels? Then it's generally a given if you're building from source, you need to be certain everything's good.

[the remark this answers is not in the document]
greglucas: Yes, that seems reasonable. I'm not sure when we'll get wheels here... Still a few PRs needed even if/when we remove GEOS.

lib/cartopy/tests/io/test_downloaders.py
```diff
@@ -17,21 +17,21 @@
...
 
     assert di.url(replacement_dict) == 'https://testing.com/example/test.zip'
 
     assert (di.target_path(replacement_dict) ==
-            os.path.join('/wibble', 'foo', 'bar', 'example', 'shape.shp'))
+            Path('/wibble') / 'foo' / 'bar' / 'example' / 'shape.shp')
 
     assert (di.pre_downloaded_path(replacement_dict) ==
-            '/project/foobar/example/sample.shp')
+            Path('/project') / 'foobar' / 'example' / 'sample.shp')
```
QuLogic, line 34: Seems like this could have just stayed as:
```suggestion
            Path('/project/foobar/example/sample.shp'))
```

lib/cartopy/tests/test_coding_standards.py
```diff
@@ -31,9 +32,9 @@
 
 # Guess cartopy repo directory of cartopy - realpath is used to mitigate
 # against Python finding the cartopy package via a symlink.
-CARTOPY_DIR = os.path.realpath(os.path.dirname(cartopy.__file__))
-REPO_DIR = os.getenv('CARTOPY_GIT_DIR',
-                     os.path.dirname(os.path.dirname(CARTOPY_DIR)))
+CARTOPY_DIR = Path(cartopy.__file__).resolve()
```
QuLogic, line 35:
```suggestion
CARTOPY_DIR = Path(cartopy.__file__).parent.resolve()
```

lib/cartopy/tests/test_img_nest.py
```diff
@@ -284,12 +281,12 @@ def wmts_data():
         for sub_tile in aerial.subtiles(tile):
             tiles.append(sub_tile)
 
-    fname_template = os.path.join(_TEST_DATA_DIR, 'z_{}', 'x_{}_y_{}.png')
+    fname_template = str(_TEST_DATA_DIR / 'z_{}' / 'x_{}_y_{}.png')
 
-    if not os.path.isdir(_TEST_DATA_DIR):
-        os.makedirs(_TEST_DATA_DIR)
+    if not _TEST_DATA_DIR.is_dir():
+        _TEST_DATA_DIR.mkdir(parents=True)
```
QuLogic, line 287:
```suggestion
    _TEST_DATA_DIR.mkdir(parents=True, exist_ok=True)
```

lib/cartopy/tests/test_img_nest.py
```diff
@@ -302,17 +299,17 @@ def wmts_data():
...
     for tile in tiles:
         x, y, z = tile
-        fname = fname_template.format(z, x, y)
-        if not os.path.exists(fname):
-            if not os.path.isdir(os.path.dirname(fname)):
-                os.makedirs(os.path.dirname(fname))
+        fname = Path(fname_template.format(z, x, y))
+        if not fname.exists():
+            if not fname.parent.is_dir():
+                fname.parent.mkdir(parents=True)
```
QuLogic, line 312:
```suggestion
            fname.parent.mkdir(parents=True, exist_ok=True)
```

lib/cartopy/io/shapereader.py
```diff
@@ -317,8 +316,8 @@ def zip_file_contents(self, format_dict):
 
         """
         for ext in ['.shp', '.dbf', '.shx', '.prj', '.cpg']:
-            yield ('ne_{resolution}_{name}'
-                   '{extension}'.format(extension=ext, **format_dict))
+            yield Path('ne_{resolution}_{name}'
+                       '{extension}'.format(extension=ext, **format_dict))
```
QuLogic, line 320: Not sure if we should change these, as `zip_file_contents` is technically public and could be overridden?

docs/source/conf.py
```diff
@@ -26,8 +26,8 @@
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+# documentation root.
```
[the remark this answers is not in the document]
greglucas: Removed that sentence.

lib/cartopy/__init__.py
```diff
@@ -3,28 +3,27 @@
...
-_data_dir = os.path.join(os.environ.get("XDG_DATA_HOME", _writable_dir),
-                         'cartopy')
-_cache_dir = os.path.join(tempfile.gettempdir(), 'cartopy_cache_dir')
+_writable_dir = Path.home() / '.local' / 'share'
+_data_dir = Path(os.environ.get("XDG_DATA_HOME", _writable_dir)) / 'cartopy'
+_cache_dir = Path(tempfile.gettempdir()) / 'cartopy_cache_dir'
 
-config = {'pre_existing_data_dir': os.environ.get('CARTOPY_DATA_DIR', ''),
+config = {'pre_existing_data_dir': Path(os.environ.get('CARTOPY_DATA_DIR',
+                                        _data_dir)),
```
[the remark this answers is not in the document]
greglucas: I think I was assuming that should be the fallback, but I agree we don't need to change it here. I've reverted now.

lib/cartopy/io/__init__.py
```diff
@@ -217,9 +217,10 @@ def acquire_resource(self, target_path, format_dict):
             expected as a minimum in their ``FORMAT_KEYS`` class attribute.
 
         """
-        target_dir = os.path.dirname(target_path)
-        if not os.path.isdir(target_dir):
-            os.makedirs(target_dir)
+        target_path = Path(target_path)
+        target_dir = target_path.parent
+        if not target_dir.is_dir():
+            target_dir.mkdir(parents=True)
```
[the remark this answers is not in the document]
greglucas: Yes, updated all of them. Thanks.

lib/cartopy/io/shapereader.py
```diff
@@ -429,9 +427,10 @@ def zip_file_contents(self, format_dict):
 
         """
         for ext in ['.shp', '.dbf', '.shx']:
-            yield (os.path.join('GSHHS_shp', '{scale}',
-                                'GSHHS_{scale}_L{level}{extension}'
-                                ).format(extension=ext, **format_dict))
+            yield (Path('GSHHS_shp'
+                        / '{scale}'.format(**format_dict)
+                        / 'GSHHS_{scale}_L{level}{extension}'.format(
+                            extension=ext, **format_dict)))
```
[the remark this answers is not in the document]
greglucas: Good catch. I think I meant for it to be `Path('GSHHS_shp') / ...` I've updated this in the new commit.

lib/cartopy/tests/io/test_downloaders.py
greglucas: Yeah, I was confused about whether a windows system would replace those with path separators or not... Or if we even want it to in this case.

On windows is `Path("/foo/bar")` equivalent to `Path("/foo") / "bar"` ?

lib/cartopy/io/shapereader.py
greglucas: OK, I've removed that change now as well and left it returning a string and reworked the path handling for later.

lib/cartopy/io/img_nest.py
```diff
@@ -113,32 +112,25 @@ def world_files(fname):
...
-            if dirname:
-                result = os.path.join(dirname, result)
+            fext_types = [f'.{fext}w', f'.{fext[0]}{fext[-1]}w']
+            fext_types.extend([ext.upper() for ext in fext_types])
+            result = [path.with_suffix(ext) for ext in fext_types]
+
+        def _convert_basename(path):
+            dirname, basename = path.parent, Path(path.name)
+            base, ext = basename.stem, basename.suffix
+            result = dirname / (base.swapcase() + ext)
```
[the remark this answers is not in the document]
greglucas: Looks like `with_stem` came in Python 3.9, so we can't move to that quite yet.
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.with_stem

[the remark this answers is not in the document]
QuLogic: Ah, we can still use `with_name` though:
```suggestion
            result = path.with_name(path.stem.swapcase() + path.suffix)
```

lib/cartopy/io/shapereader.py
```diff
@@ -429,9 +427,9 @@ def zip_file_contents(self, format_dict):
 
         """
         for ext in ['.shp', '.dbf', '.shx']:
-            yield (os.path.join('GSHHS_shp', '{scale}',
-                                'GSHHS_{scale}_L{level}{extension}'
-                                ).format(extension=ext, **format_dict))
+            p = Path('GSHHS_shp') / '{scale}'
+            p /= 'GSHHS_{scale}_L{level}{extension}'
```
QuLogic, line 431: Note that this could be written together:
```suggestion
            p = Path('GSHHS_shp', '{scale}',
                     'GSHHS_{scale}_L{level}{extension}')
```

lib/cartopy/tests/io/test_downloaders.py
QuLogic: `Path` normalizes to system directory separator and those are both equivalent. `PosixPath` always uses `/`. But in any case, Windows accepts both `/` and `\`, so it doesn't matter.

lib/cartopy/io/shapereader.py
greglucas: Thanks, updated!

lib/cartopy/tests/io/test_downloaders.py
greglucas: Good to know, thank you!

lib/cartopy/io/img_nest.py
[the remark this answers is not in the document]
greglucas: Updated, thanks!

QuLogic (approved):
Other than CircleCI in #2140.

PR #2129, opened 2023-01-26
FIX: plot after contour shouldn't shrink extent

### LINKED ISSUE #2128: Strange extent behaviour when latitude value too high
### Description

This problem behaviour was triggered by a downstream bug, so I completely understand if you prefer to ignore this issue!  The bug is that my latitude values were generated with `np.arange`, which resulted in my maximum latitude being slightly greater than 90 degrees.

My code creates a contourf and then adds a box to highlight a particular region.  The extent of the resulting plot shrinks to match the box.   If I remove the "problem line" in the below example, the output image is as expected.  The problem also goes away if the axes is instantiated with `projection=SOURCE_PROJ`.

#### Code to reproduce

```python
import numpy as np

import cartopy.crs as ccrs

SOURCE_PROJ = ccrs.PlateCarree()
TARGET_PROJ = ccrs.PlateCarree(central_longitude=200)

def sample_data(shape=(73, 145)):
    """
    Return ``lons``, ``lats`` and ``data`` of some fake data.
    """
    nlats, nlons = shape
    lats1d = np.linspace(-np.pi / 2, np.pi / 2, nlats)
    lons1d = np.linspace(0, 2 * np.pi, nlons)
    lons, lats = np.meshgrid(lons1d, lats1d)

    lats1d = np.rad2deg(lats1d)
    lons1d = np.rad2deg(lons1d)
    data = wave + mean

    lats1d[-1] += 0.00001  # <-- PROBLEM LINE

    return lons1d, lats1d, data


def latlon_region_plot(lat_range, lon_range, ax):
    """
    Add a box described by lat_range and lon_range to the geoaxes.
    """
    ax.plot(
        [lon_range[0], lon_range[1], lon_range[1], lon_range[0], lon_range[0]],
        [lat_range[0], lat_range[0], lat_range[1], lat_range[1], lat_range[0]],
        transform=SOURCE_PROJ, color='magenta')


lons, lats, data = sample_data()
fig = plt.figure()
ax = fig.add_subplot(projection=TARGET_PROJ)

cf = ax.contourf(lons, lats, data, transform=SOURCE_PROJ)
ax.coastlines()
latlon_region_plot([-5, 25], [110, 140], ax)

plt.show()
```

#### Actual Result
extent_box_bad

#### Expected Result
extent_box_good

<details>

### Operating system
RHEL7

### Cartopy version
0.20.3, 0.21.0, 0.21.1

The problem does not appear with 0.19.

### conda list

```
```

### pip list

```
```

</details>

rcomer:
## Rationale

Fixes #2128.

We already have the `ignore=False` within `update_from_data_xy` so that the contour call itself doesn't shrink the extent.  The equivalent line for `plot` is

https://github.com/matplotlib/matplotlib/blob/d731d48b2431da0cf8d7edaa687f997857b112af/lib/matplotlib/axes/_base.py#L2366-L2368

which relies on `ax.ignore_existing_data_limits` to decide whether to shrink.  Under normal circumstances, this flag is set within [update_datalim](https://github.com/matplotlib/matplotlib/blob/d731d48b2431da0cf8d7edaa687f997857b112af/lib/matplotlib/axes/_base.py#L2485-L2508), which is called by Matplotlib's `contour`.  However, for the combination of transform and invalid latitude in the linked issue, the `xys` that gets passed to `update_datalim` is all nans.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1620,6 +1620,9 @@ class GeoAxes(matplotlib.axes.Axes):
         if bboxes:
             extent = mtransforms.Bbox.union(bboxes)
             self.dataLim.update_from_data_xy(extent.get_points(), ignore=False)
+            # Ensure newly set limits aren't ignored by subsequent plotting
+            # functions.
+            self.ignore_existing_data_limits = False
 
         self.autoscale_view()
 
@@ -1668,6 +1671,9 @@ class GeoAxes(matplotlib.axes.Axes):
         if bboxes:
             extent = mtransforms.Bbox.union(bboxes)
             self.dataLim.update_from_data_xy(extent.get_points(), ignore=False)
+            # Ensure newly set limits aren't ignored by subsequent plotting
+            # functions.
+            self.ignore_existing_data_limits = False
 
         self.autoscale_view()
 
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1619,10 +1619,7 @@ class GeoAxes(matplotlib.axes.Axes):
                   if col.get_paths()]
         if bboxes:
             extent = mtransforms.Bbox.union(bboxes)
-            self.dataLim.update_from_data_xy(extent.get_points(), ignore=False)
-            # Ensure newly set limits aren't ignored by subsequent plotting
-            # functions.
-            self.ignore_existing_data_limits = False
+            self.update_datalim(extent.get_points())
 
         self.autoscale_view()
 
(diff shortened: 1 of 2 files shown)


rcomer:
The Ubuntu py3.8 tests seem to be picking up an older version of `proj` 😕

greglucas (commented):
Thanks for the clear explanation, I think I'm following now!

Should we actually be using this Axes method instead? Which I think basically does what you're doing here.
https://github.com/matplotlib/matplotlib/blob/d772e0d4306eac3a1b14f9099711f6261153a1f0/lib/matplotlib/axes/_base.py#L2485-L2508

On the tests... :( We pin to that version of proj/pyproj, but it is odd that it is failing on only a couple of tests.

rcomer:
I looked at the py38 CI on another recent PR, and it has proj-8, but this is getting proj-7.2.

greglucas:
I thought proj was only being used for geodesics at this point, but perhaps that is the issue... I'm still hopeful that we'll rely exclusively on pyproj in the future so we don't have separate pins for the two.

proj8 was released in March 2021, so we could probably pin that high if need be until we get the proj dependence removed: https://github.com/OSGeo/PROJ/releases

It also looks like we need to update something with the sphinx gallery logo for the docs too. Lets address those in separate PRs though.

greglucas:
The failling tests and docs are unrelated and have been addressed in other PRs.

rcomer:
Thanks @greglucas!

PR #2130, opened 2023-01-26
TST/FIX: Test for instance rather than string in class representation

greglucas:
Testing with the pre-release MPL 3.7 wheel has added one new test failure.

Matplotlib has merged SubplotBase into AxesBase now, which makes the class string representation here GeoAxes now, even though we are still an _instance_ of `GeoAxesSubplot`.

xref: https://github.com/matplotlib/matplotlib/pull/23573


dopplershift (approved):

dopplershift:
Failures look unrelated, but need to be fixed at some point.

PR #2096, opened 2022-10-19
Oblique Mercator class

trexfeathers:
## Rationale

Iris users are requesting (SciTools/iris#4830) support for the Oblique Mercator projection, in line with the CF Conventions. This will allow them to quickly and easily plot data from NetCDF files that use this projection.

* [Oblique Mercator in CF Conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_oblique_mercator)
* [Oblique Mercator in PROJ](https://proj.org/operations/projections/omerc.html)

(Code largely copied from `TransverseMercator`, with appropriate modifications).

## Implications

New feature, no other expected implications. Closes #1618.

To-do:

- [x] Tests
- [x] An entry in [`docs/source/reference/projections.rst`](https://github.com/SciTools/cartopy/blob/3d2f67321438b13207b5e5057715e11cbcb70d16/docs/source/reference/projections.rst)
- [x] Resolve the coastline 'folding' bug
  (https://github.com/SciTools/cartopy/pull/2096#issuecomment-1284192774)
- [x] Confirm whether the `x_limits` and `y_limits` settings are appropriate
  (https://github.com/SciTools/cartopy/pull/2096#discussion_r999621607)

## Usage example

```python
from cartopy import crs as ccrs
from matplotlib import pyplot as plt

projection = ccrs.ObliqueMercator()
ax = plt.axes(projection=projection)
ax.coastlines()
ax.gridlines()
plt.show()

```

oblique_mercator

PR #2146, opened 2023-03-12
DOC: add omerc to projections list

bjlittle:
This PR is a follow-up to #2096.

See this [comment](https://github.com/SciTools/cartopy/pull/2096#issuecomment-1464864927).

### DIFF
--- a/docs/source/reference/projections.rst
+++ b/docs/source/reference/projections.rst
@@ -6,7 +6,6 @@
 
 .. _cartopy_projections:
 
-
 Cartopy projection list
 =======================
 
@@ -60,7 +59,7 @@ AzimuthalEquidistant
     import matplotlib.pyplot as plt
     import cartopy.crs as ccrs
 
-    plt.figure(figsize=(3, 3))
+    plt.figure(figsize=(3.0101, 3))
     ax = plt.axes(projection=ccrs.AzimuthalEquidistant(
                             central_latitude=90))
     ax.coastlines(resolution='110m')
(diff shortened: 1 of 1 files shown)


docs/source/reference/projections.rst
```diff
@@ -60,7 +59,7 @@ AzimuthalEquidistant
     import matplotlib.pyplot as plt
     import cartopy.crs as ccrs
 
-    plt.figure(figsize=(3, 3))
+    plt.figure(figsize=(3.0101, 3))
```
greglucas, line 62: Where did this value come from? Seems like it was OK on the current docs: https://scitools.org.uk/cartopy/docs/latest/reference/projections.html#azimuthalequidistant

Same with the Orthographic update.

bjlittle: Struck me as kinda odd too.

I simply done a `make projection` and this is the automated change

bjlittle: Did you want me to revert that particular auto-change?

greglucas: Ahh, the width is taken from the `x_limits` of each projection, so my guess is this has to do with PROJ updating some of their limits with a version bump at some point...

greglucas (approved):

greglucas:
@bjlittle looks like this one didn't run the pre-commit.ci on the PR since it was in beforehand. Can you also fix the end-of-file in your upcoming PR? Might be a little more work and require a trailing newline in the `make_projection.py` script to be more robust for the future too.

bjlittle:
Awesome, thanks @greglucas

PR #2148, opened 2023-03-12
add pre-commit.ci badge

bjlittle:
This PR adds the `pre-commit.ci` badge to the `README.md` as a follow on to #1934 

Controversially, I've also removed the `autofix_prs` disable in the `.pre-commit-config.yaml` - totally happy to encourage debate on this choice.

Personally, the `pre-commit.ci` autofix does save the exact situation that we're initially in, where we banked a PR that wasn't compliant for all the git-hooks (#2146 - my bad, I should have rebased to include #1934 and the git hooks would have detected non-compliance locally).

Naively, I trust the hooks to do the "right thing" and auto-fix as part of the `pre-commit.ci` workflow. I'd love to hear the flip argument as to why we shouldn't adopt that pattern.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,5 @@
 ci:
+  autofix_prs: false
   autoupdate_schedule: 'quarterly'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,4 @@
 ci:
-  autofix_prs: false
   autoupdate_schedule: 'quarterly'
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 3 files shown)


bjlittle:
@greglucas Can we please, please, please disable `stickler-ci` :pray: 

I'm happy to do that :smile:

.pre-commit-config.yaml
```diff
@@ -1,5 +1,4 @@
 ci:
-  autofix_prs: false
```
rcomer, line 2: > Personally, the pre-commit.ci autofix does save the exact situation that we're initially in, where we banked a PR that wasn't compliant for all the git-hooks

With this flag as false, pre-commit would still show a failure in the PR CI, and then you can choose whether to have it fix automatically by summoning the bot ([example](https://github.com/matplotlib/matplotlib/pull/25081#issuecomment-1455195377)).

bjlittle: Seems like a happy compromise :+1:

bjlittle:
pre-commit.ci autofix

greglucas (approved):

PR #2150, opened 2023-03-12
pre-commit: adopt isort hook

bjlittle:
This PR adds support for the [isort](https://github.com/PyCQA/isort) pre-commit git hook to enforce an opinionated `import` sort order throughout the code-base.

The benefit is that there is one less distraction to consider for devs when reviewing pull-requests. Win :partying_face: 

Reference: #1934

PR #2152, opened 2023-03-12
pre-commit: adopt black hook

bjlittle:
This PR adds support for the [black](https://github.com/psf/black) pre-commit git hook to enforce opinionated code formatting throughout the code-base.

Reference: https://github.com/SciTools/cartopy/pull/1934


bjlittle:
I'll resolve the conflicts in this PR once #2150 is merged

greglucas:
In general, :+1: from me. I also looked into it a couple years ago :) https://github.com/greglucas/cartopy/tree/black (2021) This should get more buy-in from other devs though too.

dopplershift:
I think black turns too much sensibly formatted code into outright garbage. 

With that said, I won't stand in the way. But I'm also not going to be clicking the green button on this.

bjlittle:
GitHub Discussions now offers Polls, where it's super easy to garner opinions from the community.

I was just wondering whether it was worth while enabling GH Discussions for `cartopy` and have a poll to ask whether we should adopt `black` formatting or not?

I'm conscious that `cartopy` already has several community channels, so perhaps another wouldn't be ideal?

@SciTools/cartopy-admins How do you guys typically field community opinion, or isn't it a done thing?

greglucas:
I am one of the semi-indifferent people on Black, neither greatly for nor against it, but probably slightly lean in favor because of a similar sentiment to your comment about removing discussion/debate about formatting:
> What I'm attracted to about black is that it's opinionated (and yup, some of its decisions I don't quite agree with, but in general seems relatively fine to me), applies consistency regardless, but most importantly takes the burden of any formatting decisions/debate out of the equation for the core devs when reviewing pull-requests i.e., one thing less to think about.

I personally don't think we need another avenue for communication. Cartopy doesn't get a ton of use from the current channels available, so adding another one probably wouldn't get much response either is my guess.

bjlittle:
Conflicts resolved :+1:

PR #2154, opened 2023-03-13
add ci manifest check

bjlittle:
This PR adds a GHA workflow to ensure that the `MANIFEST.in` reconciles with the contents of the `sdist` using [check-manifest](https://github.com/mgedmin/check-manifest).

This should give us continued confidence that the `sdist` and binary `wheels` contain all the necessary payload for `cartopy` on PyPI.

Reference: [Including files in source distributions with `MANIFEST.in`](https://packaging.python.org/en/latest/guides/using-manifest-in/)


pyproject.toml
```diff
@@ -7,3 +7,9 @@ requires = [
     "setuptools_scm >= 7.0.0",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.check-manifest]
+ignore = [
+    "lib/cartopy/_version.py",
+    "lib/cartopy/trace.cpp",
```
bjlittle, line 14: Note that, these files are present in the `sdist`, but are not under version control, therefore require to be ignored otherwise the `check-manifest` will rightly complain that there is a mismatch.

bjlittle:
pre-commit.ci autofix

dopplershift (approved):

bjlittle:
:beers:

Issue #2145, opened 2023-03-12
Test failures with PROJ 9.2.0


dotlambda:
### Description
`AssertionError` about arrays not being almost equal in
- `tests/test_crs.py::TestCRS::test_epsg`
- `tests/crs/test_equidistant_conic.py::TestEquidistantConic::test_default`
- `tests/crs/test_equidistant_conic.py::TestEquidistantConic::test_eccentric_globe`
- `tests/crs/test_equidistant_conic.py::TestEquidistantConic::test_ellipsoid_transform`
- `tests/crs/test_sinusoidal.py::TestSinusoidal::test_eccentric_globe`

<details>
<summary> Traceback </summary>

```
============================= test session starts ==============================
platform linux -- Python 3.11.2, pytest-7.2.0, pluggy-1.0.0
Matplotlib: 3.6.2
Freetype: 2.12.1
rootdir: /build/Cartopy-0.21.1, configfile: setup.cfg
plugins: mpl-0.16.1
collecting ... INFO:matplotlib.font_manager:generated new fontManager
collected 777 items / 187 deselected / 590 selected

tests/test_coding_standards.py s                                         [  0%]
tests/test_crs.py .....F....................                             [  4%]
tests/test_crs_transform_vectors.py ........                             [  5%]
tests/test_features.py ......                                            [  6%]
tests/test_geodesic.py .............                                     [  9%]
tests/test_img_nest.py .......                                           [ 10%]
tests/test_img_tiles.py .........                                        [ 11%]
tests/test_img_transform.py ............................................ [ 19%]
......................                                                   [ 23%]
tests/test_line_string.py ..................x                            [ 26%]
tests/test_linear_ring.py ......                                         [ 27%]
tests/test_polygon.py ..............................                     [ 32%]
tests/test_shapereader.py ...                                            [ 32%]
tests/test_util.py ..............................................        [ 40%]
tests/test_vector_transform.py .......                                   [ 41%]
tests/crs/test_albers_equal_area.py ........                             [ 43%]
tests/crs/test_azimuthal_equidistant.py ........                         [ 44%]
tests/crs/test_eckert.py ..............................................  [ 52%]
tests/crs/test_equal_earth.py .....                                      [ 53%]
tests/crs/test_equidistant_conic.py FF.....F                             [ 54%]
tests/crs/test_geostationary.py ....                                     [ 55%]
tests/crs/test_gnomonic.py ...............                               [ 57%]
[...]

PR #2163, opened 2023-03-31
updated thresholds/values to work with PROJ 9.2

kirillrdy:
## Rationale

fixes https://github.com/SciTools/cartopy/issues/2145

## Implications

I cant speak for correctness of values
from release log https://proj.org/news.html#release-notes 
it looks like EPSG and ESRI databases got updated so its probably related

### DIFF
--- a/lib/cartopy/tests/crs/test_equidistant_conic.py
+++ b/lib/cartopy/tests/crs/test_equidistant_conic.py
@@ -10,6 +10,7 @@ Tests for the Equidistant Conic coordinate system.
 
 import numpy as np
 from numpy.testing import assert_almost_equal, assert_array_almost_equal
+import pyproj
 import pytest
 
 import cartopy.crs as ccrs
@@ -23,11 +24,16 @@ class TestEquidistantConic:
                       'y_0=0.0', 'lat_1=20.0', 'lat_2=50.0'}
         check_proj_params('eqdc', eqdc, other_args)
 
+        expected_x = (-22784919.35600352, 22784919.35600352)
+        expected_y = (-10001965.729313632, 17558791.85156368)
+        if pyproj.__proj_version__ >= '9.2.0':
+            expected_x = (-22784919.3559981,  22784919.3559981)
+            expected_y = (-10001965.72931272,  17558791.85157471)
         assert_almost_equal(np.array(eqdc.x_limits),
-                            (-22784919.35600352, 22784919.35600352),
+                            expected_x,
                             decimal=7)
         assert_almost_equal(np.array(eqdc.y_limits),
-                            (-10001965.729313632, 17558791.85156368),
+                            expected_y,
                             decimal=7)
 
     def test_eccentric_globe(self):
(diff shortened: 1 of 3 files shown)


greglucas:
This probably needs a check on PROJ(pyproj) versions within the tests to pass the minimum versions CI as well.

lib/cartopy/tests/crs/test_equidistant_conic.py
```diff
@@ -23,12 +24,20 @@ def test_default(self):
                       'y_0=0.0', 'lat_1=20.0', 'lat_2=50.0'}
         check_proj_params('eqdc', eqdc, other_args)
 
-        assert_almost_equal(np.array(eqdc.x_limits),
-                            (-22784919.35600352, 22784919.35600352),
-                            decimal=7)
-        assert_almost_equal(np.array(eqdc.y_limits),
-                            (-10001965.729313632, 17558791.85156368),
-                            decimal=7)
+        if pyproj.__proj_version__ >= '9.2.0':
```
greglucas, line 27: What are your thoughts about changing the number of decimal places we compare to here? It seems like if we can compare to 2-3+ decimal places that should be good enough for Cartopy's purposes. We are really relying on proj for all the "official" transformations.

lib/cartopy/tests/crs/test_equidistant_conic.py
```diff
@@ -23,12 +24,20 @@ def test_default(self):
...
+            assert_almost_equal(np.array(eqdc.x_limits),
+                                (-22784919.3559981,  22784919.3559981),
+                                decimal=7)
+            assert_almost_equal(np.array(eqdc.y_limits),
+                                (-10001965.72931272,  17558791.85157471),
+                                decimal=7)
```
greglucas, line 33: I would restructure these blocks to something like this to avoid the duplicated `assert_almost_equal`s.
```suggestion
        expected_x = ...
        expected_y = ...
        if pyproj.__proj_version__ >= '9.2.0':
            expected_x = ...
            expected_y = ...
        assert_array_equal(eqdc.x_limits, expected_x)
        assert_array_equal(eqdc.y_limits, expected_y)
```

greglucas (commented):
Looks good, just a few minor comments to consider.

lib/cartopy/tests/crs/test_equidistant_conic.py
kirillrdy: I don't have any strong opinion about number of decimals. I do find it strange that cartopy supports different versions of proj that would lead to very different results (eg boundary of UK) 

I would either not test values from proj at all, since they change between versions OR only support one version of proj that is considered "correct", most likely latest version of proj.

that's just my 2 cents but no need to change anything for this PR

lib/cartopy/tests/crs/test_equidistant_conic.py
kirillrdy: done

kirillrdy:
i see tests are failing, but passing in my setup, I have an other look

kirillrdy:
should be fixed now

lib/cartopy/tests/crs/test_equidistant_conic.py
greglucas: I agree with everything you say here :)

greglucas (approved):

PR #2169, opened 2023-04-23
MNT: bump python versions and dependency minimum versions

rcomer:
## Rationale

Python 3.8 recently fell off the NEP29 schedule, so this PR removes testing with that and also adds it for python 3.11.
https://numpy.org/neps/nep-0029-deprecation_policy.html

I also bumped versions of other packages used in the minimum versions test as I assume bad things can happen if you bump the python version without doing this!  I wasn't sure exactly which versions to use here but mostly went for the oldest that is newer than 2 years.  For shapely it seemed safer to go with the older 1.7.1 (August 2020) as 1.8.x had many patch releases stretching into last year.  Matplotlib 3.4 is also older than 2 years but 3.5 didn't come out till November 2021.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -32,6 +32,7 @@ import matplotlib.spines as mspines
 import matplotlib.transforms as mtransforms
 import numpy as np
 import numpy.ma as ma
+import packaging
 import shapely.geometry as sgeom
 
 from cartopy import config
@@ -44,7 +45,7 @@ import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 
 
-assert mpl.__version__ >= '3.4', \
+assert packaging.version.parse(mpl.__version__).release[:2] >= (3, 4), \
     'Cartopy is only supported with Matplotlib 3.4 or greater.'
 
 # A nested mapping from path, source CRS, and target projection to the
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -44,8 +44,8 @@ import cartopy.mpl.patch as cpatch
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 
 
-assert mpl.__version__ >= '3.1', \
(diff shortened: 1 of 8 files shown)


greglucas (approved):
It looks like there is one `MPL_VERSION < (3, 3)` check in the tests that could be dropped too. Are we also supposed to update the build-system dependencies in `pyproject.toml`?

rcomer:
> Are we also supposed to update the build-system dependencies in pyproject.toml?

That seems likely (I'm by no means an expert on this!)  I updated the cython version here to match the min version tests.  The setuptools version here is from 2018 and setuptools_scm only last year, so not sure they were previously following the ~2 year policy.

I removed the mpl version check from that test, and also found some handling for older mpl versions in the main code.

lib/cartopy/mpl/geoaxes.py
```diff
@@ -1783,9 +1783,7 @@ def _wrap_args(self, *args, **kwargs):
         Handle the interpolation when a wrap could be involved with
         the data coordinates before passing on to Matplotlib.
         """
-        # The shading keyword argument was added in MPL 3.3, so keep
-        # this default updating until we only support MPL>=3.3
-        default_shading = mpl.rcParams.get('pcolor.shading', 'auto')
+        default_shading = mpl.rcParams.get('pcolor.shading')
```
greglucas, line 1786: I'm not sure how to read the above comment... Can we get rid of the default altogether now and assume kwargs contains "shading" now?

```suggestion
```

rcomer: I think the user can just not pass _shading_, or am I misunderstanding the question?

lib/cartopy/mpl/geoaxes.py
```diff
@@ -44,7 +45,7 @@
 from cartopy.mpl.slippy_image_artist import SlippyImageArtist
 
 
-assert mpl.__version__ >= '3.4', \
+assert packaging.version.parse(mpl.__version__).release[:2] >= (3, 4), \
```
rcomer, line 48: I guess mpl v3.10 will ship next year, at which point the string comparison would become problematic.

PR #2166, opened 2023-04-06
ENH: enable passing RGB(A) to polormesh

rcomer:
## Rationale

Since Matplotlib v3.7.0, `pcolormesh` accepts RGBA arrays
https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.7.0.html#pcolormesh-accepts-rgb-a-colors

Cartopy just needs a small tweak to be able to pass these down, and we can adapt [this StackOverflow question](https://stackoverflow.com/questions/70541279/how-to-pcolormesh-rgba-array-with-2d-x-and-y/75950850#75950850) to give

```python
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import numpy as np

np.random.seed(100)

x = np.arange(10, 20)
y = np.arange(0, 10)
x, y = np.meshgrid(x, y)

img = np.random.randint(low=0, high=255, size=(10, 10, 4)) / 255

projection = ccrs.Mollweide()
transform = ccrs.PlateCarree()
ax = plt.axes(projection=projection)
ax.set_global()
ax.coastlines()

ax.pcolormesh(np.linspace(0, 120, 11), np.linspace(0, 80, 11), img, transform=transform)

plt.show()
```
test

Closes #2156 and I think also resolves #2045 and resolves #2171, as the request for `set_array` to handle `None` seems to be a workaround that is not needed since mpl 3.7.  I also note that the `None` handling [is not actually documented in `QuadMesh.set_array`](https://matplotlib.org/devdocs/api/collections_api.html#matplotlib.collections.QuadMesh.set_array).

This also adds test coverage for and fixes #2208.

~I'm not sure of the best way to test this.  Does it just need a smoke test of a `pcolormesh` plot with suitable data?~

PR #2174, opened 2023-04-29
Check for mask in GeoQuadMesh.get_array test

rcomer:
## Rationale

Currently the tests all pass if the `get_array` method is completely removed from `GeoQuadMesh`.  Unfortunately `np.array_equal` ignores the masks.  Since our reference/input data has no mask, the simplest way to account for this is to check the output is also unmasked.  With this change, if you remove the `get_array` method, the test now fails.

(I tripped over this while messing about with #2166 and https://github.com/matplotlib/matplotlib/pull/25027.)


rcomer:
CI failures appear to be unrelated.

QuLogic (approved):

PR #2182, opened 2023-06-03
Add support for RGB and RGBA colors to pcolormesh

kylehofmann:
## Rationale

Matplotlib's `Axes.pcolormesh` allows users to specify three types of data: Scalar data (used together with a color map), RGB data, and RGBA data. Prior to this patch, `GeoAxes.pcolormesh` only supported scalar data. This patch adds support and tests for RGB and RGBA data in `GeoAxes.pcolormesh`.

## Implications

Additional visualization options. I made https://www.reddit.com/r/dataisbeautiful/comments/13vq5pg/oc_earths_magnetic_field/ this way.

Improved compatibility with `Axes.pcolormesh`. Compatibility is better but not perfect; wrapping RGB or RGBA data is not supported because `GeoAxes.pcolormesh` ultimately calls `Axes.pcolor`, which does not support RGB or RGBA data.


rcomer:
Thank you for your contribution @kylehofmann - this seems to be a feature that a lot of people want.  See also #2166, which should support the wrapping but is waiting for https://github.com/matplotlib/matplotlib/pull/25027 which will enable `pcolor` to handle RGB(A).

kylehofmann:
@rcomer I wish I had thought to look for something like #2166! It's certainly more thorough than this patch. I particularly like how you handled the tests! (Though if I could make a suggestion: Instead of having `PARAMETRIZE_PCOLORMESH_WRAP` be `True` or `False`, make it `standard`, `rgb`, or `rgba`. When it's one of the latter two, get the mapped data as you do now; then, if it's `rgb`, set `data = data[..., 0:3]`.) Given that #2166 exists, there doesn't seem to be any reason to accept this PR. Thanks for your time!

rcomer:
Thanks @kylehofmann.  Please feel free to add any suggestions/reviews to #2166.

PR #2183, opened 2023-06-05
Fix information cut off in crs docs

lgolston:
## Rationale
In the core CRS documentation page ([https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.crs.CRS.html#](https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.crs.CRS.html)), there is rather limited information shown since a majority of the notes are simply “New in version 2.2.0.” or in a few cases cut off, e.g. “Make a CRS from:”.

Also, in the mpl docs pages a description of the input parameters are missing in the current 0.21.0 version (see e.g. https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.mpl.gridliner.Gridliner.html).

## Implementation
This is addressed in this PR by creating a new Sphinx template class_with_inherited.rst and directing autosummary to use it in the reference/crs.rst doc page, causing the full docstring to be shown rather than just the first line. To reduce the size of the generated pages for classes that inherit CRS, they are set to use class_without_inherited, which is already being used by reference/matplotlib.rst. That template was modified to show __init__, addressing the missing parameter details.

## Implications
The information identified in cartopy.crs.CRS and cartopy.mpl docs that was missing is now shown

### DIFF
--- a/docs/source/_templates/autosummary/class_with_inherited.rst
+++ b/docs/source/_templates/autosummary/class_with_inherited.rst
@@ -1,8 +1,0 @@
-{{ fullname }}
-{{ underline }}
-
-.. currentmodule:: {{ module }}
-
-.. autoclass:: {{ objname }}
-    :members:
-    :inherited-members:
--- a/docs/source/_templates/autosummary/class_with_inherited.rst
+++ b/docs/source/_templates/autosummary/class_with_inherited.rst
@@ -5,6 +5,5 @@
 
 .. autoclass:: {{ objname }}
     :members:
-    :no-inherited-members:
-    :show-inheritance:
+    :inherited-members:
     :special-members: __init__
--- a/docs/source/_templates/autosummary/class_with_inherited.rst
+++ b/docs/source/_templates/autosummary/class_with_inherited.rst
@@ -6,4 +6,3 @@
 .. autoclass:: {{ objname }}
     :members:
     :inherited-members:
-    :special-members: __init__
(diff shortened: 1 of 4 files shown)


greglucas (approved):
I agree this looks better to not have the pyproj version information within Cartopy.

@rcomer, you've been doing a lot of documentation review on Matplotlib, do you have any comments on this or know why it isn't showing the CircleCI redirect anymore?

QuLogic:
> @rcomer, you've been doing a lot of documentation review on Matplotlib, do you have any comments on this or know why it isn't showing the CircleCI redirect anymore?

We need to add a token, as was done in Matplotlib https://github.com/matplotlib/matplotlib/pull/25672 I was hoping someone from the Met Office would do that

rcomer:
I'm afraid I don't know much at all about sphinx directives.  @tkknight is our local documentation expert so perhaps he has thoughts.

For the redirect issue, I believe @bjlittle is now on the case.

bjlittle:
> We need to add a token, as was done in Matplotlib [matplotlib/matplotlib#25672](https://github.com/matplotlib/matplotlib/pull/25672) I was hoping someone from the Met Office would do that

See https://github.com/SciTools/cartopy/pull/2185

greglucas:
Nice! I think this is an improvement. Do you want the inherited members of `CRS`? I find the `..version` callouts for pyproj versions to be confusing here because I would think it would relate to Cartopy's version numbers.
The other two pages are definitely better this way IMO.

lgolston:
Thanks! Initially I liked showing the inherited methods since that library is not as common/familiar as something like matplotlib. However, I agree the version callouts are a little confusing and don't like emphasizing the various inherited from_ methods since those don't even work in cartopy. Am happy to go ahead with not showing the inherited members here, especially with it now linking to pyproj.crs.CRS at the top of the page.

greglucas:
Looks good to me. Thank you @lgolston!

Issue #2087, opened 2022-09-26
FionaReader fails on Natural Earth 10m rivers for North America. 


juseg:
### Description

`FionaReader` fails to open [Natural Earth](https://github.com/nvkelso/natural-earth-vector) 10m rivers for North America. There seems to be an empty record which raises an `AttributeError`. `BasicReader` provides a record with geometry `None`, allowing e.g. `for rec in shp.records()`. Is this a data bug or a cartopy bug?

#### Code to reproduce

```python

import cartopy

filename = cartopy.io.shapereader.natural_earth(
    resolution='10m', category='physical', name='rivers_north_america')
basic = cartopy.io.shapereader.BasicReader(filename).records()
fiona = cartopy.io.shapereader.FionaReader(filename).records()

while True:
    basic_record = next(basic)  # yields a record with geometry None
    fiona_record = next(fiona)  # raises AttributeError
```

#### Traceback 

```python
Traceback (most recent call last):
  File "/home/julien/git/code/hyoga/fionabug.py", line 12, in <module>
    fiona_record = next(fiona)
  File "/usr/lib/python3.10/site-packages/cartopy/io/shapereader.py", line 244, in records
    yield FionaRecord(item['geometry'],
  File "/usr/lib/python3.10/site-packages/cartopy/io/shapereader.py", line 116, in __init__
    self._bounds = geometry.bounds
AttributeError: 'NoneType' object has no attribute 'bounds
```

<details>
 <summary>Full environment definition</summary>

### Operating system

Manjaro Linux

### Relevant packages

```
Package                       Version
----------------------------- -----------
Cartopy                       0.21.0
Fiona                         1.8.21
pyshp                         2.3.0
```

### Full pip list

```
Package                       Version
----------------------------- -----------
absl-py                       1.2.0
absplots                      0.2.0
affine                        2.3.1
alabaster                     0.7.12
apipkg                        3.0.1
appdirs                       1.4.4
apptools                      5.2.0
arandr                        0.1.10
astroid                       2.12.7
asttokens                     2.0.8
atpublic                      3.1.1
attrs                         22.1.0
autocommand                   2.2.1
autopep8                      1.7.0
Babel                         2.10.3
backcall                      0.2.0
beautifulsoup4                4.11.1
Bottleneck                    1.3.5
btrfsutil                     5.19
build                         0.8.0
CacheControl                  0.12.11
Cartopy                       0.21.0
certifi                       2022.6.15
cf-xarray                     0.7.4
[...]

PR #2188, opened 2023-06-09
Fix bounds error in shapereader for null records

lgolston:
## Rationale

As identified in issue #2087, FionaReader is currently throwing an error when iterating through the records,of the 10m Natural Earth rivers shapefile for North America, due to some entries that have attributes but no geometry. Since geometry is None in those cases, calling geometry.bounds fails. I checked and this is new to version 5.0.0 of the shapefile - the previous 4.1.0 version works without error.

## Implications

The bug should be fixed.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -114,7 +114,11 @@ class FionaRecord(Record):
     def __init__(self, geometry, attributes):
         self._geometry = geometry
         self.attributes = attributes
-        self._bounds = geometry.bounds
+
+        if geometry is not None:
+            self._bounds = geometry.bounds
+        else:
+            self._bounds = None
 
 
 class BasicReader:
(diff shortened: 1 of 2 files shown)


dopplershift:
Is the `bounds` property not used by our codebase at all? It would seem that while your patch fixes the initial problem, it would only delay failure until that property from the base `Record` class is accessed:

https://github.com/SciTools/cartopy/blob/3ebd4a493dcfd90be01d30ca46b0a7ac1542b280/lib/cartopy/io/shapereader.py#L82-L90

lgolston:
That is an interesting question @dopplershift. I am not sure if it is used anywhere, but will look at editing the property definition you have highlighted since looking now it produces an error with both BasicReader and FionaReader on this dataset.

lgolston:
The bounds property of class Record is now modified with the same check in case it is called.

I also looked at test_shapereader, and found it was only testing FionaReader (if fiona installed). The PR now modifies it to test both Readers, improving branch coverage for shapereader.

dopplershift (approved):

dopplershift:
Thanks for the fix @lgolston !

PR #2191, opened 2023-06-11
Enhanced OGC reader

mcuntz:
Added support for general EPSG projections in OGC clients WMSRasterSource, WMTSRasterSource, and WFSGeometrySource.

Try to determine geometries from other map servers than the hard-coded mapserver.gis.umn.edu.

## Rationale

I tried to use the WFS support with a French server, which did not work because:
1. the server used the projection EPSG:2154, and
2. after implementing support for `cartopy.crs.epsg,` it still did not plot my desired features.

So I
1. implemented `cartopy.crs.epsg` as an addition to the known projections. There were only 7 projections known to the OGC readers. Now, all EPSG codes should work.
2. I try to guess the geometries of a given server because the map server was, strangely, hard-coded to be `http://mapserver.gis.umn.edu/mapserver` and the geometries were named `msGeometry`. I search the basic url in the features now and assume that the geometries have geom or Geom in their name. This is not perfect but I do not know any better; this is the first time I use web services for shapes.
3. While implementing another test case in mpl/test_features.py, I also changed the test case for natural earth from     `ax.add_feature(cfeature.BORDERS, linestyle=':')` to `ax.add_feature(cfeature.BORDERS, linestyle='-')` because this failed on my machine only due to slightly different dots. Should be more robust know.

## Implications

I stayed with the known projections in _URN_TO_CRS and simply added EPSG on top. I had first used the information in `cartopy.crs.epsg` but the information in EPSG:3031 (Antarctica) was weird and wrong in my opinion.

Guessing the geometries is a layman coding but should be o.k. until it does not work with another server ;-)


mcuntz:
Just took back the change `ax.add_feature(cfeature.BORDERS, linestyle=':')` to `ax.add_feature(cfeature.BORDERS, linestyle='-')`  for the test case of natural earth because tests fail on github otherwise.

mcuntz:
flake8 checks fail on
```
            if ((str(default_urn) not in _URN_TO_CRS) and
                (':EPSG:' not in str(default_urn))):
                raise ValueError(
                    f'Unknown mapping from SRS/CRS_URN {default_urn!r} to '
                    'cartopy projection.')

```
with
`lib/cartopy/io/ogc_clients.py:747:17: E129 visually indented line with same indent as next logical line`
If I put in some spaces like
```
            if ( (str(default_urn) not in _URN_TO_CRS) and
                 (':EPSG:' not in str(default_urn)) ):
                raise ValueError(
                    f'Unknown mapping from SRS/CRS_URN {default_urn!r} to '
                    'cartopy projection.')

```
flake8 also fails complaining that there is a space after '('.

I don't know how to resolve this :-(

akrherz:
Try this
```
    if (str(default_urn) not in _URN_TO_CRS) and (
        ":EPSG:" not in str(default_urn)
    ):
        raise ValueError(
            f"Unknown mapping from SRS/CRS_URN {default_urn!r} to "
            "cartopy projection."
        )
```

lib/cartopy/tests/mpl/test_features.py
```diff
@@ -70,3 +70,17 @@ def test_wfs():
...
+
+
+@pytest.mark.network
+@pytest.mark.skipif(not _OWSLIB_AVAILABLE, reason='OWSLib is unavailable.')
+@pytest.mark.xfail(raises=ParseError,
+                   reason="Bad XML returned from the URL")
+@pytest.mark.mpl_image_compare(filename='wfs_france.png')
+def test_wfs_france():
+    ax = plt.axes(projection=ccrs.epsg(2154))
+    url = 'https://agroenvgeo.data.inra.fr:443/geoserver/wfs'
```
greglucas, line 82: I'm wondering if it would be beneficial for us to start mocking servers rather than relying on external sources? Basically, if it'd be reasonable to add both the XML you download from there and the image to the repository and then make a Mock response from that url. Then we wouldn't need these network marks anymore.

This doesn't have to be undertaken here, it is more me just wondering how reliable these external servers are and if there is a way around hitting them constantly if they don't like that.

dopplershift: We could use vcrpy to record and playback http traffic. That would be a low-cost way to "mock" it without questioning whether there's value in having the mock. It gets messy since the only true measure of "works" is "does it communicate with this resources that's outside our control".

mcuntz: The pytest plugin for [VCR](https://github.com/kiwicom/pytest-recording) looks like an easy solution. However, I leave this to the main developers because it adds new dependencies to cartopy testing.

lib/cartopy/io/ogc_clients.py
```diff
@@ -666,7 +696,17 @@ def __init__(self, service, features, getfeature_extra_kwargs=None):
             raise ImportError(_OWSLIB_REQUIRED)
 
         if isinstance(service, str):
+            # base url without http:// or https://, and :port
```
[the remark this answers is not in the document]
mcuntz: Thanks. I use urlparse now and added comments.

lib/cartopy/io/ogc_clients.py
```diff
@@ -574,7 +588,24 @@ def _wmts_images(self, wmts, layer, matrix_set_name, extent,
...
+            meters_per_unit = METERS_PER_UNIT[tile_matrix_set.crs]
+        elif ':EPSG:' in tile_matrix_set.crs:
+            epsg_num = tile_matrix_set.crs.split(':')[-1]
+            tms_crs = ccrs.epsg(int(epsg_num))
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
+                crs_dict = tms_crs.to_dict()
+                crs_units = ''
+                if 'units' in crs_dict:
+                    crs_units = crs_dict['units']
```
greglucas, line 601:
```suggestion
                crs_units = crs_dict.get('units', '')
```

lib/cartopy/io/ogc_clients.py
```diff
@@ -574,7 +588,24 @@ def _wmts_images(self, wmts, layer, matrix_set_name, extent,
...
         tile_matrix_set = wmts.tilematrixsets[matrix_set_name]
         tile_matrices = tile_matrix_set.tilematrix.values()
-        meters_per_unit = METERS_PER_UNIT[tile_matrix_set.crs]
+        if tile_matrix_set.crs in METERS_PER_UNIT:
+            meters_per_unit = METERS_PER_UNIT[tile_matrix_set.crs]
+        elif ':EPSG:' in tile_matrix_set.crs:
+            epsg_num = tile_matrix_set.crs.split(':')[-1]
+            tms_crs = ccrs.epsg(int(epsg_num))
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
```
greglucas, line 597: What is the warning we are ignoring? We should at least mention why this is here with a comment.

lib/cartopy/io/ogc_clients.py
```diff
@@ -807,6 +869,36 @@ def _to_shapely_geoms(self, response):
                 data = self._find_polygon_coords(node, find_str)
                 points_data.extend(data)
 
+        # Try to get other geometries from other servers than
+        # http://mapserver.gis.umn.edu/mapserver
```
greglucas, line 873: Is there a way to add an extra line or two above to make that loop work for your case? i.e. can you add something above to change the `tree.findall()` call like updating the `_MAP_SERVER_NS` or just adding an extra block to find all of your geom tags?

This just reads as a lot of copied code, so another option would be to make a function that both loops could use too. Something like this to update your geometry lists that could be re-used in both loops.

`linear_rings_data, linestrings_data, point_data = _get_geometries(node, linear_rings_data, linestrings_data, point_data)

lib/cartopy/io/ogc_clients.py
```diff
@@ -22,6 +22,7 @@
 import warnings
 import weakref
 from xml.etree import ElementTree
+from urllib.parse import urlparse
```
greglucas, line 25: Should be alphabetically sorted. See the pre-commit hook warning.

lib/cartopy/io/ogc_clients.py
mcuntz: Yes, this is a UserWarning from .to_dict() from proj because the dict does not give all details of each projection. But we only check for the units here.
I added a comment.

lib/cartopy/io/ogc_clients.py
mcuntz: O.K. I was trying to not change existing code. Now it is included in the same loop.

greglucas (approved):
Looks good, thanks, @mcuntz!

PR #2194, opened 2023-06-18
Avoid a deprecation warning from numpy 1.25 in crs._safe_pj_transform

neutrinoceros:
## Rationale

Numpy 1.25 (released yesterday) [deprecates implicit conversions from 1-sized arrays to scalar types](https://github.com/numpy/numpy/pull/10615)

While testing yt + cartopy, I discovered that such implicit conversion may be performed in `pyproj.transformer.Transformer.transform`, see https://github.com/pyproj4/pyproj/blob/32565ddf266658aebc9787b7534fdbdd06762839/pyproj/_transformer.pyx#L762

In order to avoid triggering deprecation warnings, the call site (here in cartopy) should handle this special case (pyproj cannot do it itself because it doesn't depend on numpy).

## Implications

I assume there are already tests in cartopy that hit this condition, but properly testing this requires treating warnings as errors. I don't think this is done in cartopy yet, and it may require considerable efforts, way beyond the intended scope of this PR. Let me know what you guys think !


greglucas:
Is there a minimal reproducer with just Cartopy? When upgrading to numpy 1.25 and trying with scalars directly as below I don't get a deprecation warning.

```python
In [1]: import cartopy.crs as ccrs

In [2]: import numpy as np

In [3]: p = ccrs.PlateCarree()

In [4]: p2 = ccrs.Mercator()

In [5]: p2.transform_point(np.float64(1), np.float64(2), p)
```

I'm not sure if this is best addressed in Cartopy or upstream in yt. i.e. could yt do this same conversion?

neutrinoceros:
In fact your example is already a minimal reproducer, you just need to specifically treat warnings as errors to see it. Wrapping your example in a script
```python
# t.py
import cartopy.crs as ccrs
import numpy as np

p = ccrs.PlateCarree()
p2 = ccrs.Mercator()

p2.transform_point(np.float64(1), np.float64(2), p)
```
and invoking as `python -Werror t.py` will trigger the warning.


> I'm not sure if this is best addressed in Cartopy or upstream in yt. i.e. could yt do this same conversion?

I think this is out of reach for yt because the problematic data (scalar arrays) seems to be generated from within cartopy. We're seeing an error (warning) specifically with the `LambertAzimuthalEqualArea` transform. There may be a better fix than my first proposed solution (maybe somewhere in `cartopy.crs.CRS.transform_point(s)`) but I haven't been able to find another yet.

neutrinoceros:
actually the reproducer can be reduced even further
```python
# t.py
import cartopy.crs as ccrs

p = ccrs.PlateCarree()
p2 = ccrs.Mercator()

p2.transform_point(1, 2, p)
```

(the problem is with size 1 *arrays*, not actual numpy scalars, and the conversion from scalar to array is done internally)

neutrinoceros:
I've included the reprod as a test

neutrinoceros:
Sounds reasonable. I'll upstream the discussion to pyproj, and I'll come back to close this when I get things moving forward there, if that's okay with you.

greglucas:
Yes, that sounds totally reasonable :) feel free to ping me over there too since I am the one to blame for this "feature" in pyproj.

neutrinoceros:
I think the future error to be raised in NumPy 2.0 (or whenever the deprecation expires), is indeed `TypeError`, but for security I asked for confirmation in https://github.com/numpy/numpy/pull/10615#issuecomment-1601325290

neutrinoceros:
The future is confirmed to be `TypeError`, so I think there's nothing to be done in pyproj and we can just live with a warning filter. My understanding is that cartopy doesn't even need such a filter at the moment, or does it ?

neutrinoceros:
@greglucas so do you think there's anything that needs be done here ?

lib/cartopy/crs.py
```diff
@@ -53,7 +53,22 @@ def _safe_pj_transform(src_crs, tgt_crs, x, y, z=None, trap=True):
...
+        # future compatibility with numpy:
+        # pyproj implicitly converts size-1 arrays to scalars, which is
+        # deprecated in numpy 1.25
+        # see https://github.com/numpy/numpy/pull/10615
+        if isinstance(a, np.ndarray) and a.size == 1:
+            return a.item()
+        else:
+            return a
+
+    return transformer.transform(
```
greglucas, line 66: What about putting a context manager around a `filterwarnings` call here instead? The warning is sort of unfortunate and pyproj already has the workarounds for handling the single-item numpy arrays, so I guess I'd say we can just sweep the warning under the rug here because it is downstream and already taken care of.

```python
warnings.filterwarnings("ignore", message="Conversion of an array with ndim > 0")
```

lib/cartopy/tests/crs/test_transform_point.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# This file is part of Cartopy and is released under the LGPL license.
+# See COPYING and COPYING.LESSER in the root of the repository for full
+# licensing details.
+import warnings
+
+import cartopy.crs as ccrs
+
+
+def test_transform_point():
```
greglucas, line 11:
```suggestion
def test_transform_point_no_warning():
# Make sure we aren't warning on single-point numpy arrays
```
Can you move this over to `tests/test_crs.py` which already has a bunch of point transformation tests?

greglucas (commented):
Yeah, I think we should do something about this, and practicality beats purity here I think. I have one further suggestion to get your thoughts on for just putting a warnings filter in rather than getting single items out.

lib/cartopy/tests/crs/test_transform_point.py
neutrinoceros: sure thing

lib/cartopy/crs.py
neutrinoceros: Let's try it

greglucas (approved):
Looks good to me. Thanks!

PR #1916, opened 2021-10-29
Plotting Bathymetry shapefiles

mattphysics:
## Rationale

I have come across the question of how to plot ocean bathymetry in cartopy. This led me to write a short notebook of how to do that using the Shapefiles from Natural Earth. I was encouraged to share this so others can benefit.
Note that this is my first pull request on here, and I'm not sure this is exactly the right place. Please advise if appropriate.

## Implications

No implications

PR #2195, opened 2023-06-20
Create ocean bathymetry example

lgolston:
## Rationale

In a previous pull request (#1916) @mattphysics created a notebook for plotting ocean bathymetry using Natural Earth data, but it was not finished. I have here converted it to converted it to a script, reduced the length in some places, and propose adding it to the example gallery.
[](url)

## Implications

The shapereader.natural_earth interface does not work on this particular file, since the .zip contains multiple shapefiles corresponding to different depths. I think it is a nice example using cartopy's shapereader.Reader interface. The axis extent is set to [90, 160, -15, 60] to help speed up the script versus plotting the whole world.
test

### DIFF
--- a/examples/lines_and_polygons/ocean_bathymetry.py
+++ b/examples/lines_and_polygons/ocean_bathymetry.py
@@ -4,7 +4,7 @@ Ocean bathymetry
 
 Produces a map of ocean seafloor depth, demonstrating the
 :class:`cartopy.io.shapereader.Reader` interface. The data is a series
-of 10m resolution nested polygons obtained from Natural Earth, and derived
+of 10m resolution nested polygons obtained from Natural Earth, derived
 from the NASA SRTM Plus product. Since the dataset contains a zipfile with
 multiple shapefiles representing different depths, the example demonstrates
 manually downloading and reading them with the general shapereader interface,
@@ -51,16 +51,18 @@ def load_bathymetry(zip_file_url):
 
 if __name__ == "__main__":
     # Load data (14.8 MB file)
-    depths, shp_dict = load_bathymetry(
+    depths_str, shp_dict = load_bathymetry(
         'https://naturalearth.s3.amazonaws.com/' +
         '10m_physical/ne_10m_bathymetry_all.zip')
 
     # Construct a discrete colormap with colors corresponding to each depth
-    bounds = sorted(depths.astype(float))  # low to high
-    N = len(bounds)
-    norm = matplotlib.colors.BoundaryNorm(bounds, N)
+    depths = depths_str.astype(int)
+    N = len(depths)
+    nudge = 0.01  # shift bin edge slightly to include data
+    boundaries = [min(depths)] + sorted(depths+nudge)  # low to high
+    norm = matplotlib.colors.BoundaryNorm(boundaries, N)
     blues_cm = matplotlib.colormaps['Blues_r'].resampled(N)
-    colors_depths = blues_cm(norm(depths.astype(float)))
+    colors_depths = blues_cm(norm(depths))
 
     # Set up plot
     subplot_kw = {'projection': ccrs.LambertCylindrical()}
@@ -68,8 +70,8 @@ if __name__ == "__main__":
     ax.set_extent([90, 160, -15, 60], crs=ccrs.PlateCarree())  # x0, x1, y0, y1
 
     # Iterate and plot feature for each depth level
-    for i, depth in enumerate(depths):
-        ax.add_geometries(shp_dict[depth].geometries(),
+    for i, depth_str in enumerate(depths_str):
+        ax.add_geometries(shp_dict[depth_str].geometries(),
                           crs=ccrs.PlateCarree(),
                           color=colors_depths[i])
 
@@ -85,10 +87,9 @@ if __name__ == "__main__":
     sm = plt.cm.ScalarMappable(cmap=blues_cm, norm=norm)
     fig.colorbar(mappable=sm,
                  cax=axi,
-                 boundaries=depths.astype(int)[::-1],
                  spacing='proportional',
                  extend='min',
-                 ticks=depths.astype(int),
+                 ticks=depths,
                  label='Depth (m)')
 
     # Convert vector bathymetries to raster (saves a lot of disk space)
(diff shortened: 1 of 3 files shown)


lgolston:
This time the page built successfully on circleci: [bathymetry example](https://output.circle-artifacts.com/output/job/1deaa86b-5acc-4c50-a127-57edf2b68bfb/artifacts/0/docs/build/html/gallery/scalar_data/ocean_bathymetry.html#sphx-glr-gallery-scalar-data-ocean-bathymetry-py)

lgolston:
Several changes: I realized this fits better under lines_and_polygons, so moved the file there. There is also now more text at top introducing the example, and the colorbar now has an arrow to indicate the depth can go below -10,000 m. [Link to current version](https://output.circle-artifacts.com/output/job/8577cf45-16d7-43d1-9eef-7afc65b646e1/artifacts/0/docs/build/html/gallery/lines_and_polygons/ocean_bathymetry.html#sphx-glr-gallery-lines-and-polygons-ocean-bathymetry-py)

I should also bring up the earlier commit adding bbox=None to the BasicReader class. This was because the circleci docs builder is not using fiona and passing a bbox as in this example threw an error. Since the two classes are treated interchangeably, I would not expect this to cause an error. Therefore, now the behavior is that passing a bbox will not cause an error but will be silently ignored unless you are using FionaReader. Is this reasonable, or maybe that situation should raise a warning instead noting that the bounding box will not be used?

examples/lines_and_polygons/ocean_bathymetry.py
```diff
@@ -0,0 +1,96 @@
...
+
+if __name__ == "__main__":
+    # Load data (14.8 MB file)
+    depths, shp_dict = load_bathymetry(
+        'https://naturalearth.s3.amazonaws.com/' +
+        '10m_physical/ne_10m_bathymetry_all.zip')
+
+    # Construct colormap:
+    # Depth levels in the colorbar are spaced as in the data
+    norm = matplotlib.colors.Normalize(vmin=-10000, vmax=0)  # in meters
```
greglucas, line 60: Would it be better to quantize this and use BoundaryNorm() here since you have explicit levels?

examples/lines_and_polygons/ocean_bathymetry.py
```diff
@@ -0,0 +1,96 @@
...
+    # Add standard features
+    ax.add_feature(cfeature.LAND, color='grey')
+    ax.coastlines(lw=1, resolution='110m')
+    ax.gridlines(draw_labels=False)
+    ax.set_position([0.03, 0.05, 0.8, 0.9])
+
+    # Add custom colorbar
+    axi = fig.add_axes([0.85, 0.1, 0.025, 0.8])
+    ax.add_feature(cfeature.BORDERS, linestyle=':')
+    matplotlib.colorbar.ColorbarBase(ax=axi,
```
greglucas, line 85: Can you use `fig.colorbar()` here instead since I think that is more typical than going for the base classes in an example.

lgolston: Done.

examples/lines_and_polygons/ocean_bathymetry.py
lgolston: Thanks for this comment @greglucas . I have switched it to use BoundaryNorm. This ended up helping better link the data to the colorbar.

examples/lines_and_polygons/ocean_bathymetry.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+if __name__ == "__main__":
+    # Load data (14.8 MB file)
+    depths, shp_dict = load_bathymetry(
+        'https://naturalearth.s3.amazonaws.com/' +
+        '10m_physical/ne_10m_bathymetry_all.zip')
+
+    # Construct a discrete colormap with colors corresponding to each depth
+    bounds = sorted(depths.astype(float))  # low to high
```
greglucas, line 59:
```suggestion
    boundaries = sorted(depths.astype(float))  # low to high
```
minor nit, but bounds to me sounds like single low/high values, while boundaries brings up a list of quantities.

examples/lines_and_polygons/ocean_bathymetry.py
```diff
@@ -0,0 +1,96 @@
...
+    N = len(bounds)
+    norm = matplotlib.colors.BoundaryNorm(bounds, N)
+    blues_cm = matplotlib.colormaps['Blues_r'].resampled(N)
+    colors_depths = blues_cm(norm(depths.astype(float)))
```
greglucas, line 63: I'm not following why you need all of the `astype()` casting here and below because you use both float and int, but the array should already be one or the other right?

examples/lines_and_polygons/ocean_bathymetry.py
```diff
@@ -0,0 +1,96 @@
...
+    ax.gridlines(draw_labels=False)
+    ax.set_position([0.03, 0.05, 0.8, 0.9])
+
+    # Add custom colorbar
+    axi = fig.add_axes([0.85, 0.1, 0.025, 0.8])
+    ax.add_feature(cfeature.BORDERS, linestyle=':')
+    sm = plt.cm.ScalarMappable(cmap=blues_cm, norm=norm)
+    fig.colorbar(mappable=sm,
+                 cax=axi,
+                 boundaries=depths.astype(int)[::-1],
```
greglucas, line 88:
```suggestion
                 boundaries=boundaries,
```
Haven't you already done this above?

greglucas (commented):
Nice, this reads much simpler to me now. Just a few minor comments.

examples/lines_and_polygons/ocean_bathymetry.py
lgolston: The depths array contains strings, used to index shp_dict. I have now created a separate variable so the casting is only done once where the numerical version is needed.

examples/lines_and_polygons/ocean_bathymetry.py
lgolston: Thanks - bounds is now renamed to boundaries

lgolston:
Thanks for providing those comments. When working through them, I found the color displayed for depth=0 was not matching the colorbar, which is fixed now. Hopefully things are squared away now but please let me know if you see any other issues.

greglucas (approved):

greglucas:
Thanks @lgolston!

mattphysics:
Thanks a lot for all this work, it’s great to see it being finished!

Am 14.07.2023 um 22:03 schrieb Greg Lucas ***@***.***>:

﻿

Thanks @lgolston<https://github.com/lgolston>!

PR #2196, opened 2023-06-23
Pass 'keep' instead of '' in AxesGrid example

lgolston:
## Rationale

The miscellanea/axes_grid_basic now throws a matplotlib deprecation warning due to an undefined label_mode. This updated to 'keep' according to https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.7.0.html#passing-undefined-label-mode-to-grid


greglucas (approved):
Thanks @lgolston!

PR #2197, opened 2023-06-25
BLD: Wheels!

### LINKED ISSUE #2216: Packaging dependency requirement
### Description

`packaging` is only listed as a test dependency, but we are now version gating in the main importable code section. We should either figure out a different way to do the version checks within the code or add it as a primary dependency.

ping @rcomer it looks like this came in these two PRs: #2213, #2215

greglucas:
~~This is one fairly large commit on top of the commits removing GEOS in PR #2083.~~

* Updates the `pyproject.toml` file to reduce the size of setup.py and move towards the standard
* Removes all of the conda environment setup and testing and switches over to pip for all of the installations on GitHub Actions (also removes appveyor and uses GitHub Actions for the windows runners now)
* Adds cibuildwheel to the release for building wheels for the project

~~Right now, this is mostly just to demonstrate that we can build wheels (tested the produced artifacts from my fork and seemed to be good from both the sdist and wheels).~~

Closes #2216


greglucas:
Alright, GEOS is now gone from Cartopy and we can start building wheels 🚀 

This should be ready for review now, and it would be good if someone could look at this and make sure it all makes sense. Maybe pull down this branch and trying building it locally to make sure that the contents of the wheels are what we expect. I am definitely not an expert on build contents.

Once this PR goes in, we will need to remove Appveyor from the repository because it appears like it is triggering even without a `.appveyor` file in this PR.

varchasgopalaswamy:
I was able to build a wheel (`pip wheel --wheel-dir='./wheelhouse'`) and install from that wheel (`pip install wheelhouse/Cartopy-0.20.2-cp311-linux_x86_64.whl`). Python 3.11, WSL ubuntu on Win11. Looks like it loaded up just fine! Ran the example of a line from NYC to Delhi and it looks good to me. 
image

rcomer:
I can’t help with the technical side of this, but it might be good to also update [this page](https://output.circle-artifacts.com/output/job/3696248d-1576-44a3-8559-56699c200d8f/artifacts/0/docs/build/html/installing.html).

greglucas:
@varchasgopalaswamy, thank you for checking!

> pip install wheelhouse/Cartopy-0.20.2-cp311-linux_x86_64.whl

That version doesn't seem quite right to me. My initial guess is you just haven't pulled down the latest tags? Can you try pulling the remote tags `git fetch --tags` and pulling again to make sure we are getting the right version on your system. For reference, this is what I get locally on a mac following your instructions above: `pip install wheelhouse/Cartopy-0.21.1.dev183+g2dbf5b34-cp311-cp311-macosx_10_15_x86_64.whl`

> I can’t help with the technical side of this, but it might be good to also update [this page](https://output.circle-artifacts.com/output/job/3696248d-1576-44a3-8559-56699c200d8f/artifacts/0/docs/build/html/installing.html).

@rcomer, sure I updated some of that page, let me know what you think of the rewording.

INSTALL
```diff
@@ -46,11 +43,8 @@ Required dependencies
 ~~~~~~~~~~~~~~~~~~~~~
 
 In order to install Cartopy, or to access its basic functionality, it will be
-necessary to first install **GEOS**, **Shapely**, and
-**pyshp**. Many of these packages can be installed using pip or
-other package managers such as apt-get (Linux) and brew (macOS). Many of these
-dependencies are built as part of Cartopy's conda distribution, and the recipes
-for these packages can be found at https://github.com/conda-forge/feedstocks.
+necessary to first install the required dependencies lsited below.
```
rcomer, line 46:
```suggestion
necessary to first install the required dependencies listed below.
```

rcomer:
I found a typo, buy otherwise the rewording looks good to me 😀

.github/workflows/release.yml
```diff
@@ -1,63 +1,112 @@
...
+    runs-on: ubuntu-latest
 
     steps:
       - name: Download packages
         uses: actions/download-artifact@v3
-
-      - name: Consolidate packages for upload
-        run: |
-          mkdir dist
-          cp packages-*/* dist/
```
dopplershift, line 60: Are the sdist and all the wheels getting put in a single directory (or otherwise where they can be found for upload) without this?

dopplershift (commented):
🎉 Awesome to see this. Overall this seems pretty sensible, though the cibuildwheel stuff was new dark magic to me.

.github/workflows/release.yml
greglucas: Yes, they are getting put into one large artifact that you'd have to download. See here for a CI run testing on my fork to see what that looks like: https://github.com/greglucas/cartopy/actions/runs/5371702871

For the black magic, I was relying on these links: https://github.com/pypa/cibuildwheel/discussions/1261
https://iscinumpy.dev/post/cibuildwheel-2-10-0/#only-210

dopplershift: Ah, I see what's going on now with the downloading of artifacts. Nice and clean.

dopplershift (approved):
Looks good to me. I'll leave it open for a bit for any comments, but I'm happy to see this go in and get us closer to a 0.22 release.

dopplershift:
@greglucas Note there's a conflict now in `environment.yml`.

.github/workflows/release.yml
```diff
@@ -1,63 +1,112 @@
...
-          PACKAGES="cython fiona matplotlib-base numpy pyproj pykdtree scipy"
-          PACKAGES="$PACKAGES owslib pep8 pillow pyshp pytest"
-          PACKAGES="$PACKAGES pytest-xdist setuptools_scm shapely"
-          conda install $PACKAGES
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - uses: pypa/cibuildwheel@v2.13.1
```
QuLogic, line 62: Slightly outdated now; not sure if you want to use the commit hash as in Matplotlib.

.github/workflows/release.yml
```diff
@@ -1,63 +1,112 @@
...
+        with:
+          only: ${{ matrix.only }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./wheelhouse/*.whl
+
+  build_sdist:
+    name: Build source distribution
```
QuLogic, line 71: We may, as in Matplotlib's current build, want to build sdist first and instruct `cibuildwheel` to build out of that.

.github/workflows/release.yml
```diff
@@ -1,63 +1,112 @@
...
+        with:
+          python-version: '3.11'
 
-      - name: Save built packages as artifact
-        uses: actions/upload-artifact@v3
+      - name: Build sdist
+        run: |
+          python -m pip install build
+          # install compiles the cython *.pyx to *.cpp for us
+          python -m pip install .
```
QuLogic, line 88: This doesn't seem right, though I don't know how to effectively Google `build` things.

lib/cartopy/trace.pyx
```diff
@@ -305,9 +305,7 @@ cdef bool straightAndDomain(double t_start, const Point &p_start,
         p_mid = interpolator.interpolate(t_mid)
 
         # Determine the closest point on the segment to the midpoint, in
-        # normalized coordinates. We could use GEOSProjectNormalized_r
-        # here, but since it's a single line segment, it's much easier to
-        # just do the math ourselves:
+        # normalized coordinates. We can just do the math ourselves:
```
QuLogic, line 308: The "We can just do the math ourselves" sentence seems a bit dangling without the previous context.

pyproject.toml
```diff
@@ -8,6 +8,73 @@ requires = [
...
+    "pyproj>=3.1.0",
+]
+dynamic = ["version"]
+
+[project.optional-dependencies]
+doc = ["beautifulsoup4", "pydata-sphinx-theme", "sphinx", "sphinx-gallery"]
+extras = ["pykdtree", "fiona"]
+ows = ["OWSLib>=0.20.0", "pillow>=6.1.0"]
+plotting = ["pillow>=6.1.0", "scipy>=1.3.1"]
+test = ["packaging>=20", "pytest>=5.1.2", "pytest-mpl>=0.11", "pytest-xdist", "pytest-cov", "coveralls"]
```
QuLogic, line 56: `packaging` is in the runtime dependencies already.

pyproject.toml
```diff
@@ -8,6 +8,73 @@ requires = [
...
+    "shapely>=1.7",
+    "packaging>=20",
+    "pyshp>=2.1",
+    "pyproj>=3.1.0",
+]
+dynamic = ["version"]
+
+[project.optional-dependencies]
+doc = ["beautifulsoup4", "pydata-sphinx-theme", "sphinx", "sphinx-gallery"]
+extras = ["pykdtree", "fiona"]
```
QuLogic, line 53: I'm not sure `extras` is a useful name for the feature.

.github/workflows/ci-testing.yml
```diff
@@ -10,81 +10,71 @@ jobs:
...
+        # Default is to install just the minimum testing requirements,
+        # but we want to get as much coverage as possible.
+        if: steps.minimum-packages.conclusion == 'skipped'
+        run: |
+          pip install .[extras,ows,plotting]
+
       - name: Testing
         id: test
+        # we need to force bash to use line continuations
+        shell: bash
```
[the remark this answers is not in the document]
greglucas: I am pretty sure this was for the Windows CI job, which Matplotlib doesn't have.

.github/workflows/release.yml
greglucas: I'm a bit out of my element here. I tried to do that with a new commit on a separate branch.
https://github.com/greglucas/cartopy/tree/setup-updates-tmp

Failed job: https://github.com/greglucas/cartopy/actions/runs/5601696424/jobs/10245962338

I think it is failing because the sdist isn't being built quite right. It is not cythonizing to convert from `.pyx` to `.cpp`, where we need to include the `.cpp` in the sdist to distribute into the wheel. I'm not sure where we are supposed to build the extension then, do we just call `python setup.py build_ext` as a separate step to make sure the `.cpp` is created?

.github/workflows/release.yml
greglucas: I think you're right here. See my above reply. I think what this was doing for me was to compile the `.pyx` to `.cpp`, so we need another way of doing that then if we don't want to pip install here before the build. Any thoughts/ideas appreciated on this.

pyproject.toml
greglucas: Yeah, I was just copying over from the previous `requirements` directory files. I think we could re-evaluate which of these categories we want. I renamed it to `speedups` in a separate commit/branch.

.github/workflows/release.yml
greglucas: If it is easier for you to take over some of this, feel free to branch off yourself or push directly to my branch. I'm definitely happy to have some help on this.

greglucas: OK, I think we just needed to call `cythonize()` if we are going down the non-sdist route, so I've reworked the logic a bit in `setup.py` to try and simplify things a bit more for whether we are in the Cython route or the pre-made cpp route.
Here is a CI run from a separate branch on my fork that demonstrates it working now when building the wheels from the sdist: https://github.com/greglucas/cartopy/actions/runs/5612238838/jobs/10269964395

varchasgopalaswamy: Maybe this is out of scope for this PR, but any thoughts about using numba to replace the .pyx file? 
Upside: No need to compile the .pyx file during build and deal with correctly packaging it. Plus numba has support for automatic parallelization. 
Downside: Another dependency.

greglucas: Definitely out of scope for this PR. If you can show that it is faster than Cython or makes things easier in some way, then a PR would definitely be considered!

dopplershift: As much as I think Numba is cool, it has its own challenges due to its reliance on bridging LLVM and Python. For instance, Numba didn't support Python 3.11 until May 2023, which IMO is unacceptable for something we'd consider as a dependency. I understand the dev team is aware how problematic this is and intends to do better for future releases, but I'd want to see a good track record going forward before we begin to consider that.

greglucas:
@QuLogic, thanks for the review comments! I think I addressed them all, but let me know if there was anything else I missed. Otherwise, I think it could still use a once-over to see if you spot anything I did odd when making those updates too.

.github/workflows/release.yml
QuLogic: Now that we are preparing to build wheels, I wonder if all the work to make Cython optional is all that useful? It was a convenience to not require it, I suppose, but with stuff like build-requires and isolated build venvs *and* the fact that we're going to make wheels, reducing the need to build from source, maybe that isn't so great as it once was.

.circleci/config.yml
```diff
@@ -12,56 +12,21 @@ version: 2.1
...
-        --file docs/doc-requirements.txt
-    conda list -n test-environment
-
 cp-run: &cp-install
   name: Install Cartopy
   command: |
-    source activate test-environment
-    pip install -ve .
+    python -m pip install --upgrade --user pip
+    python -m pip install --user -ve .[doc,extras,ows,plotting]
```
QuLogic, line 25: s/extras/speedups/ now

.github/workflows/ci-testing.yml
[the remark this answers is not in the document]
QuLogic: Extend comment to mention Windows, then?

.github/workflows/release.yml
```diff
@@ -1,63 +1,124 @@
...
-          name: packages-${{ runner.os }}-${{ matrix.python-version }}
-          path: dist/
+          name: sdist
+          path: dist/*.tar.gz
           if-no-files-found: error
-          retention-days: 5
+
+  generate-wheels-matrix:
+    name: Generate wheels matrix
+    needs: build_sdist
```
QuLogic, line 53: Does this really need the sdist?

.github/workflows/release.yml
```diff
@@ -1,63 +1,124 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      include: ${{ steps.set-matrix.outputs.include }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install cibuildwheel
+        run: pipx install cibuildwheel==2.14.1
```
QuLogic, line 60: Change to commit hash as well.

.github/workflows/release.yml
```diff
@@ -1,63 +1,124 @@
...
+              | jq -nRc '{"only": inputs, "os": "windows-2019"}'
+            } | jq -sc
+          )
+          echo "include=$MATRIX" >> $GITHUB_OUTPUT
+        env:
+          CIBW_BUILD: "cp39-* cp310-* cp311-*"
+          # Skip 32 bit windows builds and musllinux due to lack of numpy wheels
+          CIBW_SKIP: "*-win32 *-musllinux*"
+          CIBW_ARCHS_MACOS: x86_64 arm64
+  build_wheels:
```
QuLogic, line 79:
```suggestion
          CIBW_ARCHS_MACOS: x86_64 arm64

  build_wheels:
```

.github/workflows/release.yml
greglucas: Possibly we could take it out... The Cython docs do mention to use an environment variable `USE_CYTHON` to control this, so I think we are doing similar to their recommendations. Doesn't hurt to keep it I guess?
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#distributing-cython-modules

.circleci/config.yml
greglucas: Thanks, I also got one or two other locations that missed this as well.

.github/workflows/release.yml
greglucas: I figured that would just stop it from continuing on if even the sdist failed. But, I removed it for now, we shouldn't be going down this path too often anyways.

.github/workflows/release.yml
```diff
@@ -1,63 +1,126 @@
...
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./wheelhouse/*.whl
 
   publish:
     name: Publish to PyPI
-    needs: build
+    # Only publish on releases
+    if: github.event_name == 'release'
```
QuLogic, line 109: Isn't this redundant with the entire workflow's `on` trigger? Should that be removed?

greglucas: You beat me to this :) I was just going to add a comment here indicating I added this. I agree it is the same as the trigger, but this is an extra safeguard that I think is worth it in this case. For example, when I was testing building wheels on my fork, I removed the "on: release", but this would fail because I don't have credentials. However, then I force-pushed to this branch, and all of a sudden triggered this workflow on the PR. I'm not entirely clear it would have gone through the action but it seems prudent to keep to me just to make sure I don't fat finger something again on accident. In the future, we may want to change the trigger to let us trigger it automatically on certain pushes or on request.

QuLogic (approved):
We can followup the other questions in other PRs.

.github/workflows/release.yml
dopplershift: PRs shouldn't have any of the credentials/write permissions necessary.

Also, there's no way this workflow as written should be able to trigger on a PR.

That said, there's no harm in being careful.

PR #2201, opened 2023-06-27
precompute finite value mask for kdtree query

### LINKED ISSUE #2199: BUG: incompatibility with scipy 1.11.0
### Description
Scipy 1.11.0 (released yesterday) appear to break [use cases exercised in yt's CI](https://github.com/yt-project/yt/issues/4540)

More specifically the breaking change seems to be https://github.com/scipy/scipy/pull/18502

#### Code to reproduce

```python
import numpy as np
import cartopy.crs as ccrs
from matplotlib.figure import Figure

fig = Figure()
ax = fig.add_axes((0, 0, 1, 1), projection=ccrs.Mollweide())
ax.imshow([[0, 1], [0, 1]], transform=ccrs.PlateCarree())
```

#### Traceback

```python-traceback
  File "/private/tmp/t.py", line 7, in <module>
    ax.imshow([[0, 1], [0, 1]], transform=ccrs.PlateCarree())
  File "/private/tmp/.venv/lib/python3.11/site-packages/cartopy/mpl/geoaxes.py", line 318, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/.venv/lib/python3.11/site-packages/cartopy/mpl/geoaxes.py", line 1331, in imshow
    img, extent = warp_array(img,
                  ^^^^^^^^^^^^^^^
  File "/private/tmp/.venv/lib/python3.11/site-packages/cartopy/img_transform.py", line 192, in warp_array
    array = regrid(array, source_native_xy[0], source_native_xy[1],
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/.venv/lib/python3.11/site-packages/cartopy/img_transform.py", line 278, in regrid
    _, indices = kdtree.query(target_xyz, k=1)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "_ckdtree.pyx", line 795, in scipy.spatial._ckdtree.cKDTree.query
ValueError: 'x' must be finite, check for nan or inf values
```

<details>

### Operating system

Discovered in CI on Ubuntu, and reproduced on macOS

### Cartopy version
0.21.1

### pip list

```
Package         Version
--------------- --------
Cartopy         0.21.1
certifi         2023.5.7
contourpy       1.1.0
cycler          0.11.0
fonttools       4.40.0
kiwisolver      1.4.4
matplotlib      3.7.1
numpy           1.25.0
packaging       23.1
Pillow          9.5.0
pip             23.1.2
pyparsing       3.1.0
pyproj          3.6.0
pyshp           2.3.1
python-dateutil 2.8.2
scipy           1.11.0
setuptools      65.5.0
shapely         2.0.1
six             1.16.0
```

</details>

chrishavlin:
## Rationale

scipy>=1.11 now raises an error when querying a kdtree object with coordinate arrays that contain `nan` or `inf` values. This PR adds a mask for finite values before that query occurs. 

## Implications

Closes #2199

### DIFF
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -13,11 +13,11 @@ import numpy as np
 
 
 try:
-    from pykdtree.kdtree import KDTree as _kdtree_handle
+    from pykdtree.kdtree import KDTree as _kdtreeClass
     _is_pykdtree = True
 except ImportError:
     try:
-        from scipy.spatial import cKDTree as _kdtree_handle
+        from scipy.spatial import cKDTree as _kdtreeClass
     except ImportError as e:
         raise ImportError("Using image transforms requires either "
                           "pykdtree or scipy.") from e
@@ -274,7 +274,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
     finite_xyz = np.all(np.isfinite(target_xyz), axis=-1)
 
     if _is_pykdtree:
-        kdtree = _kdtree_handle(xyz)
+        kdtree = _kdtreeClass(xyz)
         # Use sqr_dists=True because we don't care about distances,
         # and it saves a sqrt.
         _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :],
@@ -283,7 +283,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
     else:
         # Versions of scipy >= v0.16 added the balanced_tree argument,
         # which caused the KDTree to hang with this input.
-        kdtree = _kdtree_handle(xyz, balanced_tree=False)
+        kdtree = _kdtreeClass(xyz, balanced_tree=False)
         _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :], k=1)
 
     mask = ~finite_xyz | (indices >= len(xyz))
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -286,7 +286,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
         kdtree = scipy.spatial.cKDTree(xyz, balanced_tree=False)
         _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :], k=1)
 
-    mask = indices >= len(xyz)
+    mask = ~finite_xyz | (indices >= len(xyz))
     indices[mask] = 0
 
     desired_ny, desired_nx = target_x_points.shape
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -13,11 +13,11 @@ import numpy as np
 
 
 try:
-    import pykdtree.kdtree
+    from pykdtree.kdtree import KDTree as _kdtree_handle
     _is_pykdtree = True
 except ImportError:
     try:
-        import scipy.spatial
+        from scipy.spatial import cKDTree as _kdtree_handle
     except ImportError as e:
         raise ImportError("Using image transforms requires either "
                           "pykdtree or scipy.") from e
@@ -274,7 +274,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
     finite_xyz = np.all(np.isfinite(target_xyz), axis=-1)
 
     if _is_pykdtree:
-        kdtree = pykdtree.kdtree.KDTree(xyz)
+        kdtree = _kdtree_handle(xyz)
         # Use sqr_dists=True because we don't care about distances,
         # and it saves a sqrt.
         _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :],
@@ -283,7 +283,7 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
     else:
         # Versions of scipy >= v0.16 added the balanced_tree argument,
         # which caused the KDTree to hang with this input.
-        kdtree = scipy.spatial.cKDTree(xyz, balanced_tree=False)
+        kdtree = _kdtree_handle(xyz, balanced_tree=False)
         _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :], k=1)
 
     mask = ~finite_xyz | (indices >= len(xyz))
--- a/lib/cartopy/img_transform.py
+++ b/lib/cartopy/img_transform.py
@@ -268,16 +268,24 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
                                               target_x_points.flatten(),
                                               target_y_points.flatten())
 
+    # Find mask of valid points before querying kdtree: scipy >= 1.11 errors
+    # when querying nan points, might as well use for pykdtree too.
+    indices = np.zeros(target_xyz.shape[0], dtype=int)
+    finite_xyz = np.all(np.isfinite(target_xyz), axis=-1)
+
     if _is_pykdtree:
         kdtree = pykdtree.kdtree.KDTree(xyz)
         # Use sqr_dists=True because we don't care about distances,
         # and it saves a sqrt.
-        _, indices = kdtree.query(target_xyz, k=1, sqr_dists=True)
+        _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :],
+                                              k=1,
+                                              sqr_dists=True)
     else:
         # Versions of scipy >= v0.16 added the balanced_tree argument,
         # which caused the KDTree to hang with this input.
         kdtree = scipy.spatial.cKDTree(xyz, balanced_tree=False)
-        _, indices = kdtree.query(target_xyz, k=1)
+        _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :], k=1)
+
     mask = indices >= len(xyz)
     indices[mask] = 0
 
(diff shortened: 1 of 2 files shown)


lib/cartopy/img_transform.py
```diff
@@ -268,16 +268,24 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
...
+                                              k=1,
+                                              sqr_dists=True)
     else:
         # Versions of scipy >= v0.16 added the balanced_tree argument,
         # which caused the KDTree to hang with this input.
         kdtree = scipy.spatial.cKDTree(xyz, balanced_tree=False)
-        _, indices = kdtree.query(target_xyz, k=1)
+        _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :], k=1)
+
     mask = indices >= len(xyz)
```
chrishavlin, line 289: I'm not sure whether the following `mask` behavior should be refactored...

lib/cartopy/tests/test_img_transform.py
```diff
@@ -108,3 +108,20 @@ def test_gridding_data_outside_projection():
     assert_array_equal([-180, 180, -90, 90], extent)
     assert_array_equal(expected, image)
     assert_array_equal(expected_mask, image.mask)
+
+
+@pytest.mark.parametrize("target_prj", (ccrs.Mollweide(), ccrs.LambertAzimuthalEqualArea()))
+def test_invalid_regridding_with_invalid_extent(target_prj):
```
chrishavlin, line 114: these tests currently fail when running on the `main` branch with the error in #2199 and pass on the current branch. wasn't entirely sure what to check in terms of values though, so at present this is just checking that the transformation does not error.

lib/cartopy/img_transform.py
```diff
@@ -268,16 +268,24 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
                                               target_x_points.flatten(),
                                               target_y_points.flatten())
 
+    # Find mask of valid points before querying kdtree: scipy >= 1.11 errors
+    # when querying nan points, might as well use for pykdtree too.
+    indices = np.zeros(target_xyz.shape[0], dtype=int)
```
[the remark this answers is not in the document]
greglucas: I think this is fine to use 0 here as you noted because I think it is just settting it to some valid index. But, I think we still need to add these non-finite locations you found here into the `mask` below as well? Something like `mask = ~finite_xyz | (indices >= len(xyz))`

lib/cartopy/tests/test_img_transform.py
chrishavlin: Note: on running tests locally, the following tests fail with scipy 1.11.1 on main but pass on this branch:

```
============================================================================================================== short test summary info ==============================================================================================================
FAILED lib/cartopy/tests/mpl/test_images.py::test_imshow - ValueError: 'x' must be finite, check for nan or inf values
FAILED lib/cartopy/tests/mpl/test_images.py::test_imshow_rgba_alpha - ValueError: 'x' must be finite, check for nan or inf values
FAILED lib/cartopy/tests/mpl/test_images.py::test_stock_img - ValueError: 'x' must be finite, check for nan or inf values
FAILED lib/cartopy/tests/mpl/test_images.py::test_pil_Image - ValueError: 'x' must be finite, check for nan or inf values
FAILED lib/cartopy/tests/mpl/test_images.py::test_background_img - ValueError: 'x' must be finite, check for nan or inf values
```
I think these tests are likely passing the CI on other PR's because the test suite spinup installs both scipy and pykdtree, and pykdtree takes precedence when deciding on a kdtree method.

chrishavlin:
@greglucas I think the actual code changes are ready for (final?) review. 

The main outstanding question I have is how to handle tests. To actually test these changes with scipy, I'd need to refactor how the scipy/pykdtree choice happens. I'm happy to do that -- my initial inclination is to add a function to set which package to use (defaulting to pykdtree) so that there is a way to toggle between the two in the test environment. Is it OK to add that to this PR or would it be better as a followup PR?

rcomer:
I may have misunderstood but, for testing, could you [monkeypatch](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setattr) `img_transform._is_pykdtree` to `False`?

chrishavlin:
> I may have misunderstood but, for testing, could you [monkeypatch](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setattr) `img_transform._is_pykdtree` to `False`?

not exactly -- because the scipy kdtree import only happens on initial module import. So if you monkeypatch `_is_pykdtree` to False, you get an error cause the scipy kdtree hasn't been imported yet.

chrishavlin:
One option would be to store the kdtree function handle in a variable to allow monkey patching during tests.

In pseudo-code, I was thinking along the lines of:

```
def _get_kdtree_function(use_pykdtree = True):
    if use_pykdtree and pykdtree is available:
         return True, pykdtree.kdtree.KDTree
    elif use_pykdtree is False and scipy is available:
        return False, scipy.spatial.cKDTree
    else:
          error 

_is_pydktree, _kdtree_func = _get_kdtree_function()
```

and then using that `_kdtree_func` handle down in `regrid`. That would allow you to monkeypatch both the boolean and the function handle in the test environment.

chrishavlin:
OK, found a simple-ish solution. Realized that if the scipy and pykdtree kdtree object handles were imported into a module attribute similar to `_is_pykdtree`, the kdtree object could be monkeypatched in a simple way during testing (in addition to the the `_is_pykdtree` flag. I updated the new test to switch between using scipy and pykdtree. 

So I don't see anything else to add here at the moment, let me know if anything more is needed (or if there's a better approach to testing both scipy and pykdtree here).

chrishavlin:
oops, forgot to setup pre-commit locally... will fix that in a moment.

lib/cartopy/tests/test_img_transform.py
```diff
@@ -108,3 +109,28 @@ def test_gridding_data_outside_projection():
...
+    target_x, target_y, extent = img_trans.mesh_projection(target_prj, 8, 4)
+
+    if use_scipy:
+        monkeypatch.setattr(img_trans, "_is_pykdtree", False)
+        monkeypatch.setattr(img_trans, "_kdtree_handle", scipy.spatial.cKDTree)
+        _ = img_trans.regrid(data, lons, lats, data_trans, target_prj,
+                             target_x, target_y)
+    else:
+        _ = img_trans.regrid(data, lons, lats, data_trans, target_prj,
+                             target_x, target_y)
```
greglucas, line 136: Can you dedent this and only keep one of them and only do the monkeypatching in the if-block, but the regridding outside?

lib/cartopy/img_transform.py
```diff
@@ -13,11 +13,11 @@
 
 
 try:
-    import pykdtree.kdtree
+    from pykdtree.kdtree import KDTree as _kdtree_handle
```
greglucas, line 16:
```suggestion
    from pykdtree.kdtree import KDTree as _kdtreeClass
```
Thoughts on making this explicit that it is a Class you're using below? I wasn't entirely clear what handle meant at first glance.

lib/cartopy/img_transform.py
```diff
@@ -268,17 +268,25 @@ def regrid(array, source_x_coords, source_y_coords, source_proj, target_proj,
...
+
     if _is_pykdtree:
-        kdtree = pykdtree.kdtree.KDTree(xyz)
+        kdtree = _kdtree_handle(xyz)
         # Use sqr_dists=True because we don't care about distances,
         # and it saves a sqrt.
-        _, indices = kdtree.query(target_xyz, k=1, sqr_dists=True)
+        _, indices[finite_xyz] = kdtree.query(target_xyz[finite_xyz, :],
+                                              k=1,
+                                              sqr_dists=True)
```
greglucas, line 282: Since the object creation and query are so similar, do you think it'd be worth just setting the different keyword arguments in a dict and then unpacking for the same call function below. Something like this:

```python
if _is_pykdtree:
    class_kws = {}
    query_kws = {'k': 1, 'sqr_dists': True}
else:
    class_kws = {'balanced_tree': False}
    query_kws = {'k': 1}

kdtree = _kdtree_handle(xyz, **class_kws)
kdtree.query(target_xyz, **query_kws)
```

I think it is fine as-is too, just thought I'd throw that other option out there as well.

greglucas (approved):
I think this is good to go from a code perspective, I've just made a few minor nits that you can take or leave.

lib/cartopy/tests/test_img_transform.py
chrishavlin: Yes! Will do

lib/cartopy/img_transform.py
chrishavlin: I don't have a strong feeling on it. Both versions seem clear enough to me, happy to defer to you on this.

chrishavlin:
Thanks for the review! Will address the changes in the next few days.

lib/cartopy/img_transform.py
greglucas: I'd suggest just leaving it as you have it currently then and ignore this comment.

lib/cartopy/tests/test_img_transform.py
chrishavlin: done!

lib/cartopy/img_transform.py
chrishavlin: also done!

chrishavlin:
Totally OK with keeping this open til #2209 is merged then merging here to make extra sure that I'm not actually breaking any tests here.

greglucas:
Thank you @chrishavlin! This was a very nice fix.

Issue #2198, opened 2023-06-26
Stamen watercolor tile requests send back HTTP Error 404


wckoeppen:
### Description
Attempting to use the `cartopy.io.img_tiles.Stamen` tiles with the `watercolor` style sends back an HTTP Error 404: Not Found error.
 
It looks like the Stamen tile urls may have changed slightly? [This code](https://github.com/SciTools/cartopy/blob/main/lib/cartopy/io/img_tiles.py#L360) asks for
```
f'http://tile.stamen.com/{self.style}/{z}/{x}/{y}.png'
```
These requests get redirected to `https://tiles.stadiamaps.com/styles/`, and it looks like the watercolor tiles are only made available there as jpg, not png. (Other tilesets are available in both formats.)

Doesn't work: https://tiles.stadiamaps.com/styles/stamen_watercolor/12/1104/1625.png
Works:             https://tiles.stadiamaps.com/styles/stamen_watercolor/12/1104/1625.jpg

#### Code to reproduce

```
import matplotlib.pyplot as plt
import cartopy.io.img_tiles as cimgt

request = cimgt.Stamen(style='watercolor')

fig = plt.figure()
ax = fig.add_axes([0,0,1,1], projection=request.crs)
ax.set_extent([-83.35, -83.30, 34.85, 34.90])
img = ax.add_image(request, 14)
plt.show()
```
### Operating system
Mac OS Ventura 13.4.1

### Cartopy version
0.21.1

### conda list

```
  - numpy
  - gpxpy
  - python=3.10
  - shapely
  - geopandas
  - matplotlib
  - tqdm
  - cartopy
  - pandas
  - jupyterlab
```

lgolston:
Thanks @wckoeppen for reporting! I don't know if the png version is still available or not, but jpg seems to be the primary version of watercolor so we should look at using that. More broadly, could be handled something like in openlayers: [link](https://github.com/openlayers/openlayers/blob/fe68ca7370bdefe5419cec72fcb910efee250577/src/ol/source/Stamen.js#L22)

rcomer:
I think this one is fixed by #2202?

dopplershift:
I think so. This should be fixed either by running `main` or installing the next release when it becomes available. If you find that's not the case, feel free to re-open.

wckoeppen:
Thanks everyone. Much appreciated!

PR #2202, opened 2023-06-29
Preset Stamen tile file extension, opacity

lgolston:
## Rationale

In issue #2198 it was pointed out the Stamen watercolor tiles can not be downloaded currently as png. Right now, all the layers are downloaded as png and also default to RGB mode. This PR creates a dictionary to set the file extension automatically (specifically using jpg for watercolor) and desired_tile_form to RGBA for known layers which should have a transparent background (like terrain-lines).

## Implications

Stamen watercolor feature should be working again, and behavior of behavior of non-opaque layers improved. By explicitly listing the known Stamen layers in a dict, this also may help the user be aware of the available options.

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -374,10 +374,10 @@ class Stamen(GoogleWTS):
         # use optional desired_tile_form input if available
         # otherwise, use preset value based on the layer name
         if desired_tile_form is None:
-            if not layer_info['opaque']:
-                desired_tile_form = 'RGBA'
-            else:
+            if layer_info['opaque']:
                 desired_tile_form = 'RGB'
+            else:
+                desired_tile_form = 'RGBA'
 
         super().__init__(desired_tile_form=desired_tile_form,
                          cache=cache)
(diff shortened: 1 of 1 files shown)


lgolston:
The initial commit ignored the optional desired_tile_form input parameter, which is now fixed. I believe the code is ready for review.

lib/cartopy/io/img_tiles.py
```diff
@@ -350,14 +350,44 @@ class Stamen(GoogleWTS):
...
+        layer_info = layer_config.get(
+            style, {'extension': '.png', 'opaque': True})
+
+        # use optional desired_tile_form input if available
+        # otherwise, use preset value based on the layer name
+        if desired_tile_form is None:
+            if not layer_info['opaque']:
+                desired_tile_form = 'RGBA'
+            else:
+                desired_tile_form = 'RGB'
```
dopplershift, line 380:
```suggestion
            if layer_info['opaque']:
                desired_tile_form = 'RGB'
            else:
                desired_tile_form = 'RGBA'
```

dopplershift (commented):
Minor nit, but otherwise this looks great!

dopplershift (approved):
Test failures look unrelated so going to merge.

Issue #1972, opened 2021-12-17
Broken links to generated documentation of cartopy.crs.CRS methods


wjbenfold:
### Description

Some links are broken in the cartopy documentation (and consequently in documentation linked into it, e.g. Iris)

Example: https://scitools.org.uk/cartopy/docs/latest/whatsnew/v0.7.html

As of 17/12/2021, the link to cartopy.mpl.geoaxes.GeoAxes.gridlines() works fine, but the link to cartopy.crs.CRS.transform_points() is broken. Links to cartopy.crs.CRS itself are fine.

![image](https://user-images.githubusercontent.com/69585101/146535248-08385a6f-33a4-4b91-b842-230565b90369.png)

rcomer:
This appears to be fixed (in my local build at least).  Maybe by #2183?

greglucas:
Closed by #2220

PR #2220, opened 2023-07-23
Broken links in what's new pages

lgolston:
## Rationale

This fixes several broken links in the what's new pages in the documentation, and uses double backticks around deprecated methods if they weren't already being used.
As a note, in the what's new for v0.5 I found the call to :mod:`cartopy.feature` was not working because of currentmodule being used instead of module in feature.rst. For consistency, this change was then repeated in the .rst file for other modules.

## Implications

I think issue #1972 can be closed.

### DIFF
--- a/docs/source/reference/config.rst
+++ b/docs/source/reference/config.rst
@@ -3,7 +3,7 @@
 Cartopy configuration
 ---------------------
 
-.. currentmodule:: cartopy
+.. module:: cartopy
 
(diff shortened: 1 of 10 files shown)


greglucas (approved):
This looks good. Thanks!

There are still a few warnings from the Sphinx build if you look at the CircleCI logs if you're interested in fixing up some other documentation references too.

PR #2221, opened 2023-07-25
Reduce sphinx warnings

lgolston:
## Rationale

As @greglucas  pointed out in #2220, the CI server shows some warnings while building the docs. Running on a newer version of sphinx (v7.0.1), there is also a warning in GeoSpine.set_position from a link in the inherited docstring. These are addressed here. Some more broken cross-references were also found/cleaned up by looking at the output of sphinx in nitpicky mode.

## Notes
1. This adds the :module: tag to all modules (including sub-modules), so they can be cross-referenced. Previously I had only changed this for the main the top-level modules, which is why there was still a warning about not finding the SRTM modules.

2. A class.rst template is created, which is used by default. Without this, the methods within a class were not generating HTML anchors that can be linked to, other than in the matplotlib and crs modules where a template was already explicitly used.

3. Sphinx version <6 needed to be pinned in docs/doc-requirements.txt (#2131), but I think this not needed now due to pydata/pydata-sphinx-theme#1132

## Implications

Improved cross-referencing throughout the docs.

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,4 +1,4 @@
-sphinx<6
+sphinx
 sphinx-gallery
 beautifulsoup4
 pydata-sphinx-theme
--- a/docs/source/_templates/autosummary/class.rst
+++ b/docs/source/_templates/autosummary/class.rst
@@ -1,0 +1,9 @@
+{{ fullname }}
+{{ underline }}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+    :members:
+    :inherited-members:
+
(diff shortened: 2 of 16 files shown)


lgolston:
Okay, now there are several 'Inline emphasis start-string without end-string' warnings from parsing img_nest.py, such as from `'*.tif'`. I think it can be ignored?

QuLogic:
Those should be wrapped in double backticks for code styling.

lgolston:
Done. Thanks @QuLogic!

greglucas (approved):
Looks great to me and the doc logs are building clean now! Thanks, @lgolston.

PR #2222, opened 2023-07-28
DOC: Remove mentions of Travis CI

greglucas:
Noticed that the link was broken in the README. We don't use Travis anymore, so just remove it. Also removed a few paths that are no longer there (examples, sphinxext) in `.coveragerc`

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,17 +5,10 @@
 [run]
 branch = True
 omit =
-    lib/cartopy/examples/*
-    lib/cartopy/sphinxext/*
     lib/cartopy/tests/*
     lib/cartopy/_version.py
     *.pxd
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -15,7 +13,6 @@ plugins = Cython.Coverage
 [paths]
 source =
     lib/cartopy
-    /home/travis/miniconda/envs/test-environment/lib/python?.?/site-packages/cartopy
```
QuLogic, line 18: This is here to alias (non-editable) installs for coverage purposes. If we only do editable installs on CI and it all works, we can probably get rid of the whole section?

greglucas: :+1: Done.

dopplershift (approved):

PR #2229, opened 2023-08-05
REL: Include sdist in PyPI release process

### LINKED ISSUE #2228: Source code no longer included on PyPI
On conda-forge, we're trying to build 0.22.0:
https://github.com/conda-forge/cartopy-feedstock/pull/156

However, we're running into issues because a `.tar.gz` file wasn't included in the PyPI release.  First, our bot was not able to create a PR for the release (see https://github.com/conda-forge/cartopy-feedstock/issues/155).  Then, when I tried to use the source from GitHub instead, `setuptools-scm` complains:
```
      LookupError: setuptools-scm was unable to detect version for /home/conda/feedstock_root/build_artifacts/cartopy_1691225778495/work.
      
      Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
      
      For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
      [end of output]
```
While there are workarounds that I've used in the past (I'm a little rusty so I'll have to dig them up), it seems like the cleanest solution would be to include the source on PyPI agian.

greglucas:
It looks like we didn't upload the sdist to PyPI. We need to explicitly download that artifact since it is produced and named separately from the wheels.

closes #2228

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,7 +92,7 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: sdist
-          path: dist/
+          path: dist
 
       - uses: pypa/cibuildwheel@f21bb8376a051ffb6cb5604b28ccaef7b90e8ab7 # v2.14.1
         with:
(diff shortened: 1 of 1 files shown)


QuLogic (approved):
Makes sense to me.

PR #2230, opened 2023-08-06
Pass options of stock_img to imshow

### LINKED ISSUE #2179: Plea for a stock_img() alpha
### Description
I am delighted that `.add_feature()` now allows for transparency via an `alpha` parameter; this goes a long way to decluttering busy maps. Unfortunately, the same feature is not available for `stock_img()`, and I was wondering why. I am admittedly ignorant of the way the background is implemented in Cartopy, but given that a lot of matplotlib objects have an alpha parameter, I was hoping this would be a matter of exposing it at the API level. Is this thinking too simplistic?

### Proposal
`ax.stock_img(alpha=0.5)` reduces background opacity by 50%, for instance.

smartlixx:
## Rationale

This will solve the problem raised by #2179 

## Implications

`stock_img` can accept more options, which will be dealt with by `self.imshow`.

Closes #2179

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -999,12 +999,13 @@ class GeoAxes(matplotlib.axes.Axes):
 
         return super().set_yticks(yticks, minor=minor)
 
-    def stock_img(self, name='ne_shaded'):
+    def stock_img(self, name='ne_shaded', **kwargs):
         """
         Add a standard image to the map.
 
-        Currently, the only (and default) option is a downsampled version of
-        the Natural Earth shaded relief raster.
+        Currently, the only (and default) option for image is a downsampled
+        version of the Natural Earth shaded relief raster. Other options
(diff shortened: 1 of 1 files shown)


dopplershift (approved):

lib/cartopy/mpl/geoaxes.py
```diff
@@ -999,12 +999,13 @@ def set_yticks(self, ticks, minor=False, crs=None):
...
-    def stock_img(self, name='ne_shaded'):
+    def stock_img(self, name='ne_shaded', **kwargs):
         """
         Add a standard image to the map.
 
-        Currently, the only (and default) option is a downsampled version of
-        the Natural Earth shaded relief raster.
+        Currently, the only (and default) option for image is a downsampled
+        version of the Natural Earth shaded relief raster. Other options
+        (e.g., alpha) will be passed to func:`self.imshow`.
```
greglucas, line 1008:
```suggestion
        (e.g., alpha) will be passed to :func:`self.imshow`.
```
I think you need an extra colon in front of function to get that rendered properly by sphinx.

Issue #2041, opened 2022-04-29
Encoding in `io.shapereader`


smartlixx:
### Description

`io.shapereader` currently does not support `encoding` specification. It relies on `pyshp` under the hood, but `pyshp` supported `encoding` ever since v2.0 (back in the year of 2018). It would be necessary for `io.shapereader` to support `encoding` as well.

Its implementation should be straightforward. If OK, I can make a PR.

Issue #1282, opened 2019-03-13
'utf-8' codec can't decode


Mo-Dabao:
### Description

Same codes worked well last year, but not now.

#### Code to reproduce

```
import cartopy.io.shapereader as shpreader


def __extract():
    shpname = r".\map\China_province"
    provinces_records = list(shpreader.Reader(shpname).records())
    provinces_geometrys = [x.geometry for x in provinces_records]
    provinces_attributes = [x.attributes for x in provinces_records]
    return provinces_geometrys, provinces_attributes

provinces_shapes, provinces_records = __extract()
```

#### Traceback 

```
Traceback (most recent call last):

  File "<ipython-input-51-fc14594e3fb3>", line 1, in <module>
    from China_map import provinces_geometrys

  File "E:\Python\pyNMIC\FY4A\China_map.py", line 24, in <module>
    provinces_geometrys, provinces_attributes = __extract()

  File "E:\Python\pyNMIC\FY4A\China_map.py", line 18, in __extract
    provinces_records = list(shpreader.Reader(shpname).records())

  File "D:\Anaconda3\lib\site-packages\cartopy\io\shapereader.py", line 249, in records
    shape_record = self._reader.shapeRecord(i)

  File "D:\Anaconda3\lib\site-packages\shapefile.py", line 992, in shapeRecord
    return ShapeRecord(shape=self.shape(i), record=self.record(i))

  File "D:\Anaconda3\lib\site-packages\shapefile.py", line 961, in record
    return self.__record(oid=i)

  File "D:\Anaconda3\lib\site-packages\shapefile.py", line 946, in __record
    value = u(value, self.encoding, self.encodingErrors)

  File "D:\Anaconda3\lib\site-packages\shapefile.py", line 104, in u
    return v.decode(encoding, encodingErrors)

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb0 in position 0: invalid start byte
```

<details>
 <summary>Full environment definition</summary>

### Operating system
Win10 1809 64bit

### Cartopy version
Now is 0.17.0, don't remember witch version when it worked.

### conda list

```
pyshp                     2.0.1
shapely                   1.6.4
python                    3.7.2
......
```

</details>

When I 
```
from shapefile import Reader as shpReader

shpname = r".\map\China_province"
file = shpReader(shpname, encoding="gbk")
provinces_records = file.records()
provinces_shapes = file.shapes()
```
there's no Error.

[...]

PR #2231, opened 2023-08-08
Add encoding support for shapereader

smartlixx:
## Rationale
Based on the discussion of #2041, `shapereader` should support `encoding` in order to read shape files with non-default encoding. 

Currently, for a shape file with GBK encoding `aomen.shp` (containing the polygon for Macau)

```python
import cartopy.io.shapereader as shpreader
shp = shpreader.Reader('aomen.shp')
list(shp.records())
```
will produce error 

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 0: invalid start byte
```
After this PR, the `encoding` can be specified

```python
import cartopy.io.shapereader as shpreader
shp = shpreader.Reader('aomen.shp', encoding='gbk')
list(shp.records())
```
and the results are

```
[<Record: <POLYGON ((113.55 22.217, 113.55 22.214, 113.555 22.206, 113.559 22.206, 113...>, 
{'BIANMA': 820000, 'NAME': '澳门特别行政区', 'ENAME': 'Aomen Tebianxingzhengqu'}, <fields>>]
```

## Implications
None as far as I know

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -131,9 +131,9 @@ class BasicReader:
 
     """
 
-    def __init__(self, filename, bbox=None):
+    def __init__(self, filename, bbox=None, encoding='utf-8'):
         # Validate the filename/shapefile
-        self._reader = reader = shapefile.Reader(filename)
+        self._reader = reader = shapefile.Reader(filename, encoding=encoding)
         if reader.shp is None or reader.shx is None or reader.dbf is None:
             raise ValueError("Incomplete shapefile definition "
                              "in '%s'." % filename)
@@ -186,10 +186,10 @@ class FionaReader:
 
     """
 
-    def __init__(self, filename, bbox=None):
+    def __init__(self, filename, bbox=None, encoding=None):
         self._data = []
 
-        with fiona.open(filename) as f:
+        with fiona.open(filename, encoding=encoding) as f:
             if bbox is not None:
                 assert len(bbox) == 4
                 features = f.filter(bbox=bbox)


lgolston:
There was a [previous suggestion](https://github.com/SciTools/cartopy/issues/1282#issuecomment-592346078) to pass kwargs upstream, but explicitly writing encoding as done here looks right to me to ensure compatibility between Basic and Fiona readers. This will also help close other open issues #739 and #1282 related to handling encoding.

lib/cartopy/io/shapereader.py
```diff
@@ -131,9 +131,9 @@ class BasicReader:
 
     """
 
-    def __init__(self, filename, bbox=None):
+    def __init__(self, filename, bbox=None, encoding='utf-8'):
```
greglucas, line 134: I don't love that the defaults between the two readers isn't the same here... I think the `**kwargs` solution referenced by @lgolston would help by only sending along the encoding when a user explicitly sets it without needing to worry about the default between Fiona and Pyshp.

lib/cartopy/io/shapereader.py
```diff
@@ -186,10 +186,10 @@ class FionaReader:
 
     """
 
-    def __init__(self, filename, bbox=None):
+    def __init__(self, filename, bbox=None, encoding=None):
```
lgolston, line 189:
```suggestion
    def __init__(self, filename, bbox=None, encoding='utf-8'):
```
Thinking about the comment from @greglucas, there are a lot of different options since the two libraries do slightly different things (pyshp raises a TypeError if encoding=None is passed; fiona attempts to detect the encoding by default while pyshp does not; and both accept different kwargs other than encoding). I believe the simplest and most compatible change would be to simply make utf-8 the default for both. This also does not seem to lose much functionality for fiona; for example I tried reading a gbk encoded shapefile and while fiona did not raise an error, it also did not actually read it correctly without specifying the encoding.

greglucas: I don't think we should restrict users to only using utf-8 by default. If users had previously installed Fiona and were reading in other encodings, they would have been able to use the data, but if we force the encoding now then that would break them. I guess there is a question of whether we should call one of these libraries our defacto standard and go with that and then manipulate keyword arguments to be compatible with that...

I still think passing `**kwargs` through is the most flexible and least amount of work, very similar to what the stock_img pr is doing by forwarding them on to the underlying function: https://github.com/SciTools/cartopy/pull/2230

PR #2236, opened 2023-08-16
Add documentation, kwargs for Reader

lgolston:
## Rationale

There was some discussion on how to best implement support for the encoding keyword for the BasicReader and FionaReader interfaces in #2231, and passing it through kwargs was suggested. There is also a generally lack of documentation on 'Reader' that automatically selects between the two (currently the docs just say "alias of FionaReader").

## Implications

Clarifies use of shapereader.Reader and passes kwargs (notably encoding) upstream. bbox is also now passed to PyShp.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -135,7 +135,8 @@ class BasicReader:
 
     def __init__(self, filename, bbox=None, **kwargs):
         # Validate the filename/shapefile
-        self._reader = reader = shapefile.Reader(filename, bbox=bbox, **kwargs)
+        self._reader = reader = shapefile.Reader(filename, **kwargs)
+        self._bbox = bbox
         if reader.shp is None or reader.shx is None or reader.dbf is None:
             raise ValueError("Incomplete shapefile definition "
                              "in '%s'." % filename)
@@ -160,7 +161,7 @@ class BasicReader:
         :meth:`~Record.geometry` method.
 
         """
-        for shape in self._reader.iterShapes():
+        for shape in self._reader.iterShapes(bbox=self._bbox):
             # Skip the shape that can not be represented as geometry.
             if shape.shapeType != shapefile.NULL:
                 yield sgeom.shape(shape)
@@ -170,9 +171,10 @@ class BasicReader:
         Return an iterator of :class:`~Record` instances.
 
         """
+
         # Ignore the "DeletionFlag" field which always comes first
         fields = self._reader.fields[1:]
-        for shape_record in self._reader.iterShapeRecords():
+        for shape_record in self._reader.iterShapeRecords(bbox=self._bbox):
             attributes = shape_record.record.as_dict()
             yield Record(shape_record.shape, attributes, fields)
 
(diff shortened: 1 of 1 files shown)


lib/cartopy/io/shapereader.py
```diff
@@ -123,17 +123,19 @@ def __init__(self, geometry, attributes):
...
+    `Readme <https://pypi.org/project/pyshp/>`_ for more information.
 
     The primary methods used on a BasicReader instance are
     :meth:`~cartopy.io.shapereader.BasicReader.records` and
     :meth:`~cartopy.io.shapereader.BasicReader.geometries`.
 
     """
 
-    def __init__(self, filename, bbox=None):
+    def __init__(self, filename, **kwargs):
```
greglucas, line 136:
```suggestion
    def __init__(self, filename, bbox=None, **kwargs):
```

lib/cartopy/io/shapereader.py
```diff
@@ -123,17 +123,19 @@ def __init__(self, geometry, attributes):
...
         # Validate the filename/shapefile
-        self._reader = reader = shapefile.Reader(filename)
+        self._reader = reader = shapefile.Reader(filename, **kwargs)
```
greglucas, line 138:
```suggestion
        self._reader = reader = shapefile.Reader(filename, bbox=bbox, **kwargs)
```

lib/cartopy/io/shapereader.py
```diff
@@ -253,10 +258,23 @@ def records(self):
                                item.items() if key != 'geometry'})
 
 
-if _HAS_FIONA:
-    Reader = FionaReader
-else:
-    Reader = BasicReader
+def Reader(filename, bbox=None, **kwargs):
```
greglucas, line 261: Is there a reason to make this change as well? Now `Reader` is a function rather than a class.

lib/cartopy/io/shapereader.py
lgolston: Thanks! This led to finding that PyShp has supporting the bounding box filtering for a while now. Since BasicReader uses lazy loading, bbox moved from here to the records and geometries calls.

lib/cartopy/io/shapereader.py
lgolston: I don't think Reader being an alias for BasicReader / FionaReader is ideal. Using a function avoids this and also adds the docstring.

greglucas: > I don't think Reader being an alias for BasicReader / FionaReader is ideal.

I agree it isn't ideal. But, I personally would prefer that to a function that looks like a class, I won't block on it, but I also won't approve it either.

Another way of doing this would have been to create a `class Reader` that stores an instance of `self._reader = FionaReader() if _HAS_FIONA else BasicReader()` and then dispatches out the methods to whatever `self._reader` it contains, but that seems like more work than is necessary as well unless we start getting more readers.

Also, I think you can add a docstring right below the `Reader` definition like we do for the configuration dictionary: https://github.com/SciTools/cartopy/blob/96ee0a3b4ab134b58e58020620f88624c3debf3a/lib/cartopy/__init__.py#L23-L31
but maybe to get that to work you'd need to put it all on one line?
```python
Reader = FionaReader if _HAS_FIONA else BasicReader
"""Explanation of Reader"""
```

lgolston: Thanks for the tip on adding a docstring to the existing Reader line! I like that solution, which has now been made.

greglucas (approved):
I like this approach! The API docs render nicely too.

lgolston:
Thanks also @smartlixx and @wqshen for helping spur this along!

PR #2239, opened 2023-08-23
Input filename for shapefile.Reader() should be a string

JouleCai:
## Rationale

The input full path name for the function "shapefile.Reader()" should be a string. However, the file name get from e.g. natural_earth() is a pathlib.Path object. This causes an error when reading the data file. 

Change 
        self._reader = reader = shapefile.Reader(filename)
to
        self._reader = reader = shapefile.Reader(str(filename))
on line 136 in io/shapereader.py.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -133,7 +133,7 @@ class BasicReader:
 
     def __init__(self, filename, bbox=None):
         # Validate the filename/shapefile
-        self._reader = reader = shapefile.Reader(filename)
+        self._reader = reader = shapefile.Reader(str(filename))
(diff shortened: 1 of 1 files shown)


lgolston:
Maybe this should be applied only for PyShp < 2.3.0? GeospatialPython/pyshp#233 added support for path objects

greglucas:
> Maybe this should be applied only for PyShp < 2.3.0? https://github.com/GeospatialPython/pyshp/pull/233 added support for path objects

I think the solution here is for users to upgrade physhp to 2.3.0+? I'm not sure we should be version-gating on this.

rcomer:
Thank you for your work on this @JouleCai, but we decided to increase the required version of pyshp instead.  See #2290.

PR #2249, opened 2023-09-23
Make GridLiner into an Artist

rcomer:
## Rationale

Thinking about https://github.com/SciTools/cartopy/issues/2246#issuecomment-1731581546, it is not intuitive to me that `Gridliner._draw_gridliner` should be called as part of finding the `GeoAxes`'s bbox, since this method does not seem to update the `GeoAxes`'s position.  I tried moving the call so it's only used at draw and not for finding the bbox, and the only test that failed uses `fig.tight_layout()` before the first draw.  Clearly `tight_layout` cannot create space for artists that do not yet exist.

I wonder if a better approach is to have the `GridLiner` itself be an artist, then it can interact with the Matplotlib layouts and draw machinery on its own terms.  This PR is my attempt at that.  I went looking at the container artists in Matplotlib's [offsetbox.py](https://github.com/matplotlib/matplotlib/blob/v3.8.0/lib/matplotlib/offsetbox.py) for ideas about how to do it.  The `LineCollection` and `Text` objects are no longer added to the axes directly, but are children of the `Gridliner`.

Fixes #2246.  Also closes #1658 as we can now remove the `GridLiner` using the standard `Artist.remove` method.  I can add tests for those but I wanted to get some feedback on this first in case I have gone completely off piste!

This also seems to help #2247 - I only see the busy kernel when I interact with the plot.

## Implications

If user code is relying on the fact that the `LineCollection` and `Text` objects were added directly to the axes, I guess this will break that.

PR #2252, opened 2023-09-28
Gridliner: don't destroy and recreate artists

rcomer:
## Rationale

@greglucas has asked for this at least twice!

When re-drawing the `GridLiner`, update its existing `LineCollection` and `Text` objects rather than creating new ones.  This was trivial for the `LineCollection`s but a bit more complicated for the labels since we might have different numbers of them e.g. after zooming.  Clearly Matplotlib's `Axis` has already solved that problem so I got my inspiration from [there](https://github.com/matplotlib/matplotlib/blob/e3a5cee92bf10f07608f9e638f8d586ba81f16c7/lib/matplotlib/axis.py#L1620-L1626).  We now keep two lists of labels: `_all_labels` is the list of labels that have been drawn at some point.  `_labels` is the list that were used at the most recent draw (and still gets cleared out and repopulated each time).

Goes on top of #2249.  ~Only the third commit is new.~

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -434,6 +434,9 @@ class Gridliner(matplotlib.artist.Artist):
         self.yline_artists = []
 
         # List of all labels (Label objects)
+        self._all_labels = []
+
+        # List of active labels (used in current draw)
         self._labels = []
 
         # Draw status
@@ -609,6 +612,25 @@ class Gridliner(matplotlib.artist.Artist):
 
         return True
 
+    def _generate_labels(self):
+        """
+        A generator to yield as many labels as needed, re-using existing ones
+        where possible.
+        """
+        for label in self._all_labels:
+            yield label
+
+        while True:
+            # Ran out of existing labels.  Create some empty ones.
+            new_artist = matplotlib.text.Text()
+            new_artist.set_figure(self.axes.figure)
+            new_artist.axes = self.axes
+
+            new_label = Label(new_artist, None, None, None)
+            self._all_labels.append(new_label)
+
+            yield new_label
+
     def _draw_gridliner(self, nx=None, ny=None, renderer=None):
         """Create Artists for all visible elements and add to our Axes.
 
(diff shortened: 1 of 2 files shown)


rcomer:
Putting this one in draft as it will need rebasing once #2249 is merged.

greglucas (approved):
I tested this out locally zooming and panning around and it seems quite responsive to me, nice work!

One minor pre-commit failure to fix before merging.

rcomer:
I broke the minimum versions tests.  In fact I broke them at #2249: `Figure.draw_without_rendering` and the _layout_ keyword for `figure`, etc. were introduced at mpl v3.5.  Should I fix them all here or make a separate PR?

rcomer:
I note that mpl v3.5 is two years old next month, so the easiest fix would be to move the pin.

greglucas:
> I broke the minimum versions tests. In fact I broke them at https://github.com/SciTools/cartopy/pull/2249: Figure.draw_without_rendering and the layout keyword for figure, etc. were introduced at mpl v3.5. Should I fix them all here or make a separate PR?

:( I should have checked that during the review but I was only looking at the ones without min-version specified. That is the problem when you have external issues causing red x's throughout the test suite.

I'd say it should probably be a separate commit, but I'm fine having it in this PR if that is easier than a new PR.

> I note that mpl v3.5 is two years old next month, so the easiest fix would be to move the pin.
The problem calls are only in the tests.

I'd say that bumping the pin is fine then. We are not going to release within the next month. Note that you should also probably remove any other MPL3_4 checks there are (if any) and remove that code too, and if so then a new PR might be worthwhile.

dopplershift:
:+1: to what @greglucas said above.

rcomer:
I decided to put it in a separate PR #2258

rcomer:
I think this one is now ready to go.  The only change since @greglucas' approval is the flake8 fix and I have also checked that all the test runs have exactly ten failures.

greglucas:
Still looks good to me. I verified the test runs as well.

PR #2258, opened 2023-10-04
Bump minimum Matplotlib version to 3.5

rcomer:
## Rationale

As agreed in #2252.  Hopefully the minimum version tests are now no more broken than the others.


lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -497,7 +498,10 @@ def test_gridliner_title_adjust():
     plt.rcParams['axes.titley'] = None
 
     fig = plt.figure(layout='constrained')
-    fig.get_layout_engine().set(h_pad=1/8)
+    if _MPL_36:
+        fig.get_layout_engine().set(h_pad=1/8)
```
rcomer, line 502: The deprecation note on `set_constrained_layout_pads` suggest that the above should work from v3.6, but I have not actually tested.
https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.set_constrained_layout_pads

lib/cartopy/tests/mpl/test_crs.py
```diff
@@ -8,12 +8,12 @@
...
 
 import cartopy.crs as ccrs
-from cartopy.mpl import _MPL_35
+from cartopy.mpl import _MPL_36
 
 
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(
-    filename="igh_land.png", tolerance=0.5 if _MPL_35 else 3.6)
+    filename="igh_land.png", tolerance=0.5 if _MPL_36 else 3.6)
```
rcomer, line 16: CI fails here if the tolerance is set to 0.5 with mpl 3.5.

greglucas: I wouldn't read too much into that. Maybe the image was made with a `main` checkout at the time or something? Our tests are often brittle to these things, so this seems fine to me to bump one version chalking it up to we didn't know exactly which minimum version was needed.

Maybe we should think about testing with all versions of Matplotlib that we support though rather than only the newest and oldest.

greglucas (approved):

dopplershift (approved):

PR #2261, opened 2023-10-06
STY: increase max line length to 88

### LINKED ISSUE #2253: flake8: increase max line length?
### Description

Would Cartopy consider increasing its maximum line length to 88 characters, like Matplotlib did at https://github.com/matplotlib/matplotlib/pull/24893?  I keep finding myself needing to rearrange lines because they are just a few characters over the limit.

rcomer:
## Rationale

Closes #2253.  For now I also updated the `isort` line length to match the `flake8` maximum.  This does re-flow some imports.  I'm not sure there is any advantage to that other than "it's nice to be consistent", so I am slightly leaning towards dropping the second commit, but I thought I should include it for now to see what others think.  Also happy to squash into one commit if we do want the `isort` change.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -19,8 +19,8 @@ import shapely.geometry as sgeom
 
 import cartopy
 from cartopy.crs import PlateCarree, Projection, _RectangularProjection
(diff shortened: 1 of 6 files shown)


QuLogic (approved):
Doesn't seem like too many import changes to be a problem.

PR #2263, opened 2023-10-07
MNT: Move cartopy feature download script into the package

### LINKED ISSUE #2227: Don't use generic feature_download executable name
### Description

Since 0.22.0 `tools/cartopy_feature_download.py` gets installed as `/usr/bin/feature_download` which is a far too generic name.

The `cartopy_` prefix should be preserved as per this patch:
```diff
Description: Don't use generic feature_download executable name.
Author: Bas Couwenberg <sebastic@debian.org>

--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ plotting = ["pillow>=6.1.0", "scipy>=1.3
 test = ["pytest>=5.1.2", "pytest-mpl>=0.11", "pytest-xdist", "pytest-cov", "coveralls"]
 
 [project.scripts]
-feature_download = "tools.cartopy_feature_download.py:__main__"
+cartopy_feature_download = "tools.cartopy_feature_download.py:__main__"
 
 [project.urls]
 documentation='https://scitools.org.uk/cartopy/docs/latest/'
```

### Operating system

Debian unstable

### Cartopy version

0.22.0

</details>

greglucas:
Move it from the tools directory into the feature module and make it an executable module by calling it __main__.py

It can be invoked with `python -m cartopy.feature.download` now.

Does this location make sense? This adds a new download module so that the execution has download in the name which I think helps with discoverability/readability (`python -m cartopy.feature -h` isn't as readable IMO if we put `__main__.py` in the top-level), I'm open to other suggestions as well.

closes #2227

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ plotting = ["pillow>=6.1.0", "scipy>=1.3.1"]
 test = ["pytest>=5.1.2", "pytest-mpl>=0.11", "pytest-xdist", "pytest-cov", "coveralls"]
 
 [project.scripts]
-feature_download = "tools.cartopy_feature_download.py:__main__"
+cartopy_feature_download = "cartopy.feature.download.__main__:main"
 
 [project.urls]
 documentation='https://scitools.org.uk/cartopy/docs/latest/'
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -72,7 +72,9 @@ jobs:
         shell: bash
         run: |
           # Check that the downloader tool at least knows where to get the data from (but don't actually download it)
-          python tools/cartopy_feature_download.py gshhs physical --dry-run
+          python -m cartopy.feature.download gshhs physical --dry-run
+          # It should also be available as a script
+          cartopy_feature_download gshhs physical --dry-run
           CARTOPY_GIT_DIR=$PWD
           pytest -ra -n 4 \
               --color=yes \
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -55,9 +55,6 @@ ows = ["OWSLib>=0.20.0", "pillow>=6.1.0"]
 plotting = ["pillow>=6.1.0", "scipy>=1.3.1"]
 test = ["pytest>=5.1.2", "pytest-mpl>=0.11", "pytest-xdist", "pytest-cov", "coveralls"]
 
-[project.scripts]
```
stefraynaud, line 58: In addition to the Why not leaving it also as executable script that is installed with the package, typically named `cartopy_feature_download`? Maybe you must declare a `main` function in the `__main__` module for that.

greglucas: Yeah, I think you're right, we need to define a function that can be called from the script for that. I moved the content in the if block inside a separate `main()` function like you suggested.

Now you can access this script via: `python -m cartopy.feature.download -h` or `cartopy_feature_download gshhs physical --dry-run`

I guess this just brings up the question of whether this installed script is desired or not. I don't have a major preference either way, but would probably slightly lean towards not including it in the scripts section since it is not a critical piece of cartopy. But, it looks like you @stefraynaud were the one that added this as an installation piece, so maybe it is useful for you as a standalone script? Previously, I think it made sense because it was only accessible via a git clone of the repository to access the tools directory, but now you can execute the module which is installed with the package as well.

greglucas: Agree with the first part on this function being useful without git cloning.

My second part was just suggesting to now tell users from wherever you are type: `python -m cartopy.feature.download -h`

But, I'm fine either way with including the install script, so left it in for now.

QuLogic (approved):
Seems reasonable to me.

PR #2264, opened 2023-10-07
Add xfails to current owslib failures

greglucas:
This adds xfails to all of the 10 current failures in the test suite. It also adds the failing documentation examples to the expected failures there as well.

In addition, running the examples locally I noticed that there is now a 404 from the Vesta image example, so another commit to change the URL for that example.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -93,6 +93,12 @@ subsection_order = ExplicitOrder(['../../examples/lines_and_polygons',
 sphinx_gallery_conf = {
     'capture_repr': (),
     'examples_dirs': ['../../examples'],
+    # NASA wmts servers are returning bad content metadata
+    "expected_failing_examples": [
+        '../../examples/web_services/reprojected_wmts.py',
+        '../../examples/web_services/wmts.py',
+        '../../examples/web_services/wmts_time.py',
+    ],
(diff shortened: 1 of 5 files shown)


docs/source/conf.py
```diff
@@ -93,6 +93,13 @@
 sphinx_gallery_conf = {
     'capture_repr': (),
     'examples_dirs': ['../../examples'],
+    # NASA wmts servers are returning bad content metadata
+    "expected_failing_examples": [
+        '../../examples/scalar_data/eyja_volcano.py',
```
greglucas, line 98: It seems like these two are flaky right now. I get failures when I remove them. I get failures when I add them :) Not sure if there is a way to say something is flaky in sphinx gallery?
```suggestion
        '../../examples/scalar_data/eyja_volcano.py',
        '../../examples/web_services/image_tiles.py',
```

rcomer: Apparently the Stamen tiles have moved https://stackoverflow.com/questions/77248120/stamen-terrain-map-not-working-in-cartopy

greglucas: Interesting, I did not realize that was going away. Let's follow that up in a separate PR then and leave the docs failing here for now.

rcomer (approved):

PR #2265, opened 2023-10-08
DOC: Change Stamen image tiles over to Google

greglucas:
This changes examples that used Stamen over to Google. The Stamen versions are no longer accessible without an API key, so switching over to a public tile server is the easiest fix for now.

### DIFF
--- a/examples/scalar_data/eyja_volcano.py
+++ b/examples/scalar_data/eyja_volcano.py
@@ -3,7 +3,7 @@ Map tile acquisition
 --------------------
 
 Demonstrates cartopy's ability to draw map tiles which are downloaded on
-demand from the Stamen tile server. Internally these tiles are then combined
+demand from the Google tile server. Internally these tiles are then combined
 into a single image and displayed in the cartopy GeoAxes.
 
 """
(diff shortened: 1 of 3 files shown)


examples/scalar_data/eyja_volcano.py
```diff
@@ -15,19 +15,19 @@
...
     fig = plt.figure()
 
     # Create a GeoAxes in the tile's projection.
-    ax = fig.add_subplot(1, 1, 1, projection=stamen_terrain.crs)
+    ax = fig.add_subplot(1, 1, 1, projection=google_terrain.crs)
 
     # Limit the extent of the map to a small longitude/latitude range.
     ax.set_extent([-22, -15, 63, 65], crs=ccrs.Geodetic())
 
     # Add the Stamen data at zoom level 8.
```
rcomer, line 29:
```suggestion
    # Add the Google data at zoom level 8.
```
Also in the module docstring.

rcomer (commented):
Line 72 of the ellipse example also mentions Stamen in the comment.

examples/scalar_data/eyja_volcano.py
greglucas: Oops, thanks! Just pushed an update with both of those.

rcomer: This line still says "Stamen"... 👀

greglucas: Uggh, sorry! Not enough coffee yet apparently.

rcomer (approved):

rcomer:
Cycling to re-run tests...

rcomer:
Circle did not re-start 😕

rcomer:
OK Circle now re-ran but it looks like it didn't pick up the changes from #2264.  @greglucas do you fancy rebasing?

greglucas:
> OK Circle now re-ran but it looks like it didn't pick up the changes from https://github.com/SciTools/cartopy/pull/2264. @greglucas do you fancy rebasing?

Done, and all looks good now.

> Circle did not re-start 😕

I think you might be able to log into CircleCI and manually click the re-run button on the job, but I think I also remember something about CircleCI not referencing the latest HEAD of upstream in that case, so the other updates that were merged wouldn't be brought along, but I am not certain of that.

rcomer:
> I think you might be able to log into CircleCI and manually click the re-run button on the job, but I think I also remember something about CircleCI not referencing the latest HEAD of upstream in that case, so the other updates that were merged wouldn't be brought along, but I am not certain of that.

Yep, that's what I did after the close-reopen didn't restart it, but then discovered what you say about it not getting the updates from upstream.

Great to have a full set of green ticks!

PR #2266, opened 2023-10-08
FIX: Change Stamen url over to Stadia Maps

greglucas:
The Stamen tiles are now being served by Stadia Maps. Stadia maps requires an API key, and has different naming convention for their tiles, so update both of those things here.

xref: https://stamen.com/stamen-x-stadia-the-end-of-the-road-for-stamens-legacy-map-tiles/

I could also see making a new `Stadia` class for this instead, and deprecating `Stamen`? I'm not sure what the best approach is, especially because the `Stamen` class is not even usable in its current form, so not sure it even needs a deprecation...

Apparently we don't have any tests for this... I did make an API key and test locally with the examples and those did seem to work, but depending on the route we go we should probably add a few tests at least.

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -322,26 +322,31 @@ class OSM(GoogleWTS):
 
 class Stamen(GoogleWTS):
     """
-    Retrieves tiles from maps.stamen.com. Styles include
+    Retrieves tiles from stadiamaps.com. Styles include
+    ``stamen_terrain-background``, ``stamen_terrain``,
+    ``stamen_toner``, and ``stamen_watercolor``. Previously,
+    these were available as
     ``terrain-background``, ``terrain``, ``toner`` and ``watercolor``.
 
     For a full reference on the styles available please see
-    http://maps.stamen.com. Of particular note are the sub-styles
-    that are made available (e.g. ``terrain`` and ``terrain-background``).
-    To determine the name of the particular [sub-]style you want,
-    follow the link on http://maps.stamen.com to your desired style and
-    observe the style name in the URL. Your style name will be in the
-    form of: ``http://maps.stamen.com/{STYLE_NAME}/#9/37/-122``.
+    https://docs.stadiamaps.com/themes/. Of particular note are the sub-styles
+    that are made available (e.g. ``stamen_terrain`` and
+    ``stamen_terrain_background``).
 
     Except otherwise noted, the Stamen map tile sets are copyright Stamen
-    Design, under a Creative Commons Attribution (CC BY 3.0) license.
+    Design, under a Creative Commons Attribution (CC BY 4.0) license.
 
     Please see the attribution notice at http://maps.stamen.com on how to
     attribute this imagery.
+    For Toner and Terrain:
+    Map tiles by Stamen Design, under CC BY 4.0. Data by OpenStreetMap, under ODbL.
+    For Watercolor:
+    Map tiles by Stamen Design, under CC BY 4.0. Data by OpenStreetMap, under CC BY SA.
 
     References
     ----------
 
+     * https://docs.stadiamaps.com/themes/
      * http://mike.teczno.com/notes/osm-us-terrain-layer/background.html
      * http://maps.stamen.com/
      * https://wiki.openstreetmap.org/wiki/List_of_OSM_based_Services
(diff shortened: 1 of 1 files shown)


lib/cartopy/io/img_tiles.py
```diff
@@ -350,23 +355,33 @@ class Stamen(GoogleWTS):
...
-          'toner-lines':        {'extension': 'png', 'opaque': False},
-          'toner-lite':         {'extension': 'png', 'opaque': True},
-          'watercolor':         {'extension': 'jpg', 'opaque': True},
+          'stamen_terrain':            {'extension': 'png', 'opaque': True},
+          'stamen_terrain_background': {'extension': 'png', 'opaque': True},
+          'stamen_terrain_labels':     {'extension': 'png', 'opaque': False},
+          'stamen_terrain_lines':      {'extension': 'png', 'opaque': False},
+          'stamen_toner_background':   {'extension': 'png', 'opaque': True},
+          'stamen_toner':              {'extension': 'png', 'opaque': True},
+          'stamen_toner_hybrid':       {'extension': 'png', 'opaque': False},
```
lgolston, line 371: Toner hybrid is now deprecated according to https://docs.stadiamaps.com/map-styles/stamen-toner/

lib/cartopy/io/img_tiles.py
```diff
@@ -350,23 +355,33 @@ class Stamen(GoogleWTS):
     """
 
     def __init__(self, style='toner',
-                 desired_tile_form=None, cache=False):
-
+                 desired_tile_form=None, cache=False,
+                 apikey=None):
+        if apikey is None:
+            raise ValueError("An apikey is now required to use Stamen maps. "
+                             "Visit https://maps.stadiamaps.com/ for more details.")
```
lgolston, line 362: Dead URL? Suggest instead: https://docs.stadiamaps.com/authentication/

lib/cartopy/io/img_tiles.py
```diff
@@ -350,23 +355,33 @@ class Stamen(GoogleWTS):
...
+          'stamen_watercolor':         {'extension': 'jpg', 'opaque': True},
         }
 
+        # Stadia Maps has a different naming convention for the styles
+        # they prepend stamen_ and dashes are now underscores
+        if "stamen" not in style:
+            style = f"stamen_{style.replace('-', '_')}"
+            warnings.warn("Stamen maps now require the stamen_ prefix for "
+                          f"the style name, please update to the new style: {style}")
+
```
lgolston, line 384: This gets to your question about potentially switching the class name over to Stadia, but the name replacement here does not seem ideal since it would prevent using styles now available like Alidade Smooth

lgolston: Yes, that seems preferable to me. Since the user has to get an api key from Stadia, might as well make all the styles available!

lib/cartopy/io/img_tiles.py
ianthetechie: Yes, I can confirm this is the correct URL. `maps.stadiamaps.com` is not a valid URL.

lib/cartopy/io/img_tiles.py
ianthetechie: Yeah, this is the approach taken by OpenLayers, Leaflet Providers, and more. It's already a breaking change, so changing the name makes sense.

You can find a full list of our styles here: https://docs.stadiamaps.com/themes/. Some notes:

* Stamen styles are (currently) the only ones with multiple variations per style. All others have just one identifier.
* Alidade Satellite is _not_ currently available as pure raster tiles; we expect that to change in the future, but probably doesn't make sense to include that one now.

lib/cartopy/io/img_tiles.py
```diff
@@ -322,26 +322,31 @@ def _image_url(self, tile):
...
     Except otherwise noted, the Stamen map tile sets are copyright Stamen
-    Design, under a Creative Commons Attribution (CC BY 3.0) license.
+    Design, under a Creative Commons Attribution (CC BY 4.0) license.
 
     Please see the attribution notice at http://maps.stamen.com on how to
     attribute this imagery.
+    For Toner and Terrain:
+    Map tiles by Stamen Design, under CC BY 4.0. Data by OpenStreetMap, under ODbL.
+    For Watercolor:
+    Map tiles by Stamen Design, under CC BY 4.0. Data by OpenStreetMap, under CC BY SA.
```
ianthetechie, line 344: The updated Stamen styles have some updated attribution requirements, which you can find here: https://docs.stadiamaps.com/attribution/. We've also shortened the text for each required attribution, since there are more of them.

Attributions needed for all Stadia Maps styles
* Stadia Maps
* OpenMapTiles
* OpenStreetMap

Additionally, Stamen styles require a credit to Stamen. The page linked above has recommended text.

ianthetechie (changes requested):
Hey, thanks for opening this PR @greglucas! I'm one of the cofounders of Stadia Maps, and it looks like we completely missed CartoPy in our scan for open-source projects to open PRs with!

I just added a few comments for completeness in addition to what @lgolston already mentioned.

lib/cartopy/io/img_tiles.py
greglucas: Thanks for chiming in here! I just pinged you on #2269 which I opened with the new style class for this. I put in the updated attribution text, let me know if you want more in the docstring or I missed something.

PR #2267, opened 2023-10-08
MNT: Update gitignore items

greglucas:
This removes some old files that are no longer being produced and also ignores some new files that may come in via doc builds. In particular the ne_10m_bathymetry example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,22 +7,13 @@ dist
 
 # cartopy build files
 lib/cartopy/_version.py
-lib/cartopy/trace.cpp
-lib/cartopy/trace.so
-lib/cartopy/trace.*.so
-lib/cartopy/trace.pyd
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,22 +7,12 @@ dist
 
 # cartopy build files
 lib/cartopy/_version.py
-lib/cartopy/trace.cpp
-lib/cartopy/trace.so
-lib/cartopy/trace.*.so
-lib/cartopy/trace.pyd
```
QuLogic, line 13: There should be `pyd` files on Windows?

greglucas: Ahh, yes. Thanks, just added `*.pyd` to the gitignore.

QuLogic (approved):

PR #2269, opened 2023-10-12
ENH: Add Stadia Maps image tile server class

greglucas:
The Stamen tiles are now being served by Stadia Maps. Stadia Maps requires an API key and has different naming conventions for their tiles, so create a new class to interact with the new server. Replaces #2266

xref: https://stamen.com/stamen-x-stadia-the-end-of-the-road-for-stamens-legacy-map-tiles/

Apparently, we don't have any tests for this... I did make an API key and test locally with the examples and those did seem to work.

cc @lgolston @ianthetechie

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -320,6 +320,66 @@ class OSM(GoogleWTS):
         return f'https://a.tile.openstreetmap.org/{z}/{x}/{y}.png'
 
 
+class StadiaMapsTiles(GoogleWTS):
+    """
+    Retrieves tiles from stadiamaps.com.
+
+    For a full reference on the styles available please see
+    https://docs.stadiamaps.com/themes/. A few of the specific styles
+    that are made available are ``alidade_smooth``, ``stamen_terrain`` and
+    ``osm_bright``.
+
+    Using the Stadia Maps API requires including an attribution. Please see
+    https://docs.stadiamaps.com/attribution/ for details.
+
+    For most styles that means including the following attribution:
+
+    `© Stadia Maps <https://www.stadiamaps.com/>`_
+    `© OpenMapTiles <https://openmaptiles.org/>`_
+    `© OpenStreetMap contributors <https://www.openstreetmap.org/about/>`_
+
+    with Stamen styles *additionally* requiring the following attribution:
+
+    `© Stamen Design <https://stamen.com/>`_
+
+    Parameters
+    ----------
+    apikey : str, required
+        The authentication key provided by Stadia Maps to query their APIs
+    style : str, optional
+        Name of the desired style. Defaults to ``alidade_smooth``.
+        See https://docs.stadiamaps.com/themes/ for a full list of styles.
+    resolution : str, optional
+        Resolution of the images to return. Defaults to an empty string,
+        standard resolution (256x256). You can also specify "@2x" for high
+        resolution (512x512) tiles.
+    cache : bool or str, optional
+        If True, the default cache directory is used. If False, no cache is
+        used. If a string, the string is used as the path to the cache.
+    """
+
+    def __init__(self,
+                 apikey,
+                 style="alidade_smooth",
+                 resolution="",
+                 cache=False):
+        super().__init__(cache=cache, desired_tile_form="RGBA")
+        self.apikey = apikey
+        self.style = style
+        self.resolution = resolution
+        if style == "stamen_watercolor":
+            # Known style that has the jpg extension
+            self.extension = "jpg"
+        else:
+            self.extension = "png"
+
+    def _image_url(self, tile):
+        x, y, z = tile
+        return ("http://tiles.stadiamaps.com/tiles/"
+                f"{self.style}/{z}/{x}/{y}{self.resolution}.{self.extension}"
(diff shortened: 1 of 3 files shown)


lib/cartopy/io/img_tiles.py
```diff
@@ -320,6 +320,57 @@ def _image_url(self, tile):
...
+        See https://docs.stadiamaps.com/themes/ for a full list of styles.
+    cache : bool or str, optional
+        If True, the default cache directory is used. If False, no cache is
+        used. If a string, the string is used as the path to the cache.
+    """
+
+    def __init__(self, apikey, style='alidade_smooth', cache=False):
+        super().__init__(cache=cache)
+        self.apikey = apikey
+        if style in ("alidade_satellite", "stamen_watercolor"):
```
ianthetechie, line 360: Some notes on `alidade_satellite` that'd be really easy to miss in our docs...

1. We don't actually support this as a full rendered style at the moment (this style ID won't actually work).
2. The JPGs that do exist under a slightly different URL are for the raw imagery imagery only and are not a standard 256x256 tile but rather @2x 512 tiles, which are intended to be used in conjuction with a vector style.

The terms on this raw imagery are also a little different per our upstream provider (some details here: https://docs.stadiamaps.com/map-styles/alidade-satellite/), so I TL;DR I would suggest _not_ adding this condition for the satellite style or mentioning it. We do intend to add this as a "normal" raster style in the future with the other layers on top, but it's probably best to just remove this for now.

lib/cartopy/io/img_tiles.py
```diff
@@ -320,6 +320,57 @@ def _image_url(self, tile):
...
+            # Known styles that have the jpg extension
+            self.extension = "jpg"
+        else:
+            self.extension = "png"
+        self.style = style
+
+    def _image_url(self, tile):
+        x, y, z = tile
+        return ("http://tiles.stadiamaps.com/tiles/"
+                f"{self.style}/{z}/{x}/{y}.{self.extension}"
```
ianthetechie, line 370: Dumb question more than anything as I regrettably haven't used CartoPy myself, but are @2x (aka retina or 512 tiles) useful? If so, we _do_ support those on all styles except for Stamen Watercolor by adding `@2x` right before the extension.

ianthetechie (commented):
Just one minor change request and question out of my ignorance :) Thanks for the ping @greglucas!

lgolston:
Thanks @greglucas! This will be a nice enhancement switching things over to Stadia. It looks good to me, except I think the desired_tile_form input should be kept?

lib/cartopy/io/img_tiles.py
greglucas: Thanks for the info, I removed the satellite option for now.

lib/cartopy/io/img_tiles.py
greglucas: 👍 It is easy enough to add, so I added an extension parameter as well now and it looks good locally to me.

greglucas:
> Thanks @greglucas! This will be a nice enhancement switching things over to Stadia. It looks good to me, except I think the desired_tile_form input should be kept?

Yeah, I wasn't entirely sure what the `desired_tile_form` was used for and how it was chosen for the various tile styles. I don't see any mention of anything like RGB / RGBA on the Stadia or Stamen website, so I figured we would just leave it out for now and if a user wants a specific form we can add it later? Most of the other classes don't have the desired_tile_form either which is why I dropped it.

lgolston:
> > Thanks @greglucas! This will be a nice enhancement switching things over to Stadia. It looks good to me, except I think the desired_tile_form input should be kept?
> 
> Yeah, I wasn't entirely sure what the `desired_tile_form` was used for and how it was chosen for the various tile styles. I don't see any mention of anything like RGB / RGBA on the Stadia or Stamen website, so I figured we would just leave it out for now and if a user wants a specific form we can add it later? Most of the other classes don't have the desired_tile_form either which is why I dropped it.

I don't see it explicitly mentioned but plotting, for instance, stamen_toner_labels without having desired_tile_form gives a warning from `PIL\Image.py:992: UserWarning: Palette images with Transparency expressed in bytes should be converted to RGBA images`. It seems a necessary if anyone is using the sub-styles that are just overlaying lines/labels and not a full image

greglucas:
Thanks for identifying that @lgolston. Rather than trying to special-case the various tile styles, I think it might be easier to just go straight to RGBA for all of Stadia so we can also accommodate potential future styles as well without needing to explicitly handle each style's version. This does add an extra alpha channel to the other images when they aren't explicitly needed I guess, but that doesn't seem like too big of a deal to me to simplify the code here for flexibility...

lgolston:
> Thanks for identifying that @lgolston. Rather than trying to special-case the various tile styles, I think it might be easier to just go straight to RGBA for all of Stadia so we can also accommodate potential future styles as well without needing to explicitly handle each style's version. This does add an extra alpha channel to the other images when they aren't explicitly needed I guess, but that doesn't seem like too big of a deal to me to simplify the code here for flexibility...

Sounds good! That idea has actually been floated by @pelson before as well, in https://github.com/SciTools/cartopy/issues/762#issuecomment-452962755_

PR #2272, opened 2023-10-13
MNT: Add Python 3.12 to the tests and enable release testing

greglucas:
This adds Python 3.12 to the CI runs and also adds a CI: build wheels tag that can be added to PRs to build the wheels for testing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,14 @@ on:
   release:
     types:
       - published
+  # Also allow running this action on PRs if requested by applying the
+  # "Run cibuildwheel" label.
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - labeled
 
 jobs:
   build_sdist:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -49,6 +57,17 @@ jobs:
...
+      github.event_name == 'release' ||
+      (github.event_name == 'pull_request' && (
+        (
+          github.event.action == 'labeled' &&
+          github.event.label.name == 'CI: build wheels'
+        ) ||
+        contains(github.event.pull_request.labels.*.name,
+                'CI: build wheels')
+      )
+      )
```
QuLogic, line 70: Alignment is a bit off
```suggestion
        )
      )
```

.github/workflows/release.yml
```diff
@@ -73,7 +92,7 @@ jobs:
         env:
           CIBW_BUILD: "cp39-* cp310-* cp311-* cp312-*"
           # Skip 32 bit windows builds and musllinux due to lack of numpy wheels
```
QuLogic, line 94:
```suggestion
          # Skip 32 bit builds and musllinux due to lack of numpy wheels
```

QuLogic (approved):

.github/workflows/release.yml
greglucas: fixed up both of those. Thanks.

PR #2273, opened 2023-10-14
MNT: Use ruff for formatting checks

greglucas:
Moves from flake8 and isort to ruff for the formatting checks. Also updates some of the sorted and formatted imports that were flagged by ruff. Ruff is significantly faster than flake8 / isort and we can also start adding additional checks later if we want. I looked at adding "D", but that is quite intimidating to update the docstrings right now with ~500-1k failures depending on which checks we want.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,10 +25,6 @@ repos:
       - id: no-commit-to-branch
         # Trims trailing whitespace
       - id: trailing-whitespace
-  - repo: https://github.com/pycqa/flake8
-    rev: "6.1.0"
-    hooks:
-      - id: flake8
   - repo: https://github.com/codespell-project/codespell
     rev: "v2.2.6"
     hooks:
@@ -40,9 +36,8 @@ repos:
     hooks:
       - id: sort-all
         types: [file, python]
-  - repo: https://github.com/PyCQA/isort
-    rev: "5.12.0"
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: 'v0.1.5'
     hooks:
-      - id: isort
-        types: [file, python]
-        args: [--filter-files]
+    - id: ruff
+      args: [--fix]
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -17,8 +17,13 @@ from shapely.geos import geos_version
 import cartopy.crs as ccrs
 from cartopy.mpl import _MPL_36
 from cartopy.mpl.geoaxes import GeoAxes
-from cartopy.mpl.gridliner import (LATITUDE_FORMATTER, LONGITUDE_FORMATTER, Gridliner,
-                                   classic_formatter, classic_locator)
+from cartopy.mpl.gridliner import (
+    LATITUDE_FORMATTER,
+    LONGITUDE_FORMATTER,
+    Gridliner,
+    classic_formatter,
+    classic_locator,
+)
 from cartopy.mpl.ticker import LongitudeFormatter, LongitudeLocator
 
 
--- a/environment.yml
+++ b/environment.yml
@@ -31,7 +31,7 @@ dependencies:
   - sphinx
   - sphinx-gallery
   # Extras
-  - flake8
   - pre-commit
   - pykdtree
+  - ruff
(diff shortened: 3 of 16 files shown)


lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -17,8 +17,13 @@
...
 from cartopy.mpl.geoaxes import GeoAxes
-from cartopy.mpl.gridliner import (LATITUDE_FORMATTER, LONGITUDE_FORMATTER, Gridliner,
-                                   classic_formatter, classic_locator)
+from cartopy.mpl.gridliner import (
+    LATITUDE_FORMATTER,
+    LONGITUDE_FORMATTER,
+    Gridliner,
+    classic_formatter,
+    classic_locator,
+)
```
rcomer, line 26: I think I prefer the old style here, as I like the horizontal separation between modules and their objects.  I can't see a setting for that though, and it isn't a deal-breaker.

rcomer (approved):
Makes sense to me.  I tried it locally and in _flies_.

It looks like ruff will also autofix some white space problems.  Does that mean that some of the smaller hooks are also redundant, or are they still needed for the non-python files?
https://docs.astral.sh/ruff/rules/#warning-w

lib/cartopy/tests/mpl/test_gridliner.py
greglucas: I don't have a strong preference either way, so I was fine applying this. I think that ruff likely takes the black formatting approach and applies it liberally here even if we aren't using black. I also haven't found a setting to override this so I think we are stuck with it if we want to replace isort with ruff.

greglucas:
> It looks like ruff will also autofix some white space problems. Does that mean that some of the smaller hooks are also redundant, or are they still needed for the non-python files?
https://docs.astral.sh/ruff/rules/#warning-w

I think we can probably start slowly moving over to ruff for these. But, I wasn't 100% sure so I didn't apply that here. But, we can definitely start adding more codes as we see fit and removing the redundant checks.

.pre-commit-config.yaml
```diff
@@ -40,9 +36,8 @@ repos:
     hooks:
       - id: sort-all
         types: [file, python]
-  - repo: https://github.com/PyCQA/isort
-    rev: "5.12.0"
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: 'v0.0.292'
```
rcomer, line 40: v0.1.0 just dropped 👀
https://github.com/astral-sh/ruff/releases/tag/v0.1.0

setup.cfg
```diff
@@ -1,6 +0,0 @@
-[flake8]
-max-line-length = 88
```
QuLogic, line 2: Can we keep a flake8 line length config in the interim?

greglucas: Do you want to keep running flake8?

ruff defaults to 88 line length in our E checks:
https://docs.astral.sh/ruff/rules/line-too-long/
https://docs.astral.sh/ruff/settings/#line-length

QuLogic: Yes, I haven't switched to `ruff` locally, and am not sure when I will. I'm sure I will eventually, but no time right now.

greglucas: :( It doesn't seem ideal to keep that around when we don't include flake8 in any of the installation/documentation instructions or checks. But, I did add the file back now.

setup.cfg
```diff
@@ -1,6 +1,2 @@
 [flake8]
 max-line-length = 88
-exclude = \
-    build, \
-    setup.py, \
-    docs/source/conf.py
```
rcomer, line 6: If I run `flake8` with this locally I get
```
./docs/source/conf.py:369:1: E266 too many leading '#' for block comment
./docs/source/conf.py:378:1: E266 too many leading '#' for block comment
./docs/source/conf.py:392:1: E266 too many leading '#' for block comment
```
Is it worth reinstating this exclude?

greglucas: Yeah, good catch, I should have just left that file alone. Just fixed that in the latest push.

.pre-commit-config.yaml
greglucas: Bumped to v0.1.5 now.

rcomer:
Let's do this!

PR #2274, opened 2023-10-15
MNT: Remove beautifulsoup4 from docs requirements

greglucas:
beautifulsoup4 was not used in our docs build. It is brought in by pydata-sphinx-theme, but we can leave that dependency resolution there. We do use beautifulsoup4 in SRTM tiles, but that has nothing to do with the docs installation. Not sure if we want to move that into a different category or create a new one altogether `srtm = ["beautifulsoup4"]`?

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -1,4 +1,3 @@
 sphinx
 sphinx-gallery
-beautifulsoup4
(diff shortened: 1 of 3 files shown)


dopplershift:
I lean towards a new category for the dependency.

dopplershift (approved):

PR #2276, opened 2023-10-16
Update Matplotlib min version in pyproject.toml

rcomer:
## Rationale

I missed this one at #2258.

Does the fact that this pin is specified in the main dependencies here mean that it is impossible to install Cartopy in an environment with an older Matplotlib and therefore this check is redundant?

https://github.com/SciTools/cartopy/blob/53260b6e8aea3233eba232475332918257ff7b9a/lib/cartopy/mpl/__init__.py#L16

I have a vague memory that Matplotlib was once an optional dependency, which could be why it was written to error on import.

### DIFF
--- a/lib/cartopy/mpl/__init__.py
+++ b/lib/cartopy/mpl/__init__.py
@@ -9,8 +9,5 @@ import packaging.version
 
 
 _MPL_VERSION = packaging.version.parse(matplotlib.__version__)
-_MPL_35 = _MPL_VERSION.release[:2] >= (3, 5)
 _MPL_36 = _MPL_VERSION.release[:2] >= (3, 6)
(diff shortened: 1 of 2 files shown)


greglucas:
> Does the fact that this pin is specified in the main dependencies here mean that it is impossible to install Cartopy in an environment with an older Matplotlib and therefore this check is redundant?

Yes that should be the case, so I think you can remove that line.

> I have a vague memory that Matplotlib was once an optional dependency, which could be why it was written to error on import.

You are correct. We updated it here: https://github.com/SciTools/cartopy/pull/1876 which I still think makes sense because a user should probably use pyproj/shapely directly if they don't want the interaction with MPL.

greglucas (approved):

PR #2277, opened 2023-10-18
Matplotlib 3.8 test warning

lgolston:
## Rationale

tests\mpl\conftest.py gives the following warning:  "MatplotlibDeprecationWarning: Auto-close()ing of figures upon backend switching is deprecated since 3.8 and will be removed two minor releases later. To suppress this warning, explicitly call plt.close('all') first."

## Implications

The warning is addressed.

### DIFF
--- a/lib/cartopy/tests/mpl/conftest.py
+++ b/lib/cartopy/tests/mpl/conftest.py
@@ -20,6 +20,7 @@ def mpl_test_cleanup(request):
         plt.switch_backend('Agg')
         stack.enter_context(plt.rc_context())
         yield
+        plt.close('all')
 
 
 def pytest_itemcollected(item):
--- a/lib/cartopy/tests/mpl/conftest.py
+++ b/lib/cartopy/tests/mpl/conftest.py
@@ -15,12 +15,12 @@ def mpl_test_cleanup(request):
     with ExitStack() as stack:
         # At exit, close all open figures and switch backend back to original.
         stack.callback(plt.switch_backend, plt.get_backend())
+        stack.callback(plt.close, 'all')
 
         # Run each test in a context manager so that state does not leak out
         plt.switch_backend('Agg')
         stack.enter_context(plt.rc_context())
         yield
-        plt.close('all')
 
 
 def pytest_itemcollected(item):


lib/cartopy/tests/mpl/conftest.py
```diff
@@ -20,6 +20,7 @@ def mpl_test_cleanup(request):
         plt.switch_backend('Agg')
         stack.enter_context(plt.rc_context())
         yield
+        plt.close('all')
```
QuLogic, line 23: You should make this a `stack.callback` before the backend switch so it runs on context exit.

rcomer:
This is probably a naive question, but: why do we switch backends individually for each test, rather than setting to agg once globally?

lgolston:
> This is probably a naive question, but: why do we switch backends individually for each test, rather than setting to agg once globally?

That is a good question. I just wanted to quiet the warning (which is repeated for every mpl test), but looking at the backend setting and switching now in the tests I do wonder whether what is really necessary. Especially, tests/mpl/__init__.py which switches to tkagg and back.

greglucas:
I think the backend switch every time is there to guard against a test that may choose to use a different backend so then you don't get that leaking out into subsequent tests.

That `tests.mpl.show` function seems like it could be removed... It is not used anywhere other than in commented-out code that looks more like it is for debugging what was going on previously. That can be a follow-up PR to clean that portion up though.

lib/cartopy/tests/mpl/conftest.py
```diff
@@ -15,6 +15,7 @@ def mpl_test_cleanup(request):
     with ExitStack() as stack:
         # At exit, close all open figures and switch backend back to original.
         stack.callback(plt.switch_backend, plt.get_backend())
+        stack.callback(plt.close, 'all')
```
greglucas, line 18: Does it matter whether we close all before or after the switch backend?

lgolston: Yes, if line 17 and 18 are swapped the warning is still given. Since the stack goes in reverse order, I think the flow is that on exit the plot is closed, then backend is switched to original.

QuLogic (approved):

PR #2278, opened 2023-10-18
Skipped test in test_coding_standards

lgolston:
## Rationale

The CI server gives a message like: "SKIPPED [1] ../../../../../opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12/site-packages/cartopy/tests/test_coding_standards.py:78: cartopy installation did not look like a git repo: /opt/hostedtoolcache/Python/3.12.0/x64/lib/python3.12 is not a git repository." suggesting it cannot find the root directory (locally seems to work fine).

## Implications

This PR tries a different approach to finding the git repository location so the test [which checks the license headers] will hopefully not be skipped.


lib/cartopy/tests/test_coding_standards.py
```diff
@@ -29,12 +26,11 @@
...
-# Guess cartopy repo directory of cartopy - realpath is used to mitigate
-# against Python finding the cartopy package via a symlink.
-CARTOPY_DIR = Path(cartopy.__file__).parent.resolve()
-REPO_DIR = Path(os.getenv('CARTOPY_GIT_DIR',
-                          CARTOPY_DIR.parent.parent))
+# Get repo directory for cartopy (https://stackoverflow.com/a/53675112)
+REPO_DIR = subprocess.Popen(['git', 'rev-parse', '--show-toplevel'],
+                            stdout=subprocess.PIPE
+                            ).communicate()[0].rstrip().decode('utf-8')
+REPO_DIR = Path(REPO_DIR)
```
greglucas, line 33: Do we need to keep supporting `CARTOPY_GIT_DIR`?
```suggestion
REPO_DIR = Path(os.getenv("CARTOPY_GIT_DIR", REPO_DIR))
```

greglucas (approved):
Looks like this is being hit in CI now.

QuLogic:
This might be problematic for downstream packagers if they put their stuff in git and Cartopy is unpacked in that directory. But downstream packagers probably don't want to check stylistic things like this, so tests like this should probably get a pytest marker so they can be skipped.

rcomer:
Over at https://github.com/SciTools/cf-units/pull/364#discussion_r1501764951, @pelson suggested these checks would be better via pre-commit hooks.  I found this [insert-license](https://github.com/Lucas-C/pre-commit-hooks?tab=readme-ov-file#insert-license) is a [supported hook](https://pre-commit.com/hooks.html).

PR #1507, opened 2020-04-04
Adding deprecations for old gridliner labels

greglucas:
Fixes #1383
## Rationale

The previous gridliners were called `xbottom_labels` and have been renamed `bottom_labels`, this is adding in a deprecation to allow the old versions to still be used, but warning on it.

Also updated the gridliner example to use the new attributes.

PR #2280, opened 2023-10-28
MNT: remove deprecated Gridliner properties

rcomer:
## Rationale

These properties were first deprecated in v0.18 (#1507), which [came out three and a half years ago](https://github.com/SciTools/cartopy/releases/tag/v0.18.0).


rcomer:
It seems the wmts failures are no longer failing, which is causing the doc build to fall over with "I expected these to fail but they didn't".

greglucas (approved):

PR #2282, opened 2023-10-29
TST: Skip scipy tests if scipy isn't installed

### LINKED ISSUE #2275: Tests fail without scipy
### Description

Doing a clean install of Cartopy with `pip install .[test]` ends up with 5 failures and 3 errors due to missing scipy imports. We could either make these optional tests, or add scipy to the testing optional install. Further, would we want to add the other optional things to the test installation? `test = ["cartopy[ows,plotting,speedups]", ...` i.e. should the testing option be as complete as possible, or should those truly only be if people want to test those branches?

```python
FAILED lib/cartopy/tests/mpl/test_mpl_integration.py::test_quiver_regrid - ImportError: Regridding vectors requires scipy.
FAILED lib/cartopy/tests/mpl/test_mpl_integration.py::test_quiver_regrid_with_extent - ImportError: Regridding vectors requires scipy.
FAILED lib/cartopy/tests/mpl/test_mpl_integration.py::test_barbs_regrid - ImportError: Regridding vectors requires scipy.
FAILED lib/cartopy/tests/mpl/test_mpl_integration.py::test_barbs_regrid_with_extent - ImportError: Regridding vectors requires scipy.
FAILED lib/cartopy/tests/mpl/test_mpl_integration.py::test_streamplot - ImportError: Regridding vectors requires scipy.
ERROR lib/cartopy/tests/test_img_transform.py
ERROR lib/cartopy/tests/test_vector_transform.py
ERROR lib/cartopy/tests/mpl/test_contour.py
```

greglucas:
Scipy isn't a required dependency, so skip the tests that require scipy if scipy isn't found.

`pytest.importorskip` in the module-level imports caused flake8 warnings, so I decided to just try/except and skip instead which passes flake8 without requiring a bunch of extra ignores.

closes #2275

### DIFF
--- a/lib/cartopy/tests/mpl/test_contour.py
+++ b/lib/cartopy/tests/mpl/test_contour.py
@@ -8,10 +8,9 @@ import matplotlib.pyplot as plt
 import numpy as np
 from numpy.testing import assert_array_almost_equal
 import pytest
-from scipy.interpolate import NearestNDInterpolator
-from scipy.signal import convolve2d
 
 import cartopy.crs as ccrs
+from cartopy.tests.conftest import requires_scipy
 
 
 def test_contour_plot_bounds():
@@ -72,8 +71,12 @@ def test_plot_after_contour_doesnt_shrink(func):
     assert_array_almost_equal(ax.get_extent(), expected)
 
 
+@requires_scipy
 def test_contour_linear_ring():
     """Test contourf with a section that only has 3 points."""
+    from scipy.interpolate import NearestNDInterpolator
+    from scipy.signal import convolve2d
+
     ax = plt.axes([0.01, 0.05, 0.898, 0.85], projection=ccrs.Mercator(),
                   aspect='equal')
     ax.set_extent([-99.6, -89.0, 39.8, 45.5])
--- a/lib/cartopy/tests/test_img_transform.py
+++ b/lib/cartopy/tests/test_img_transform.py
@@ -7,8 +7,16 @@
 import numpy as np
 from numpy.testing import assert_array_equal
 import pytest
-import scipy.spatial
 
+from cartopy.tests.conftest import (
+    _HAS_PYKDTREE_OR_SCIPY,
+    requires_pykdtree,
+    requires_scipy,
+)
+
+
+if not _HAS_PYKDTREE_OR_SCIPY:
+    pytest.skip("pykdtree or scipy are required", allow_module_level=True)
 import cartopy.crs as ccrs
 import cartopy.img_transform as img_trans
 
@@ -113,7 +121,8 @@ def test_gridding_data_outside_projection():
 
 @pytest.mark.parametrize("target_prj",
                          (ccrs.Mollweide(), ccrs.Orthographic()))
-@pytest.mark.parametrize("use_scipy", (True, False))
+@pytest.mark.parametrize("use_scipy", (pytest.param(True, marks=requires_scipy),
+                                       pytest.param(False, marks=requires_pykdtree)))
 def test_regridding_with_invalid_extent(target_prj, use_scipy, monkeypatch):
     # tests that when a valid extent results in invalid points in the
     # transformed coordinates, the regridding does not error.
@@ -128,6 +137,7 @@ def test_regridding_with_invalid_extent(target_prj, use_scipy, monkeypatch):
 
     if use_scipy:
         monkeypatch.setattr(img_trans, "_is_pykdtree", False)
+        import scipy.spatial
         monkeypatch.setattr(img_trans, "_kdtreeClass", scipy.spatial.cKDTree)
     _ = img_trans.regrid(data, lons, lats, data_trans, target_prj,
                          target_x, target_y)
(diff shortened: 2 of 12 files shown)


lib/cartopy/tests/test_img_transform.py
```diff
@@ -7,7 +7,13 @@
 import numpy as np
 from numpy.testing import assert_array_equal
 import pytest
-import scipy.spatial
+
+
+try:
+    import scipy.spatial
+except ImportError:
+    pytest.skip("scipy is required for image transforms", allow_module_level=True)
```
rcomer, line 15: In here, everything except `test_regridding_with_invalid_extent` will pass providing you have at least one of `scipy` or `pykdtree` installed.  Is it worth modifying this to try both imports?  Or is it too unlikely that a developer would have `pykdtree` but not `scipy`?

I am also unclear how to add appropriate skips to the parametrization of `test_regridding_with_invalid_extent` if we did that...

lib/cartopy/tests/mpl/test_contour.py
```diff
@@ -8,8 +8,13 @@
...
 import pytest
-from scipy.interpolate import NearestNDInterpolator
-from scipy.signal import convolve2d
+
+
+try:
+    from scipy.interpolate import NearestNDInterpolator
+    from scipy.signal import convolve2d
+except ImportError:
+    pytest.skip("scipy is required for contouring", allow_module_level=True)
```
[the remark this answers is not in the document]
greglucas: Good point. Moved that down into the test.

lib/cartopy/tests/test_img_transform.py
greglucas: Sure, it is simple enough to do the nested try/except and put the skip only at the nested version. I also added a skip to the parameterization if scipy can't be imported there.

rcomer (approved):
If I run without `scipy` or `pykdtree`, I get failures in `test_examples.py` and `test_imges.py`.  That could arguably be out of scope of this PR though.

greglucas:
> If I run without scipy or pykdtree, I get failures in test_examples.py and test_imges.py. That could arguably be out of scope of this PR though.

Good catch, I didn't realize I had pykdtree hanging around still. In fact, I get even more `ERRORS` due to image dependency imports in other modules like the `ows` tests. This is adding a lot of conditional logic for the skips in many places at this point considering no one else has complained about this before! I'm not sure how valuable this is...

dopplershift:
I think it's definitely helpful given that neither of these are considered required, so IMO we should be able to run the test suite successfully without them.

dopplershift:
I guess the alternative is to add one/both to the "test" optional dependencies.

greglucas:
Alright, I refactored it a bit and tested this with the various combinations of pykdtree/scipy available/not available so that there are no errors.

There are a few problems that were slightly annoying and made this maybe a little more verbose than ideal:
1. We import our img_transform in other modules, for example to get `_OWS_LIB_AVAILABLE` we go through that entire module that has an import of img_transform, which requires us to skip above that import statement. This also means we can't set `pytestmark = requires_pykdtree_scipy` at the top of the module because that occurs after the module imports.
2. We have optional dependencies on two imports scipy or pykdtree, so `pytest.importorskip` I don't think will work in that situation. So I added a `_HAS_PYKDTREE_SCIPY` variable that we then use to just apply a `pytest.skip()` early on in the test module imports.

QuLogic:
Seems to still be conflicts now?

lib/cartopy/tests/test_img_transform.py
```diff
@@ -7,7 +7,17 @@
...
+from cartopy.tests.conftest import (
+    _HAS_PYKDTREE,
+    _HAS_SCIPY,
+    requires_pykdtree,
+    requires_scipy,
+)
+
+
+if not (_HAS_PYKDTREE or _HAS_SCIPY):
+    pytest.skip("pykdtree or scipy are required", allow_module_level=True)
```
QuLogic, line 20:
```suggestion
from cartopy.tests.conftest import (
    _HAS_PYKDTREE_OR_SCIPY,
    requires_pykdtree,
    requires_scipy,
)


if not _HAS_PYKDTREE_OR_SCIPY:
    pytest.skip("pykdtree or scipy are required", allow_module_level=True)
```

lib/cartopy/tests/test_vector_transform.py
```diff
@@ -6,6 +6,14 @@
...
+
+
+try:
+    import scipy  # noqa: F401
+except ImportError:
+    pytest.skip("scipy is required for vector transforms", allow_module_level=True)
+
 
 import cartopy.crs as ccrs
 import cartopy.vector_transform as vec_trans
```
QuLogic, line 19:
```suggestion
from cartopy.tests.conftest import requires_scipy

import cartopy.crs as ccrs
import cartopy.vector_transform as vec_trans


pytestmark = requires_scipy
```

greglucas: This one unfortunately doesn't work because `import cartopy.vector_transform` raises an import error before the pytestmark takes effect.

rcomer (approved):

rcomer (approved):
Sorry for previously messed up review - I started adding comments a couple of weeks ago but then PRed that change instead.  Then forgot I still had the comment here when I approved it.  Can’t seem to delete the comment now either.

Anyway, I think this is good to go.

QuLogic (approved):

QuLogic:
Needs a rebase.

Issue #2137, opened 2023-02-21
Mollweide projection - completely corrupted contourf when there are gridpoints at the pole


jacopok:
### Description
This is a similar issue to #1333 or #2016, but the effect is much more dramatic, not only limited to the poles.
It seems to happen when there are gridpoints exactly _at_ the poles, not when they are further north than 89.14,
so #1334 might not fix it.
Therefore, I'm thinking that it might be a different problem.

Basically, plotting the exact same data, this is what I'm getting:

![projections_compared](https://user-images.githubusercontent.com/11020774/220460068-71b3fd6f-b305-4f95-8888-34bff1a0d46f.png)

__Setting the latitude grid to go from -89.9 to 89.9 degrees fixes the problem, but it still seems like a bug.__

#### Code to reproduce

```python
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

def function(theta, phi):
    return 0.5 * (1 + np.cos(theta) ** 2) * np.cos(2 * phi)

nlon = 240
nlat = 120
lon_degrees = np.linspace(0, 360.0, nlon)
lat_degrees = np.linspace(-90., 90., nlat)

lon = np.deg2rad(lon_degrees)
lat = np.deg2rad(lat_degrees)

lon2d, lat2d = np.meshgrid(lon, lat)

data = function(np.pi/2 - lat2d, lon2d)

fig = plt.figure(figsize=(10, 5))
ax1 = fig.add_subplot(2, 2, 1)

mappable = ax1.contourf(
    np.rad2deg(lon2d),
    np.rad2deg(lat2d),
    data,
    cmap=plt.get_cmap("seismic"),
    levels=np.linspace(-1, 1, num=50),
)
plt.colorbar(mappable, ax=ax1)
ax1.set_title('Bare contourf')

proj = ccrs.Mollweide()
ax2 = fig.add_subplot(2, 2, 2, projection=proj)
mappable = ax2.contourf(
    np.rad2deg(lon2d),
    np.rad2deg(lat2d),
    data,
    cmap=plt.get_cmap("seismic"),
    levels=np.linspace(-1, 1, num=50),
    transform=ccrs.PlateCarree(),
)
ax2.coastlines()
ax2.set_global()
plt.colorbar(mappable, ax=ax2)
ax2.set_title('Mollweide')

proj = ccrs.PlateCarree()
ax3 = fig.add_subplot(2, 2, 3, projection=proj)
mappable = ax3.contourf(
[...]

PR #2298, opened 2023-12-05
FIX: Remove force_path_ccw of contours

### LINKED ISSUE #2288: cartopy 0.22 + matplotlib 3.8 + contourf bug?
### Description
After upgrading to Matplotlib3.8, it is suspected that there is a bug in cartopy 0.22 when drawing contourfs. Test code and draw results in [test code and draw result](https://github.com/david-shu/cartopy-contourf-data)

The problem seems to be here. when not force_path_ccw, got the right contourf.
```
class GeoAxes(matplotlib.axes.Axes):
    def contourf(self, *args, **kwargs):
            ##....................
            if isinstance(t, mtransforms.Transform):
                for sub_trans, _ in t._iter_break_from_left_to_right():
                    if isinstance(sub_trans, InterProjectionTransform):
                        if not hasattr(sub_trans, 'force_path_ccw'):
                            sub_trans.force_path_ccw = True
           ##.......................
    
```

The same code is drawn correctly on Matplotlib3.7.

the same data. but got different contourf. see

contourf result

#### Code to reproduce

```
import PyQt6.QtCore
import cartopy.crs as ccrs
import matplotlib as mpl
import numpy as np
from urllib.request import urlopen


proj0 = ccrs.PlateCarree( central_longitude=0 )

x = np.load( urlopen( 'https://github.com/david-shu/cartopy-contourf-data/raw/main/x.dat' ) )
y = np.load( urlopen( 'https://github.com/david-shu/cartopy-contourf-data/raw/main/y.dat' ) )
v = np.load( urlopen( 'https://github.com/david-shu/cartopy-contourf-data/raw/main/v.dat' ) )

shape = ( 61,240 )

x = x.reshape( shape )
y = y.reshape( shape )
v = v.reshape( shape )

levels = [-4.0, -2.0, -1.0, -0.5, 0.0, 0.5, 1.0, 2.0, 4.0]

### contourf wrong ###### 
ax = plt.subplot( 2, 2, 1, projection=proj0 )
cs = ax.contourf( x, y, v,
                  levels=levels,
                  extend='both',
                  transform_first=False
                  )
plt.colorbar( cs )
ax.set_title('PlateCarree + extend both' )
###########################


ax = plt.subplot( 2, 2, 2, projection=proj0 )
cs = ax.contourf( x, y, v,
                  levels=levels,
                  extend='min',
                  transform_first=False
                  )
plt.colorbar( cs )
ax.set_title('PlateCarree + extend min' )


ax = plt.subplot( 2, 2,3, projection=proj0 )
cs = ax.contourf( x, y, v,
                  levels=levels,
                  extend='both',
                  transform_first=True
                  )
plt.colorbar( cs )
ax.set_title('PlateCarree + transform_first' )

ax = plt.subplot( 2, 2, 4 )
cs = ax.contourf( x, y, v,
                  levels=levels,
                  extend='both',
                  )
plt.colorbar( cs )
ax.set_title('Axes' )

plt.show()
```

<details>

### Operating system
win10 / centos

### Cartopy version
cartopy 0.22.0
cartopy 0.21.1

### matplotlib version
matplotlib 3.8.1
</details>

greglucas:
This was added to fix previously unoriented paths from Matplotlib's contouring code. This has seemingly been fixed upstream somewhere in the past and there is no test/reproducer in Cartopy, so we can remove this for now because the polygons are oriented as Shapely expects from Matplotlib.

closes #2288

If anyone has ideas for good test cases, it would be nice to add some, but I haven't really given much thought to what produces the bad geometries. An example of what I think would be nice to test below with some minimal values.
1. Contour with a hole in middle of projection
2. Contour with a hole wrapped around the edge of the globe (set `central_longitude=180` in the `proj_map` call)
3. Others...?

```python
import numpy as np
from cartopy import crs as ccrs


proj_data = ccrs.PlateCarree()
proj_map = ccrs.Robinson()

xx, yy = np.meshgrid([-20, 0, 20], [-20, 0, 20])
levels = (10, 50)
zz = np.abs(xx) + np.abs(yy)

ax = plt.figure().add_subplot(projection=proj_map)

cset = ax.contourf(xx, yy, zz, transform=proj_data, levels=levels)
ax.coastlines()

plt.show()
```

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -171,12 +171,7 @@ class InterProjectionTransform(mtransforms.Transform):
             return mpath.Path(self.transform(src_path.vertices))
 
         transformed_geoms = []
-        # Check whether this transform has the "force_path_ccw" attribute set.
-        # This is a cartopy extension to the Transform API to allow finer
-        # control of Path orientation handling (Path ordering is not important
-        # in matplotlib, but is in Cartopy).
-        geoms = cpatch.path_to_geos(src_path,
-                                    getattr(self, 'force_path_ccw', False))
+        geoms = cpatch.path_to_geos(src_path)
 
         for geom in geoms:
             proj_geom = self.target_projection.project_geometry(
@@ -1639,16 +1634,6 @@ class GeoAxes(matplotlib.axes.Axes):
             arguments X and Y must be provided and be 2-dimensional.
             The default is False, to compute the contours in data-space.
         """
-        t = kwargs.get('transform')
-        if isinstance(t, ccrs.Projection):
-            kwargs['transform'] = t = t._as_mpl_transform(self)
-        # Set flag to indicate correcting orientation of paths if not ccw
-        if isinstance(t, mtransforms.Transform):
-            for sub_trans, _ in t._iter_break_from_left_to_right():
-                if isinstance(sub_trans, InterProjectionTransform):
-                    if not hasattr(sub_trans, 'force_path_ccw'):
-                        sub_trans.force_path_ccw = True
-
         result = super().contourf(*args, **kwargs)
 
         # We need to compute the dataLim correctly for contours.


rcomer (approved):
I don't know how you begin to write a test for this given that we do not have the data from the original issue.  I think it makes sense to remove these line as we have no evidence they are still needed.

However, I think there is more going on with #2288: noting that the latitudes in the example run north-south (which I assume is why the contours end up clockwise), if we [flip the latitude direction](https://github.com/david-shu/cartopy-contourf-data/compare/main...rcomer:cartopy-2288:flip-latitudes) we can still reproduce the problem plot regardless of whether we use `main` or this branch.

greglucas:
I think that is a separate issue/bug then having to do with the wrapping of polygons probably... If we flip the indices in that example to be sorted -180 -> 180 (like we do in the `transform_first` path) then there is no yellow polygon filling everything.

```python
ind = np.argsort(x, axis=1)
x = np.take_along_axis(x, ind, axis=1)
y = np.take_along_axis(y, ind, axis=1)
v = np.take_along_axis(v, ind, axis=1)
```

greglucas:
Here is a good issue with links to others all with some geometry getting wrapped around boundaries incorrectly: https://github.com/SciTools/cartopy/issues/2137
which might have the same effect as what is happening here.

rcomer:
Let's get this in anyway...

Issue #2314, opened 2024-01-14
WMTS cache capability in WMTSRasterSource


dnowacki-usgs:
### Description

I asked a relevant [question on SO](https://stackoverflow.com/questions/77283841/cached-wmts-requests) a few months back but raising as an issue here as the current documentation is unclear and/or misleading.

When accessing a WMTS source, [`ax.add_wmts()`](https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.mpl.geoaxes.GeoAxes.html#cartopy.mpl.geoaxes.GeoAxes.add_wmts) calls [`WMTSRasterSource()`](https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.io.ogc_clients.WMTSRasterSource.html#cartopy.io.ogc_clients.WMTSRasterSource).  The docs for `WMTSRasterSource` state that it "Uses tile caching for fast repeated map retrievals" and a [glance at the code](https://scitools.org.uk/cartopy/docs/latest/_modules/cartopy/io/ogc_clients.html#WMTSRasterSource) indicates cache-related stuff going on.

In my experience, however, no caching is performed, and repeated identical calls result in network activity every time. Are the docs referring to some other kind of caching?

#### Code to reproduce

```
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

plt.figure()
ax = plt.subplot(1,1,1,projection=ccrs.Mercator())
ax.set_extent([-122.55, -122, 37.4, 37.85], crs=ccrs.PlateCarree())
ax.add_wmts('https://basemap.nationalmap.gov/arcgis/rest/services/USGSImageryOnly/MapServer/WMTS/1.0.0/WMTSCapabilities.xml', layer_name='USGSImageryOnly')
plt.show()
```

<details>
 <summary>Full environment definition</summary>

### Operating system
macOS 13.6.3

### Cartopy version

0.22.0

### conda list

```
# Name                    Version                   Build  Channel
adal                      1.2.7              pyhd8ed1ab_0    conda-forge
adlfs                     2023.10.0          pyhd8ed1ab_0    conda-forge
affine                    2.4.0              pyhd8ed1ab_0    conda-forge
aiobotocore               2.9.0              pyhd8ed1ab_0    conda-forge
aiohttp                   3.9.1           py311he705e18_0    conda-forge
aioitertools              0.11.0             pyhd8ed1ab_0    conda-forge
[...]

PR #2316, opened 2024-01-18
Implement local caching for WMTS requests

dnowacki-usgs:
This is an in-progress PR to implement local caching of WMTS image tiles. It is based heavily on the caching approach and code implemented in #1533. No tests added yet; this is really a working proof of concept.

## Rationale

The rationale is described in #2314.

Briefly, this PR implements local caching of image tiles so repeated WMTSRasterSource calls do not incur additional network traffic.

## Implications

- Reduced network usage for repeated calls.
- Creation of files on disk in a temporary directory.
- Ability to work offline and use WMTSRasterSource


lib/cartopy/io/ogc_clients.py
```diff
@@ -358,7 +361,7 @@ class WMTSRasterSource(RasterSource):
 
     """
 
-    def __init__(self, wmts, layer_name, gettile_extra_kwargs=None):
+    def __init__(self, wmts, layer_name, gettile_extra_kwargs=None, cache=False):
```
greglucas, line 364: Add the description to the Parameters section of the docstring

lib/cartopy/io/ogc_clients.py
```diff
@@ -398,6 +401,18 @@ def __init__(self, wmts, layer_name, gettile_extra_kwargs=None):
...
 
+        # Enable a cache mechanism when cache is equal to True or to a path.
+        self._default_cache = False
+        if cache is True:
+            self._default_cache = True
+            self.cache_path = cartopy.config["cache_dir"]
+        elif cache is False:
+            self.cache_path = None
+        else:
+            self.cache_path = cache
```
greglucas, line 412: We should stick with Pathlib objects here if we can which will make things easier later on.
```suggestion
            self.cache_path = Path(cache)
```

lib/cartopy/io/ogc_clients.py
```diff
@@ -511,6 +526,26 @@ def fetch_raster(self, projection, extent, target_resolution):
 
         return located_images
 
+    @property
+    def _cache_dir(self):
+        """Return the name of the cache directory"""
+        return os.path.join(
+            self.cache_path,
+            self.__class__.__name__
+        )
```
greglucas, line 535: If you take the pathlib suggestion above this can be written as
```suggestion
        return self.cache_path / self.__class__.__name__
```

Below this you can do cachedir.exists() and similar as well.

lib/cartopy/io/ogc_clients.py
```diff
@@ -643,21 +678,41 @@ def _wmts_images(self, wmts, layer, matrix_set_name, extent,
...
-                    except owslib.util.ServiceException as exception:
-                        if ('TileOutOfRange' in exception.message and
-                                ignore_out_of_range):
-                            continue
-                        raise exception
-                    img = Image.open(io.BytesIO(tile.read()))
-                    image_cache[img_key] = img
+                    # Try it from disk cache
+                    if self.cache_path is not None:
+                        filename = "_".join([str(i) for i in img_key]) + ".npy"
```
greglucas, line 685: Might be more explicit to fill it in as an f-string since there are only two values?
```suggestion
                        filename = f"{img_key[0]}_{img_key[1]}.npy"
```

greglucas (commented):
Seems reasonable. There is a lot of duplicated code here between this and the other cache PR, so I am curious if there would be a good way to consolidate some of it into helper functions or class to be shared somehow. That can be a follow-up PR as well though if you don't want to expand the scope here.

dnowacki-usgs:
Thanks for your feedback. Working on implementing these suggestions!

greglucas:
@dnowacki-usgs , sorry for losing track of this one. Were you still making updates to it or is it ready for review now? It looks like at a minimum it would be good to add a test for the caching. Could you rebase on the latest main branch updates to get all the checks passing again?

dnowacki-usgs:
Hi Greg--I did make some progress as I recall but it's been a few months. I will look back into how far I got and update the PR.

dnowacki-usgs:
After merging the updated main branch and adding a missing import the PR is running again, but the cache is wonky--only some of the tiles are being saved and/or loaded with gaps. This wasn't the case when I was working on this months ago, so this will take a deeper dive.

greglucas (commented):
:+1: Thanks for looking into all of this! Overall looks good to me, just some minor nits. Could you also add a similar test to the GoogleImageTiles one for this new addition too?

https://github.com/SciTools/cartopy/blob/0d4e39f54b7da5eb5205b49ddd53b747ae02725f/lib/cartopy/tests/test_img_tiles.py#L333

raphaeldussin:
@dnowacki-usgs thank you very much for this PR, it really solved slowness and dropout issues for me. For anyone interested, this is a code snippet to create N figures using the cache for bluemarble:

```python
import cartopy
import cartopy.crs as ccrs
import owslib.wmts

# caching the WMTS
cache_dir = "~/.cartopy"
cartopy.config.update({"cache_dir": cache_dir})
url = 'https://map1c.vis.earthdata.nasa.gov/wmts-geo/wmts.cgi'
layer = 'BlueMarble_NextGeneration'
wmts = owslib.wmts.WebMapTileService(url)

# loop over all the first 10 frames
for kt in range(0, 10):
    subplot_kws={'projection': ccrs.NearsidePerspective()}
    ax = plt.axes(**subplot_kws)
    ax.add_wmts(wmts, layer, cache=cache_dir)
    plt.close("all")

I really hope this PR will be merged soon as it is solving quite a big problem for WMTS users.

**Comment by dnowacki-usgs** [comment]
@raphaeldussin glad to hear that! Currently I'm trying to get tests integrated, which is taking time since I'm not super familiar with the test infrastructure as implemented in cartopy. I'm trying to base it off the GoogleTiles test. Knowing someone is using it is good inspiration to get the tests working at the PR updated.

File: lib/cartopy/tests/test_img_tiles.py [hunk 5]
```diff
@@ -328,6 +331,81 @@ def test_azuremaps_get_image():
     assert extent1 == extent2
 
 
+@pytest.mark.network
+@pytest.mark.parametrize('cache_dir', ["tmpdir", True, False])
```
greglucas, line 335: I think we need to add a skipif OWSLIB to handle it not installed as well since it is an optional dependency.
```suggestion
@pytest.mark.parametrize('cache_dir', ["tmpdir", True, False])
@pytest.mark.skipif(not ogc._OWSLIB_AVAILABLE, reason='OWSLib is unavailable.')
```

greglucas (approved):
I think this looks great (one minor note on adding a skipif decorator), thank you for adding the test!

I agree there is a lot of duplicated code in the test, but to me that could be a follow-up PR, lets not bog this one down. It might be more complicated than it is worth to refactor it.

greglucas:
Thanks for your patience with the reviews, @dnowacki-usgs!

PR #2317, opened 2024-01-24
Allow GSHHS levels 5 and 6 for Antarctica.

kdungs:
## Rationale

The [documentation for `GSHHSFeature`](https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.feature.GSHHSFeature.html) already states that levels 1 through 6 are allowed but the code so far only allowed 1 through 4.

As per https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhg/readme.txt levels 5 and 6 were introduced in GSHHS version 2.3.0 in February 2014 to distinguish the Antarctic ice-front (L5) and grounding lines (L6).

Importantly, this also means that L1 does not contain any coastlines for Antarctica, so allowing only L1-L4 is insufficient.

See also https://www.soest.hawaii.edu/pwessel/gshhg/.

## Implications

 * Antarctica coastlines are supported.
 * Users have to manually choose L5 or L6 or both.

Side note: there are no existing tests I could piggyback on but "it works on my machine"... Let me know if you have a more structured way to make sure this works for everyone.

### DIFF
--- a/lib/cartopy/feature/__init__.py
+++ b/lib/cartopy/feature/__init__.py
@@ -365,7 +365,7 @@ class GSHHSFeature(Feature):
         if levels is None:
             levels = [1]
         self._levels = set(levels)
-        unknown_levels = self._levels.difference([1, 2, 3, 4])
+        unknown_levels = self._levels.difference([1, 2, 3, 4, 5, 6])
         if unknown_levels:
             raise ValueError(f"Unknown GSHHS levels {unknown_levels!r}.")
 


greglucas:
Is that unknown levels error even valuable? Maybe we should just remove that check altogether. What happens if they add a 7, 8 in the future, we would then raise again which doesn't seem ideal.

kdungs:
Thanks for taking a look, Greg!

That's a good point, actually. Two things that might be worth considering - also in the interest of not making this a larger refactor (I also thought about adding enum values for the different levels; but again, that would have made this a larger PR):

 1. GSHHS as far as I know isn't actively maintained anymore. The last update is from 2017 (see https://www.ngdc.noaa.gov/mgg/shorelines/gshhs.html). This is only my limited understanding, though so it might be worth checking with the SOEST folks.
 2. The current check has a nice side-effect in that it also functions as a de-facto type check. Since the expected values are integers, callers can't e.g. provide strings. From what I can tell from where it's used in [shapereader.py](https://github.com/SciTools/cartopy/blob/main/lib/cartopy/io/shapereader.py#L455) it should be safe. In general, allowing arbitrary user-controlled strings that then go into URLs or filesystem paths can easily lead to security issues 🙈.

So for me personally, the most important aspect is making a minimal change 😄. Then, I think (1) is a good enough argument for keeping it as is for the time being.

What do you think?

---

Oh and btw. thank you and your team so much for Cartopy. All I wanted to do was quickly slap together a visualization of my friends' and my Antarctica sailing trip and now I'm deep into the rabbit hole of cartography, bathymetry, open data, shapefiles, etc. Even though it eats up considerable amounts of my free time, it's great fun and Cartopy is really nice to work with! 💕

rcomer (approved):
The docstring change came in at #1823, which also involved some changes in `shapereader.py`, including a hard-coded version number.  So it looks like, regardless of this check, we would have to do some work if a new version with more levels ever came out.

Given that and @kdungs's points above, I will vote to take this change as is.

greglucas (approved):
I agree, this is a good update as-is. Thanks for updating this @kdungs, we hope to hear from you again!

kdungs:
Awesome! Thanks a lot, folks!

PR #2324, opened 2024-02-08
DOC: don't use deprecated collections attribute

rcomer:
## Rationale

I just spotted this while grepping for something else.  The `ContourSet.collections` attribute was [deprecated at mpl 3.8](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.8.0.html#contourset-collections).  We could now just call `set_visible` on the `ContourSet` itself, but that will break with older Matplotlib versions.  I have checked that the current suggestion works with Cartopy main and mpl 3.8.2, and also with Cartopy 0.21.1 and mpl 3.6.3.

### DIFF
--- a/examples/scalar_data/contour_labels.py
+++ b/examples/scalar_data/contour_labels.py
@@ -44,13 +44,10 @@ def main():
     # Add colourful filled contours.
     filled_c = ax.contourf(x, y, z, transform=ccrs.PlateCarree())
 
-    # And black line contours.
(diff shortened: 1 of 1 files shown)


rcomer:
I guess another option would be to just have two commented-out lines, one each for older and newer Matplotlib versions.  Is it more efficient at draw time to have the visible flag set to `False` than the color set to 'none'?

PR #2325, opened 2024-02-09
FIX: use compound paths to draw transformed geometries

rcomer:
## Rationale

When a geometry gets transformed to a different crs, it can become multiple geometries.  Currently we plot these as separate paths which could be problematic if the user passes a list of e.g. facecolors.

```python
import cartopy.crs as ccrs

square1 = sgeom.Polygon([(-10, -10), (10, -10), (10, 10), (-10, 10), (-10, 10) ])
square2 = sgeom.Polygon([(30, 0), (50, 0), (50, 20), (30, 20), (30, 0)])

fig = plt.figure(figsize=(4, 5))

ax1 = fig.add_subplot(211, projection=ccrs.PlateCarree())
ax2 = fig.add_subplot(212, projection=ccrs.PlateCarree(central_longitude=180))

for ax in ax1, ax2:
    ax.coastlines()
    ax.add_geometries([square1, square2], crs=ccrs.PlateCarree(),
                      facecolors=['r', 'b'])

plt.show()
```
Before:
image

After:
image

### DIFF
--- a/lib/cartopy/mpl/feature_artist.py
+++ b/lib/cartopy/mpl/feature_artist.py
@@ -15,6 +15,7 @@ import weakref
 
 import matplotlib.artist
 import matplotlib.collections
+import matplotlib.path as mpath
 import numpy as np
 
 import cartopy.mpl.patch as cpatch
@@ -179,15 +180,21 @@ class FeatureArtist(matplotlib.artist.Artist):
                 geom_key, geom)
             mapping = FeatureArtist._geom_key_to_path_cache.setdefault(
                 geom_key, {})
-            geom_paths = mapping.get(key)
-            if geom_paths is None:
+            geom_path = mapping.get(key)
+            if geom_path is None:
                 if ax.projection != feature_crs:
                     projected_geom = ax.projection.project_geometry(
                         geom, feature_crs)
                 else:
                     projected_geom = geom
+
                 geom_paths = cpatch.geos_to_path(projected_geom)
-                mapping[key] = geom_paths
+                if not geom_paths:
+                    continue
+                # The transform may have split the geometry into two paths, we only want
+                # one compound path.
+                geom_path = mpath.Path.make_compound_path(*geom_paths)
+                mapping[key] = geom_path
 
             if not self._styler:
                 style = prepared_kwargs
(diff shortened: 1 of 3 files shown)


lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -585,7 +585,7 @@ def test_pcolormesh_set_clim_with_mask():
 
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(filename='pcolormesh_limited_area_wrap.png',
-                               tolerance=1.82)
+                               tolerance=1.83)
```
rcomer, line 588: Diff looks like
temp

lib/cartopy/mpl/feature_artist.py
```diff
@@ -179,15 +180,22 @@ def draw(self, renderer, *args, **kwargs):
...
                     projected_geom = geom
+
                 geom_paths = cpatch.geos_to_path(projected_geom)
-                mapping[key] = geom_paths
+                if not geom_paths:
+                    continue
+                elif len(geom_paths) > 1:
+                    geom_path = mpath.Path.make_compound_path(*geom_paths)
+                else:
+                    geom_path, = geom_paths
```
greglucas, line 197: do you need the special case, or can you input a single path into it as well?
```suggestion
                # The transform may have split the geometry into two paths, we only want one compound path
                geom_path = mpath.Path.make_compound_path(*geom_paths)
```

Should this logic be moved up into the `geos_to_path()` function though? That currently returns a list of paths, but it seems like if you're passing in one geometry that should be a compound path like you have here.

rcomer: It tried pushing this to `geos_to_path`, but got failures because sometimes `projected_geom` is a tuple, having been transformed from a `LinearRing`
https://github.com/SciTools/cartopy/blob/25a2154604517cc6d46bbb82b83f1c46cb0df381/lib/cartopy/crs.py#L907

greglucas: sigh :( Thanks for looking into that! I think the geometry transforms needs a good rework with all the great fixes you're coming up with! Looking through the linearring code, it looks like it is primarily used within `_project_polygon` to project the boundaries and holes and then attaches the dangling linestring to the boundary. I wonder if we could just return a single MultiLineString in that case and consider the start/end ourselves for whether it is closed (LinearRing) or not.

This is good as-is, I'm just writing down thoughts for future ideas at this point.

greglucas (approved):

PR #2333, opened 2024-02-25
MNT: remove FeatureArtist.draw args and kwargs

rcomer:
## Rationale

I was going to propose deprecating these so I added some deprecation warnings.  However, when I tried to trigger those warnings I found that you just can't pass extra parameters here.  The method has Matplotlib's `allow_rasterization` decorator which has not passed args and kwargs since https://github.com/matplotlib/matplotlib/pull/20331.  Since Matplotlib v3.5 is now our [minimum supported version](https://github.com/SciTools/cartopy/blob/7570278c8f31cd1aedc4878882016e0a825a3675/pyproject.toml#L44), there is no way these can be used any more.  So I propose to just remove them.  I had hoped this would lead to simpler code but, unless I'm missing a trick, it doesn't make much difference.

## Implications

This removes our internal use of the [style module](https://github.com/SciTools/cartopy/blob/main/lib/cartopy/mpl/style.py) so we could consider deprecating that if we wanted.

### DIFF
--- a/lib/cartopy/mpl/feature_artist.py
+++ b/lib/cartopy/mpl/feature_artist.py
@@ -9,7 +9,6 @@ This module defines the :class:`FeatureArtist` class, for drawing
 
 """
 
-from collections import OrderedDict
 import warnings
 import weakref
 
@@ -21,7 +20,6 @@ import numpy as np
 import cartopy.feature as cfeature
 from cartopy.mpl import _MPL_38
 import cartopy.mpl.patch as cpatch
-from .style import merge as style_merge
 
 
 class _GeomKey:
@@ -154,7 +152,7 @@ class FeatureArtist(matplotlib.collections.Collection):
             self._paths = paths
 
     @matplotlib.artist.allow_rasterization
-    def draw(self, renderer, *args, **kwargs):
+    def draw(self, renderer):
         """
         Draw the geometries of the feature that intersect with the extent of
         the :class:`cartopy.mpl.geoaxes.GeoAxes` instance to which this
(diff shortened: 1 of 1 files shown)


lib/cartopy/mpl/feature_artist.py
```diff
@@ -186,12 +185,10 @@ def draw(self, renderer, *args, **kwargs):
...
             geoms = self._feature.intersecting_geometries(extent)
 
-        # Freeze the kwargs so that we can use them as a dict key. We will
-        # need to unfreeze this with dict(frozen) before passing to mpl.
-        prepared_kwargs = _freeze(kwargs)
-
         # Project (if necessary) and convert geometries to matplotlib paths.
-        stylised_paths = OrderedDict()
+        stylised_paths = OrderedDict()  # used with styler
+        unstyled_paths = []             # used if there is no styler
```
greglucas, line 190: Do we need OrderedDict anymore? The versions of Python we support are now insertion order guaranteed.

greglucas (approved):
I think this makes sense, nice job tracking that down.

One thought for simplification is whether we need to track both stylized and un-stylized paths? Could we instead add an empty style dictionary if there is no style applied? Then we wouldn't need to keep track of both paths.

If we aren't using the style module ourselves it seems like it would be worth deprecating that too.

rcomer:
> Could we instead add an empty style dictionary if there is no style applied?

Done.  I had already been in two minds about whether to do that as it makes the no-styler case itself more complicated but as you say it reduces paths through the code.

PR #2337, opened 2024-03-04
MNT: deprecate the style module

rcomer:
## Rationale

#2323 and #2333 removed our internal use of this module.

This module does 2 things
* Merges dictionaries of styles taking account of a hard-coded subset of Matplotlib aliases (specifically not plural aliases like _facecolors_).  If this was intended for general use I think it should handle all aliases, maybe making use of [ArtistInspector.get_aliases](https://matplotlib.org/stable/api/_as_gen/matplotlib.artist.ArtistInspector.html#matplotlib.artist.ArtistInspector.get_aliases) so it could be artist specific.  However a utility function for that probably belongs in Matplotlib (if anywhere) rather than here.
* Handles the Cartopy-specific `facecolor="never"`, but there is some discussion of removing that completely at https://github.com/SciTools/cartopy/pull/2323#discussion_r1493368803.

## Implications

One less module to maintain.

### DIFF
--- a/lib/cartopy/mpl/style.py
+++ b/lib/cartopy/mpl/style.py
@@ -11,6 +11,10 @@ Handles matplotlib styling in a single consistent place.
 import warnings
 
 
+warnings.warn('The style module is deprecated and will be removed in a future release.',
+              DeprecationWarning, stacklevel=2)
+
+
 # Define the matplotlib style aliases that cartopy can expand.
 # Note: This should not contain the plural aliases
 # (e.g. linewidths -> linewidth).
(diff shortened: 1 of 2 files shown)


greglucas (commented):
This module seems like it should have been a private helper and hopefully won't be used by others downstream... Although it does look like it came in fairly recently:
https://github.com/SciTools/cartopy/pull/1190

With your recent changes making `FeatureArtist` a `Collection` https://github.com/SciTools/cartopy/pull/2323 I think it is OK to get rid of this now, but it would be good to hear if @pelson or @QuLogic had any other things in mind for this module.

QuLogic:
If we can avoid having to write our own stuff, that's probably for the best. I don't have any particular plans in mind for this module.

greglucas (approved):

PR #2338, opened 2024-03-05
Prepare for Numpy 2.0

greglucas:
I installed the latest pre-release of numpy 2.0 to see if we had any failures appear and these were the only ones that showed up. Pretty simple fixes to all internal/test machinery.

Here is the quick command to help others wanting to test this out by upgrading some local packages as well if wanted.
`python -m pip install numpy matplotlib contourpy --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --pre --upgrade numpy`

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -53,21 +53,11 @@ _ROTATE_LABEL_PROJS = _POLAR_PROJS + (
 )
 
 
-def _fix_lons(lons):
-    """
-    Fix the given longitudes into the range ``[-180, 180]``.
-
-    """
-    lons = np.array(lons, copy=False, ndmin=1)
-    fixed_lons = ((lons + 180) % 360) - 180
-    # Make the positive 180s positive again.
-    fixed_lons[(fixed_lons == -180) & (lons > 0)] *= -1
-    return fixed_lons
-
-
 def _lon_hemisphere(longitude):
     """Return the hemisphere (E, W or '' for 0) for the given longitude."""
-    longitude = _fix_lons(longitude)
+    # Wrap the longitude to the range -180 to 180, keeping positive 180s
+    lon_wrapped = ((longitude + 180) % 360) - 180
+    longitude = 180 if (longitude > 0 and lon_wrapped == -180) else lon_wrapped
     if longitude > 0:
         hemisphere = 'E'
     elif longitude < 0:
--- a/lib/cartopy/tests/crs/test_oblique_mercator.py
+++ b/lib/cartopy/tests/crs/test_oblique_mercator.py
@@ -187,7 +187,7 @@ def test_equality(oblique_variants):
     "reverse_coord", [False, True], ids=["xy_order", "yx_order"]
 )
 def test_nan(oblique_mercator, plate_carree, reverse_coord):
-    coord = (0.0, np.NaN)
+    coord = (0.0, np.nan)
     if reverse_coord:
         coord = tuple(reversed(coord))
     res = oblique_mercator.transform_point(*coord, src_crs=plate_carree)


lib/cartopy/mpl/gridliner.py
```diff
@@ -53,21 +53,20 @@
 )
 
 
-def _fix_lons(lons):
+def _fix_lon(lon):
```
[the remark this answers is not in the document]
greglucas: Good call. I changed up the logic a bit to accommodate that, so make sure that looks right again.

lib/cartopy/mpl/gridliner.py
```diff
@@ -53,21 +53,11 @@
...
-    fixed_lons[(fixed_lons == -180) & (lons > 0)] *= -1
-    return fixed_lons
-
-
 def _lon_hemisphere(longitude):
     """Return the hemisphere (E, W or '' for 0) for the given longitude."""
-    longitude = _fix_lons(longitude)
+    if longitude != 180:
+        # Wrap the longitude to the range -180 to 180, keeping positive 180s
+        longitude = ((longitude + 180) % 360) - 180
```
QuLogic, line 60: There's a minor difference here in that `180 + 360 *n` was formerly considered +180, whereas now it'll be -180.

greglucas: True, I was ignoring the larger multiples. I think we need an intermediate variable to handle that as before, I just pushed up a new change.

greglucas: Thinking more generally here... this is only in the W/E labels function. Do we actually want a label on the 180s, or should we just strip that off altogether like we do for 0 already? Remove the E/W ambiguity.

dopplershift (approved):
Looks good to me.

lib/cartopy/mpl/gridliner.py
QuLogic: Yea, that might be something to do instead. I feel like there are probably differing opinions on which one to use, too.

dopplershift: I agree.

greglucas: OK, there was one image failure with that update, so I pushed the change to remove the E/W label as that simplifies the logic a bit. I'm happy to remove that and go back to the previous logic if preferred too.

I looked up a few examples and it seems there are versions with and without the labels on 180, so no standard I could see at a quick glance.

Matlab does produce E/W labels on 180: https://www.mathworks.com/help/map/the-map-grid.html
Images without the E/W labels on 180: https://manoa.hawaii.edu/exploringourfluidearth/physical/world-ocean/locating-points-globe

Wikipedia seems to indicate you should label both the prime and antimeridian with either E or W: https://en.wikipedia.org/wiki/180th_meridian
> The [longitude](https://en.wikipedia.org/wiki/Longitude) at this line can be given as either [east](https://en.wikipedia.org/wiki/East) or [west](https://en.wikipedia.org/wiki/West).

greglucas: I was too undecided. I now went to the old version and kept the E/W labels as they were so we aren't changing anything anymore. This is the simple change for now to get ready for numpy 2.0 and we can re-evaluate the label naming in a follow-up PR later.

QuLogic: I think we'd eventually want to go for something "smarter", that adds the direction if there are (potential for) two 180 labels. But this can be a separate PR.

lib/cartopy/mpl/gridliner.py
```diff
@@ -53,21 +53,11 @@
...
-    fixed_lons[(fixed_lons == -180) & (lons > 0)] *= -1
-    return fixed_lons
-
-
 def _lon_hemisphere(longitude):
     """Return the hemisphere (E, W or '' for 0) for the given longitude."""
-    longitude = _fix_lons(longitude)
+    # Wrap the longitude to the range -180 to 180, keeping positive 180s
+    lon_wrapped = ((longitude + 180) % 360) - 180
+    longitude = longitude if (longitude > 0 and lon_wrapped == -180) else lon_wrapped
```
QuLogic, line 60: This would produce 180+n*360 for positives; do we not expect such high numbers? Otherwise:
```suggestion
    longitude = 180 if (longitude > 0 and lon_wrapped == -180) else lon_wrapped
```

greglucas: Good call. I think it just didn't matter because we are only verifying if longitude > 0 below, so the check would produce the same result. But this makes more sense, so taking it.

QuLogic: Oh, I forgot the specific value isn't used for anything right now, but this is probably simpler to understand at least.

PR #2340, opened 2024-03-06
DOC: add example showing FeatureArtist is a ScalarMappable

rcomer:
## Rationale

Since #2323 `FeatureArtist` is a `ScalarMappable` so we can colour geometries using a data array.  I don't think this functionality is very discoverable so here is an example.  Bikeshedding suggestions welcome!

### DIFF
--- a/examples/scalar_data/geometry_data.py
+++ b/examples/scalar_data/geometry_data.py
@@ -1,0 +1,48 @@
+"""
+Associating data with geometries
+--------------------------------
+
+This example shows how to colour geometries based on a data array.  This
+functionality is available since Cartopy 0.23.
+
+"""
+import matplotlib.colors as mcolors
+import matplotlib.pyplot as plt
+
+import cartopy.crs as ccrs
+import cartopy.io.shapereader as shpreader
+
+
+def main():
+    # Load Natural Earth's country shapefiles.
+    shpfilename = shpreader.natural_earth(resolution='110m',
+                                          category='cultural',
+                                          name='admin_0_countries')
+    reader = shpreader.Reader(shpfilename)
+    countries = reader.records()
+
+    # Get hold of the geometry and population estimate from each country's record.
+    geometries = []
+    population_estimates = []
+
+    for country in countries:
+        geometries.append(country.geometry)
+        population_estimates.append(country.attributes['POP_EST'])
+
+    # Set up a figure and an axes with the Eckert VI projection.
+    fig = plt.figure()
+    ax = fig.add_subplot(projection=ccrs.EckertVI())
+
+    # Plot the geometries coloured according to population estimate.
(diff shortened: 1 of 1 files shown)


rcomer:
[Renders like this](https://output.circle-artifacts.com/output/job/f2585295-f72a-4ae3-b4ce-39242e073685/artifacts/0/docs/build/html/gallery/scalar_data/geometry_data.html#sphx-glr-gallery-scalar-data-geometry-data-py)

examples/scalar_data/geometry_data.py
```diff
@@ -0,0 +1,46 @@
...
+        geometries.append(country.geometry)
+        population_estimates.append(country.attributes['POP_EST'])
+
+    # Set up a figure and an axes with the Eckert VI projection.
+    fig = plt.figure()
+    ax = fig.add_subplot(projection=ccrs.EckertVI())
+
+    # Plot the geometries coloured according to population estimate.
+    art = ax.add_geometries(geometries, crs=ccrs.PlateCarree(),
+                            array=population_estimates, cmap='YlGnBu', norm='log')
```
greglucas, line 37: Set a lower limit of 1 million to get a little more color variation?
```suggestion
                            array=population_estimates, cmap='YlGnBu', norm='log', vmin=1e6)
```

examples/scalar_data/geometry_data.py
```diff
@@ -0,0 +1,46 @@
+"""
+Associating data with geometries
+--------------------------------
+
+This example shows how to colour geometries based on a data array.  This
+functionality is available since Cartopy 0.23.
```
greglucas, line 6: Is there a way to do a `.. since:` or something to automatically tag this with sphinx-gallery. I like that you put this in there to let people know when it came in!

greglucas (approved):
Looks good to me, a nice succinct example to demo it. Feel free to merge yourself with or without my minor suggestions.

examples/scalar_data/geometry_data.py
rcomer: The mappable won't take _vmin_ so we have to pass it to the `LogNorm`.  I think that's OK and I had already been wondering about limiting the range, so 👍

examples/scalar_data/geometry_data.py
rcomer: I don't know if there is but I am far from an expert in sphinx and sphinx-gallery.  Over at https://github.com/matplotlib/matplotlib/issues/27292 Hannah was advocating using sphinx tags for that sort of thing.

rcomer:
And the tests are broken again :-(

rcomer:
The test failure seems to be coming from deep in owslib.  Anyway it's clearly not caused by this PR so I'm going to merge.

PR #2347, opened 2024-03-11
CI: Only run coverage on a single runner

greglucas:
This bumps the coveralls job to Python 3.12 ubuntu-latest so that only one job gets run and submitted. It has been flaky when submitting from two sources. https://github.com/SciTools/cartopy/actions/runs/8218053565/job/22477742731?pr=2322#step:10:16
The Python 3.11 ubuntu-latest has a shapely-dev=true and a shapely-dev=false version, so hopefully this can avoid that if it is truly an issue.


QuLogic (approved):

QuLogic:
I'll ignore the docs failure.

Issue #2293, opened 2023-11-29
Natural Earth: Include version in URL


kaedonkers:
### Description
There are now five versions of Natural Earth features but currently the URL to download them in cartopy does not specify a version:
https://github.com/SciTools/cartopy/blob/main/lib/cartopy/io/shapereader.py#L323

The AWS S3 bucket that now hosts Natural Earth features supports version numbers:
https://github.com/nvkelso/natural-earth-vector/issues/246#issuecomment-1134290221
but this has not been implemented in cartopy, which instead pulls the latest version available.

While this is not an issue in itself, adjusting the URL to optionally include the version of the Natural Earth feature will result in more reproducible behaviour.

PR #2351, opened 2024-03-15
ENH: Allow Natural Earth version to be specified

lgolston:
## Rationale
As requested in #2293, it would be useful for reproducibility to be able to specify what Natural Earth version to use, in addition to current default behavior of downloading the latest version.

## Implications
This enhancement is added. If a version is specified, it will be downloaded to subfolders within cultural or physical corresponding to the version requested and then used.

## Checklist
Testing: Since Natural Earth is already in several places in the test suite, I modified one of them to now use a specific version.


rcomer:
Do we also need something to expose the version option on `NaturalEarthFeature`, similar to how scale is handled?

lgolston:
> Do we also need something to expose the version option on `NaturalEarthFeature`, similar to how scale is handled?

Yes, thanks. Version should definitely be an option there too but I wasn't tracking that feature - have added it in now.

lgolston:
Related, I have linked there to https://github.com/nvkelso/natural-earth-vector/releases as a good source for version information, because it was confusing that most of the versions shown on https://www.naturalearthdata.com/ (e.g. 4.0.0, 2.0.0, etc.) are not available on the AWS service to download, yet there are many newer versioned releases that are available which are not shown on https://www.naturalearthdata.com/.

lgolston:
CI helped pick up an omission of version in one place in NaturalEarthFeature - a clean version is now pushed with that edit.

lib/cartopy/tests/mpl/test_features.py
```diff
@@ -37,7 +37,7 @@ def test_natural_earth():
 @pytest.mark.mpl_image_compare(filename='natural_earth_custom.png')
 def test_natural_earth_custom():
     ax = plt.axes(projection=ccrs.PlateCarree())
-    feature = cfeature.NaturalEarthFeature('physical', 'coastline', '50m',
+    feature = cfeature.NaturalEarthFeature('physical', 'coastline', '50m', '5.1.0',
```
kaedonkers, line 40: Did you find that version `5.1.0` successfully passed this test?

lib/cartopy/tests/test_shapereader.py
```diff
@@ -83,7 +83,8 @@ class TestRivers:
     def setup_class(self):
         RIVERS_PATH = shp.natural_earth(resolution='110m',
                                         category='physical',
-                                        name='rivers_lake_centerlines')
+                                        name='rivers_lake_centerlines',
+                                        version='5.0.0')
```
kaedonkers, line 87: Similarly, did this specific version (`5.0.0`) of rivers pass the image comparison tests?

kaedonkers (commented):
An elegant solution, thank you for implementing this!
I have not yet tested it myself, but it looks like a viable solution to me

lib/cartopy/tests/mpl/test_features.py
lgolston: Thanks! Yes, no diff between the baseline image here of Iceland and when using 5.1.0

lib/cartopy/tests/test_shapereader.py
lgolston: Yes, it was passing. Note there is no image comparison here, the test just looks at a few very specific features and there was no change.

lib/cartopy/io/shapereader.py
```diff
@@ -394,13 +400,38 @@ def default_downloader():
...
+        Return a NEShpDownloader instance, which extends the default downloader
+        with support for specific file versions.
+
+        """
+        _NE_URL_TEMPLATE = ('https://naturalearth.s3.amazonaws.com/'
+                            '{version}/{resolution}_{category}/'
+                            'ne_{resolution}_{name}.zip')
+
+        default_spec = ('shapefiles', 'natural_earth', '{category}', '{version}',
+                        'ne_{resolution}_{name}.shp')
```
greglucas, line 415: Would you be able to inject this into the current NE template? It looks like the only difference is `{version}/`, and I'm wondering if you could put that in and always have a version with something like `_version_string = "" if version is None else f"{version}/".

greglucas (commented):
I didn't test this out yet, but would you be able to add some description to the docstrings about where the versions are cached/stored? I think they are stored separately looking at the code, but just want to make sure we aren't getting a mixture of versions in the default namespace. If the latest is `5.1.2` and I explicitly ask for `5.1.2` do we have to download/store both?

greglucas:
ping @lgolston, I had a few questions on this, but it looks like it may be very close to ready to go?

lib/cartopy/io/shapereader.py
lgolston: Yes, I couldn't figure out how to inject version without making the second default downloader, but I think this solution works! Only problem is it obscures the template slightly but worth it to avoid duplicating the other code. For the other question, yes, if latest is 5.1.2 and then you specifically ask for 5.1.2 it will download it twice, since version-specified data will go in its own subfolder (within the relevant physical or cultural folder). Unfortunately, the .VERSION file is currently filtered out so it is hard to tell what version you have downloaded in the base folder.

lgolston:
> ping @lgolston, I had a few questions on this, but it looks like it may be very close to ready to go?

Sorry for the long delay. I have made an update based on your review, however there is a test error I need to look into.

lib/cartopy/feature/__init__.py
```diff
@@ -251,6 +253,8 @@ def __init__(self, category, name, scale, **kwargs):
             The dataset scale, i.e. one of '10m', '50m', or '110m',
             or Scaler object. Dataset scales correspond to 1:10,000,000,
             1:50,000,000, and 1:110,000,000 respectively.
+        version: optional
```
greglucas, line 256: Can you also update the above cases to follow numpydoc formatting by adding ` : str` to all of them.
```suggestion
        version : str, optional
```

lib/cartopy/io/shapereader.py
```diff
@@ -299,10 +301,11 @@ def natural_earth(resolution='110m', category='physical', name='coastline'):
     # get hold of the Downloader (typically a NEShpDownloader instance)
     # which we can then simply call its path method to get the appropriate
     # shapefile (it will download if necessary)
+    _version_string = "" if version is None else f"{version}/"
```
greglucas, line 304: This doesn't need to be private if it is just used within the method.
```suggestion
    version_string = "" if version is None else f"{version}/"
```

lib/cartopy/io/shapereader.py
```diff
@@ -386,7 +389,7 @@ def default_downloader():
 ne_{resolution}_{name}.shp
 
         """
-        default_spec = ('shapefiles', 'natural_earth', '{category}',
+        default_spec = ('shapefiles', 'natural_earth', '{category}', '{version}',
```
greglucas, line 392: Looks like the doctest failure is related to the docstring right above this. Where it got a `/{version}/` added in now. I actually don't think that should be included there now because it is path joining the version, but we don't want that if version is `None`, we want to leave off the `/`. I guess the question is does this need to be udpated since it is the default case (no version) or can we just leave it alone for now?

PR #2353, opened 2024-03-15
DOC: Update several links

lgolston:
## Rationale

Updates several dated links in the documentation, and fixes two dead links to the feature creation example.

## Other notes
On the install page (https://scitools.org.uk/cartopy/docs/latest/installing.html), I also suspect reference to OSGeo Live is outdated (at least, last mention on https://github.com/OSGeo/OSGeoLive/blob/master/CHANGES.txt is for version 0.17), but have not confirmed so left that alone.


CHANGES
```diff
@@ -1 +1 @@
-Please see docs/source/whats_new.rst for a changelog.
+Please see docs/source/whatsnew/index.rst for a changelog.
```
rcomer, line 1: The index file doesn’t really tell you anything, so maybe point to the directory so a specific version can be selected?  Alternatively have a link to the built docs?
```suggestion
Please see docs/source/whatsnew/ for a changelog.
```

lgolston: I think pointing to the directory makes sense, so have updated that plus addressed three more broken links and a redirect.

rcomer (approved):

PR #2354, opened 2024-03-18
DOC: Show module docstring

lgolston:
## Rationale

A remaining issue with the latest docs, comparing against https://scitools.org.uk/cartopy/docs/v0.17/, is that the module level docstring is not currently showing but can contain very useful information.

## Implications

This is fixed simply using automodule instead of module. However, in some cases documentation has been added to the rst file directly, which I have either transferred to the corresponding module/class, or combined if there was already a docstring, which should be better for maintainability of the information. I have have also tried to reduce repetition of 'this module... ' in several places for readability purposes.

### DIFF
--- a/docs/source/reference/crs.rst
+++ b/docs/source/reference/crs.rst
@@ -3,10 +3,7 @@
 Coordinate reference systems (CRS)
 ----------------------------------
 
-.. module:: cartopy.crs
-
-The :class:`cartopy.crs.CRS` class is the very core of cartopy, all coordinate reference systems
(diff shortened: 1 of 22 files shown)


dopplershift (approved):

lgolston:
Thanks for approving @dopplershift. Just wanted to make two small changes, which are added now:
- Combined text in one more place in cartopy.io I had missed.
- Added a link in Feature to FeatureArtist with regard to where zorder of 1.5 is defined, since that doc was slightly confusing.

PR #2356, opened 2024-03-19
FIX: Deepcopy of CRS with str/epsg

lgolston:
## Rationale
When using initializing crs.CRS with a string, or calling crs.epsg, there are errors when loading from a pickled copy or when performing deepcopy. This PR updates __reduce__ to get those processes working.

## Implications
Closes https://github.com/SciTools/cartopy/issues/2042 and the new issue recently reported in https://github.com/SciTools/cartopy/issues/1336

## Checklist
Two previously failing cases are added to the crs test fixture.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -33,7 +33,7 @@ jobs:
         if: matrix.python-version == '3.9' && matrix.os == 'ubuntu-latest'
         id: minimum-packages
         run: |
-          pip install cython==0.29.24 matplotlib==3.5.3 numpy==1.21 owslib==0.24.1 pyproj==3.1 scipy==1.6.3 shapely==1.7.1 pyshp==2.3.1
+          pip install cython==0.29.24 matplotlib==3.5.3 numpy==1.21 owslib==0.24.1 pyproj==3.3.1 scipy==1.6.3 shapely==1.7.1 pyshp==2.3.1
 
       - name: Coverage packages
         id: coverage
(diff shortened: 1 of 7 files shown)


lgolston:
The minimum packages run failed, and now I see pyproj has a related bug that was fixed in version 3.2.0 (https://github.com/pyproj4/pyproj/pull/901). Would bumping the minimum version from 3.1.0 to 3.2.0 be okay?

greglucas:
Yes, I think that should be fine, we've been trying to do roughly 2 years for dependencies and that was released in 2021.
https://github.com/pyproj4/pyproj/releases/tag/3.2.0

lgolston:
Okay, I test locally and v3.3.1 is actually minimum needed for this, and have pushed a new update.
https://github.com/pyproj4/pyproj/releases/tag/3.3.1

dopplershift:
I'm happy to go with Proj 3.3.1 since that's at 23 months old.

lgolston:
I realized that checking if self.proj4_params == {} was a little myopic. That made deepcopy work if the proj4_params input to crs.CRS was a string, but still did not work for a dict. The PR is now updated to handle both cases.

pyproj is bumped as before (and one place matplotlib to 3.5 was missed in #2276)

As an aside, I also think __getstate__ only needs to pop the _local variable, since it is unpickable, but just left it alone since it works either way.

There seem to be intermittent / unrelated failures recently that go away upon re-running (here with windows-latest, 3.9, false), which I do not understand.

lgolston:
Re-ran a fourth time and all tests are passing this time.
Run #1 there were a lot of errors in windows 3.9, I saw something similar when my conda environment was a little messed up before.
Run #2 and #3 there was a HTTP 502 error in downloading a few tiles from Google.
I think it is ready to go, but let me know if there are any issues.

dopplershift (approved):

PR #2367, opened 2024-04-10
TST: Pin pytest-mpl less than 0.17

greglucas:
v0.17 of pytest-mpl might be producing flaky results on CI, so pin it to the old version for now until a new fix comes out.

This is a WIP to see if it is actually the culprit or not. This is just an educated guess based on their release date in February and images failing that don't require external resources.
Possibly related to: https://github.com/matplotlib/pytest-mpl/issues/225

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,7 @@ speedups = ["pykdtree", "fiona"]
 ows = ["OWSLib>=0.20.0", "pillow>=6.1.0"]
 plotting = ["pillow>=6.1.0", "scipy>=1.3.1"]
 srtm = ["beautifulsoup4"]
-test = ["pytest>=5.1.2", "pytest-mpl>=0.11", "pytest-xdist", "pytest-cov", "coveralls"]
+test = ["pytest>=5.1.2", "pytest-mpl>=0.11,<0.17", "pytest-xdist", "pytest-cov", "coveralls"]
 
 [project.scripts]
 cartopy_feature_download = "cartopy.feature.download.__main__:main"


rcomer:
I was just looking at some of the failures from the last commit to main.  For some of the gridliner tests we appear to have lost an island!  (I think it's St Patrick's Island - furthest north on the 120W line).

Baseline:
baseline


Result:
result

rcomer:
Because several of the failures I looked at were due to missing bits of coastline, I had been wondering if this was some kind of local i/o problem within GHA.  Now I have a new failure:

```
E   fiona.errors.DriverError: 'C:\Users\runneradmin\.local\share\cartopy\shapefiles\natural_earth\physical\ne_110m_coastline.shp' not recognized as a supported file format.
```
https://github.com/rcomer/cartopy/actions/runs/8673960218/job/23785789451#step:9:170

greglucas:
🤔 I wonder if Fiona is causing issues?
1.9.6 was released March 8, which is around when we started seeing the flaky issues. https://github.com/Toblerity/Fiona/releases

rcomer:
> 🤔 I wonder if Fiona is causing issues? 1.9.6 was released March 8, which is around when we started seeing the flaky issues. https://github.com/Toblerity/Fiona/releases

I should probably have mentioned that I tried this: https://github.com/rcomer/cartopy/commit/1800e8b4afad2de2cd35740807896d4afdce7e2b

rcomer:
It does seem like there was some kind of change to actions storage space around the end of February, but this is getting way out of my wheelhouse https://github.com/actions/runner/issues/3184

greglucas:
Closing as this doesn't seem to be the problem.

Issue #2370, opened 2024-04-12
A GEOSException error occurred when calling contourf


Clarmy:
### Description

I encountered a strange `shapely.errors.GEOSException` error when using cartopy to draw maps with `contourf`. The plotting works fine if I just use matplotlib without the cartopy projection transformation.

sample data: https://drive.google.com/file/d/1n1PkwgZez3Ukz-0klkll2YAJX46vOH-6/view?usp=sharing

#### Code to reproduce
The code runs normally if I execute the following.
```python
import numpy as np
import matplotlib.pyplot as plt

data = np.load('./data.npy')
lons = np.load('./lons.npy')
lats = np.load('./lats.npy')

plt.contourf(lons, lats, data, levels=np.arange(0, 101, 5))
plt.savefig('./output.png')
```

However, if I execute the following code, it results in an error.
```python
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs

data = np.load('./data.npy')
lons = np.load('./lons.npy')
lats = np.load('./lats.npy')

fig = plt.figure(figsize=(20, 10))
ax = fig.add_subplot(111, projection=ccrs.GOOGLE_MERCATOR)
ax.contourf(lons, lats, data, levels=np.arange(0, 101, 5), cmap=plt.cm.PiYG, transform=ccrs.PlateCarree())
ax.set_extent((50, 160, 0, 60))
ax.set_axis_off()
fig.subplots_adjust(left=0, right=1, bottom=0, top=1)
fig.savefig('./output.png')
```

#### Traceback

```
GEOSException                             Traceback (most recent call last)
Cell In[7], line 11
      9 fig = plt.figure(figsize=(20, 10))
     10 ax = fig.add_subplot(111, projection=ccrs.GOOGLE_MERCATOR)
---> 11 ax.contourf(lons, lats, data, levels=np.arange(0, 101, 5), cmap=plt.cm.PiYG, transform=ccrs.PlateCarree())
     12 ax.set_extent((50, 160, 0, 60))
     13 ax.set_axis_off()

File ~/miniconda3/lib/python3.9/site-packages/cartopy/mpl/geoaxes.py:315, in _add_transform.<locals>.wrapper(self, *args, **kwargs)
    310     raise ValueError(f'Invalid transform: Spherical {func.__name__} '
    311                      'is not supported - consider using '
    312                      'PlateCarree/RotatedPole.')
    314 kwargs['transform'] = transform
[...]

Issue #1546, opened 2020-04-30
contourf plot from csv file on cartopy map - Error:  TopologyException: side location conflict at


jwynnsmith:
### Description
 I am using cartopy to produce a contour plot on a grid.  The csv file contains positive and negative float values.  I have used code nearly identical to this but have never gotten this error below.

#### Error
opologyException: side location conflict at 1.0537325376034968e+29 1.0537325376034968e+29
Traceback (most recent call last):
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/shapely/predicates.py", line 15, in __call__
    return self.fn(this._geom, other._geom, *args)
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/shapely/geos.py", line 548, in errcheck_predicate
    raise PredicateError("Failed to evaluate %s" % repr(func))
shapely.errors.PredicateError: Failed to evaluate <_FuncPtr object at 0x7fa3209f6cc8>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "statsplots_corr_map.py", line 58, in <module>
    mm = plt.contourf(xlon, ylat, grid, levels = levs, zorder=2, cmap=cmap)
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/matplotlib/pyplot.py", line 2530, in contourf
    **kwargs)
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/cartopy/mpl/geoaxes.py", line 1404, in contourf
    for col in result.collections
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/cartopy/mpl/geoaxes.py", line 1405, in <listcomp>
    if col.get_paths()])
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/matplotlib/collections.py", line 188, in get_datalim
    paths = [transform.transform_path_non_affine(p) for p in paths]
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/matplotlib/collections.py", line 188, in <listcomp>
    paths = [transform.transform_path_non_affine(p) for p in paths]
  File "/home/jwsmith/.conda/envs/glmval/lib/python3.6/site-packages/matplotlib/transforms.py", line 2390, in transform_path_non_affine
    return self._a.transform_path_non_affine(path)
[...]

PR #2373, opened 2024-04-19
fix: Invalid Polygon from Contouring (Issue #2370)

Clarmy:
## Rationale

Resolve the issue mentioned in #2370

## Implications

This change will affect the behavior of contourf and essentially fixes a bug in the contourf function.

### DIFF
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -190,8 +190,23 @@ def path_to_geos(path, force_ccw=False):
                     # is invalid. In this case, we can't use the contains method.
                     # Therefore, we need to perform a repair on the Polygon object
                     # and then attempt the contains method again. Related Issue: #2370
-                    polygon = collection[-1][0].buffer(0.001).buffer(-0.001)
-                    collection[-1] = (polygon, collection[-1][1])
+                    invalid_polygon = collection[-1][0]
+
+                    # Due to some invalid Polygon objects being connected by isolated points, 
+                    # using buffer(0) would convert them into a MultiPolygon, leading to more 
+                    # complex situations. Therefore, in this case, applying a small +/- buffer 
+                    # offset is used to connect these polygons together and resolve the issue.
+
+                    # To adapt to processing at different scales, the choice of the buffer 
+                    # parameter needs to be dynamically adjusted based on the size of the polygon. 
+                    # Here, 0.0001 times the maximum latitude and longitude difference of the 
+                    # polygon is used as the buffer size.
+                    minx, miny, maxx, maxy = invalid_polygon.bounds
+                    dimension = max(maxx - minx, maxy - miny)
+                    buffer_size = dimension * 0.0001
+
+                    valid_polygon = invalid_polygon.buffer(buffer_size).buffer(-buffer_size)
+                    collection[-1] = (valid_polygon, collection[-1][1])
                     is_inside = collection[-1][0].contains(geom.exterior)
             else:
                 is_inside = False
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -14,8 +14,8 @@ and `Matplotlib Path API <https://matplotlib.org/stable/api/path_api.html>`_.
 
 from matplotlib.path import Path
 import numpy as np
-import shapely.geometry as sgeom
 import shapely
+import shapely.geometry as sgeom
 
 if shapely.__version__ < '2.0.0':
     from shapely.errors import TopologicalError as GEOSException
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -190,7 +190,7 @@ def path_to_geos(path, force_ccw=False):
                     # is invalid. In this case, we can't use the contains method.
                     # Therefore, we need to perform a repair on the Polygon object
                     # and then attempt the contains method again. Related Issue: #2370
-                    polygon = collection[-1][0].buffer(0)
+                    polygon = collection[-1][0].buffer(0.001).buffer(-0.001)
                     collection[-1] = (polygon, collection[-1][1])
                     is_inside = collection[-1][0].contains(geom.exterior)
             else:
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -17,6 +17,7 @@ import numpy as np
 import shapely
 import shapely.geometry as sgeom
 
+
 if shapely.__version__ < '2.0.0':
     from shapely.errors import TopologicalError as GEOSException
 else:
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -22,6 +22,7 @@ if shapely.__version__ < '2.0.0':
 else:
     from shapely.errors import GEOSException
 
+
 def geos_to_path(shape):
     """
     Create a list of :class:`matplotlib.path.Path` objects that describe
@@ -175,34 +176,34 @@ def path_to_geos(path, force_ccw=False):
         # lake within a land mass).  Sometimes there is a further geom within
         # this interior (e.g. an island in a lake, or some instances of
         # contours).  This needs to be a new external geom in the collection.
-        try:
-            is_inside = (len(collection) > 0 and
-                isinstance(collection[-1][0], sgeom.Polygon) and
-                isinstance(geom, sgeom.Polygon) and
-                collection[-1][0].contains(geom.exterior))
-        except GEOSException:
-            # If the GEOSException is raised, it is likely that the polygon
-            # is invalid.  In this case, we can't use the contains method.
-            # Therefore, we need to perform a repair on the Polygon object
-            # and then attempt the contains method again. Related Issue: #2370
-            polygon = collection[-1][0].buffer(0)
-            collection[-1] = (polygon, collection[-1][1])
-            is_inside = (len(collection) > 0 and
-                isinstance(collection[-1][0], sgeom.Polygon) and
-                isinstance(geom, sgeom.Polygon) and
-                collection[-1][0].contains(geom.exterior))
-
         if geom.is_empty:
             pass
-        elif is_inside:
-            if any(internal.contains(geom) for internal in collection[-1][1]):
-                collection.append((geom, []))
-            else:
-                collection[-1][1].append(geom)
-        elif isinstance(geom, sgeom.Point):
-            other_result_geoms.append(geom)
         else:
-            collection.append((geom, []))
+            if (len(collection) > 0 and
+                    isinstance(collection[-1][0], sgeom.Polygon) and
+                    isinstance(geom, sgeom.Polygon)):
+                try:
+                    is_inside = collection[-1][0].contains(geom.exterior)
+                except GEOSException:
+                    # If the GEOSException is raised, it is likely that the polygon
+                    # is invalid. In this case, we can't use the contains method.
+                    # Therefore, we need to perform a repair on the Polygon object
+                    # and then attempt the contains method again. Related Issue: #2370
+                    polygon = collection[-1][0].buffer(0)
+                    collection[-1] = (polygon, collection[-1][1])
+                    is_inside = collection[-1][0].contains(geom.exterior)
+            else:
+                is_inside = False
+
+            if is_inside:
+                if any(internal.contains(geom) for internal in collection[-1][1]):
+                    collection.append((geom, []))
+                else:
+                    collection[-1][1].append(geom)
+            elif isinstance(geom, sgeom.Point):
+                other_result_geoms.append(geom)
+            else:
+                collection.append((geom, []))
 
     # Convert each (external_geom, [internal_polygons]) pair into a
     # a shapely Polygon that encapsulates the internal polygons, if the
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -15,7 +15,12 @@ and `Matplotlib Path API <https://matplotlib.org/stable/api/path_api.html>`_.
 from matplotlib.path import Path
 import numpy as np
 import shapely.geometry as sgeom
-from shapely.errors import GEOSException
+import shapely
+
+if shapely.__version__ < '2.0.0':
+    from shapely.errors import TopologicalError as GEOSException
+else:
+    from shapely.errors import GEOSException
 
 def geos_to_path(shape):
     """
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -192,21 +192,22 @@ def path_to_geos(path, force_ccw=False):
                     # and then attempt the contains method again. Related Issue: #2370
                     invalid_polygon = collection[-1][0]
 
-                    # Due to some invalid Polygon objects being connected by isolated points, 
-                    # using buffer(0) would convert them into a MultiPolygon, leading to more 
-                    # complex situations. Therefore, in this case, applying a small +/- buffer 
-                    # offset is used to connect these polygons together and resolve the issue.
+                    # Due to some invalid Polygon objects being connected by isolated
+                    # points, using buffer(0) would convert them into a MultiPolygon,
+                    # leading to more complex situations. Therefore, in this case,
+                    # applying a small +/- buffer offset is used to connect these
+                    # polygons together and resolve the issue.
 
-                    # To adapt to processing at different scales, the choice of the buffer 
-                    # parameter needs to be dynamically adjusted based on the size of the polygon. 
-                    # Here, 0.0001 times the maximum latitude and longitude difference of the 
-                    # polygon is used as the buffer size.
+                    # To adapt to processing at different scales, the choice of the
+                    # buffer parameter needs to be dynamically adjusted based on the
+                    # size of the polygon. Here, 0.0001 times the maximum latitude and
+                    # longitude difference of the polygon is used as the buffer size.
                     minx, miny, maxx, maxy = invalid_polygon.bounds
                     dimension = max(maxx - minx, maxy - miny)
                     buffer_size = dimension * 0.0001
 
-                    valid_polygon = invalid_polygon.buffer(buffer_size).buffer(-buffer_size)
-                    collection[-1] = (valid_polygon, collection[-1][1])
+                    polygon = invalid_polygon.buffer(buffer_size).buffer(-buffer_size)
+                    collection[-1] = (polygon, collection[-1][1])
                     is_inside = collection[-1][0].contains(geom.exterior)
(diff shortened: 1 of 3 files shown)


lib/cartopy/mpl/patch.py
```diff
@@ -170,12 +175,26 @@ def path_to_geos(path, force_ccw=False):
...
+            polygon = collection[-1][0].buffer(0)
+            collection[-1] = (polygon, collection[-1][1])
+            is_inside = (len(collection) > 0 and
                 isinstance(collection[-1][0], sgeom.Polygon) and
                 isinstance(geom, sgeom.Polygon) and
-                collection[-1][0].contains(geom.exterior)):
+                collection[-1][0].contains(geom.exterior))
+
+        if geom.is_empty:
+            pass
```
rcomer, line 196: Since this check is cheap compared with the contains method, I think it would be good to keep it first and only define `is_inside` if `geom` is not empty.

rcomer (commented):
Hi @Clarmy thanks for digging into this and sorry I didn't get a chance to think about it properly till now.  Your change makes sense to me and I just have a couple of minor comments about the code structure.

Ideally this should also get a test but, since this problem is so data dependent, I'm struggling to think how it could be written without uploading a lot of sample data (which we don't want to do!)

One other thing I'm not sure about is which branch the PR should target.  @greglucas since you have handled recent releases can you comment on that?

lib/cartopy/mpl/patch.py
```diff
@@ -170,12 +175,26 @@ def path_to_geos(path, force_ccw=False):
         # lake within a land mass).  Sometimes there is a further geom within
         # this interior (e.g. an island in a lake, or some instances of
         # contours).  This needs to be a new external geom in the collection.
-        if geom.is_empty:
-            pass
-        elif (len(collection) > 0 and
+        try:
+            is_inside = (len(collection) > 0 and
+                isinstance(collection[-1][0], sgeom.Polygon) and
+                isinstance(geom, sgeom.Polygon) and
```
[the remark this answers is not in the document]
Clarmy: I agree, I will modify and test it later.

greglucas:
> One other thing I'm not sure about is which branch the PR should target. @greglucas since you have handled recent releases can you comment on that?

We should target main and then backport the commit to the other branch if we want to make a v0.23.1 release. I've updated this PR to target main now.

Do we know why we are getting duplicate points/bad geometries from our other code paths? This is likely OK to fix here, but it might be nice to at some point to figure out why we are producing the bad geometries in the first place and see if we can fix this issue at the source.

Clarmy:
@rcomer I'm not very familiar with the CI process for cartopy, but I see that some tests have failed in the CI, and pre-commit.ci has also been failing continuously. I'm not sure what impact this has on the final merge, and how to check and resolve the issue of inconsistent test drawing results under some versions in the CI?

lgolston:
pre-commit is failing on the ruff step, with the issue here just being that `import shapely` should go above `import shapely.geometry as sgeom`.

For the test failures on two of the builds, it shouldn't affect this PR as it appears related to an unexplained issue where some builds are intermittently failing (currently being investigated in #2367).

lgolston:
Guess it also is looking for two empty lines after the the last import, instead of one.

Clarmy:
@lgolston It worked! Thank you for your guidance.

rcomer:
> Do we know why we are getting duplicate points/bad geometries from our other code paths? This is likely OK to fix here, but it might be nice to at some point to figure out why we are producing the bad geometries in the first place and see if we can fix this issue at the source.

If we plot the data with Matplotlib's `contour` and zoom in, we can see the problem loop in the middle:

image

So I agree with @Clarmy that the problem does not originate in Cartopy and the best we can do here is handle it when it comes up.  We could open an issue with Contourpy to see if there is anything to be done there though.

rcomer:
I have opened an issue at https://github.com/contourpy/contourpy/issues/377, and in the process came up with an example that I think is small enough to make into a test (though I see you have also added a test in the meantime @Clarmy):

```python
import numpy as np
import cartopy.crs as ccrs

lons = [96.75, 97.  , 97.25]
lats = np.arange(17.25, 15.9, -0.25)

data = [[26.9, 43.7, 26.8],
        [33.2, 65.8, 34.6],
        [54.7, 66.9, 55.5],
        [65.3, 65. , 65.7],
        [65.9, 65. , 65.6],
        [65.8, 65.2, 65.5]]

fig = plt.figure()
ax = fig.add_subplot(projection=ccrs.Mercator())
ax.contourf(lons, lats, data, levels=[60, 65], transform=ccrs.PlateCarree())
```

Clarmy:
@rcomer Thank you for your continued effort on this issue; your test case is superbly done. The test case I added was specifically for testing the function `path_to_geos`, whereas yours more closely matches the actual plotting scenario. However, I'm unsure about the best script to insert your test case into. Could you possibly add a suggestion during the code review?

rcomer:
> test case is superbly done

I just took your data and sliced it as much as I could while still reproducing the error!  It's great to have the example by the way - the error has been reported previously (e.g. #1546) but we did not have enough information to reproduce.

We have a test file for contour plots here:
https://github.com/SciTools/cartopy/blob/main/lib/cartopy/tests/mpl/test_contour.py

rcomer:
By the way, you can also install pre-commit locally to help with linting each time you commit: https://scitools.org.uk/cartopy/docs/latest/contribute.html

Clarmy:
Okay, I understand. Thank you very much!

Clarmy:
Sorry, I've found that the modified code introduced a new bug in actual use, and I'm currently addressing it. I will provide a more detailed description later.

Clarmy:
The situation we are now facing is more complex than it was at the beginning. I've discovered a new issue: if we use [this data](https://drive.google.com/file/d/11ke5iViG3OBLIxQ8WFUMAdTmAX6QKKN2/view?usp=sharing) to plot a graph, it still causes an error when run on my modified cartopy code. The error it generates is:

```
  File "/opt/main.py", line 105, in <module>
    rc.plot(
  File "/opt/plot.py", line 237, in plot
    ax.contourf(
  File "/usr/local/lib/python3.9/site-packages/cartopy/mpl/geoaxes.py", line 307, in wrapper
  File "/usr/local/lib/python3.9/site-packages/cartopy/mpl/geoaxes.py", line 351, in wrapper
  File "/usr/local/lib/python3.9/site-packages/cartopy/mpl/geoaxes.py", line 1653, in contourf
    self.update_datalim(result.get_datalim(self.transData))
  File "/usr/local/lib/python3.9/site-packages/matplotlib/collections.py", line 267, in get_datalim
    paths = [transform.transform_path_non_affine(p) for p in paths]
  File "/usr/local/lib/python3.9/site-packages/matplotlib/collections.py", line 267, in <listcomp>
    paths = [transform.transform_path_non_affine(p) for p in paths]
  File "/usr/local/lib/python3.9/site-packages/matplotlib/transforms.py", line 2426, in transform_path_non_affine
    return self._a.transform_path_non_affine(path)
  File "/usr/local/lib/python3.9/site-packages/cartopy/mpl/geoaxes.py", line 175, in transform_path_non_affine
    geoms = cpatch.path_to_geos(src_path)
  File "/usr/local/lib/python3.9/site-packages/cartopy/mpl/patch.py", line 216, in path_to_geos
    geom = sgeom.Polygon(external_geom.exterior, exteriors)
AttributeError: 'MultiPolygon' object has no attribute 'exterior'
```

The reason for this error is that in my modified code, the step `collection[-1][0].buffer(0)` that fixes topological errors results in a `MultiPolygon` object (originally it was a `Polygon`), meaning the polygon object type changes during topological repair.

To investigate the cause, I exported both the original `Polygon` and the new `MultiPolygon` objects into GeoJSON files, and then displayed them in QGIS:

image

It seems there is no apparent problem, so why does it turn into multiple polygons? After reading the `MultiPolygon` object, I found it consists of two separate polygons. The smaller of these polygons is located around the coordinates [53.0, 47.5] and is a small polygon composed of over 20 points. When I zoom into this area in QGIS, I can see that there is actually a **breakpoint** that has **cut** the polygon into two parts. This is somewhat similar to the 'conflict points' we encountered before, but the impact of this conflict point is different this time.

image

I'm sharing this information for now, and as for the solution, I am still exploring.

[fix_polygon.geojson.zip](https://github.com/SciTools/cartopy/files/15089612/fix_polygon.geojson.zip)
[raw_polygon.geojson.zip](https://github.com/SciTools/cartopy/files/15089613/raw_polygon.geojson.zip)

lib/cartopy/mpl/patch.py
```diff
@@ -172,18 +179,32 @@ def path_to_geos(path, force_ccw=False):
...
+                    isinstance(collection[-1][0], sgeom.Polygon) and
+                    isinstance(geom, sgeom.Polygon)):
+                try:
+                    is_inside = collection[-1][0].contains(geom.exterior)
+                except GEOSException:
+                    # If the GEOSException is raised, it is likely that the polygon
+                    # is invalid. In this case, we can't use the contains method.
+                    # Therefore, we need to perform a repair on the Polygon object
+                    # and then attempt the contains method again. Related Issue: #2370
+                    polygon = collection[-1][0].buffer(0.001).buffer(-0.001)
```
Clarmy, line 193: I have tried using the .buffer(0.001).buffer(-0.001) function to perform a dilation and contraction operation on the polygon, and so far, the results are quite satisfactory. Although the code is not particularly elegant, it can solve the actual problem. At the same time, it's important to note that this kind of scaling operation produces a slight rounded effect in the inner corners. When I set the scaling parameter small enough (for example, 0.001), the difference is almost negligible.

image

image

greglucas:
@Clarmy Are you able to recursively or iteratively add the geometries to the original set of geometries? I think it makes sense to have multiple distinct polygons if we need them and not buffer to add artificial gaps to connect geometries.
```python
if geom is multi-polygon:
    for poly in geom:
        # Might need to check for linestrings here too?
        append to collection
elif geom is Polygon:
    append to collection
```

Perhaps using "make_valid" is better than try/except as well? https://github.com/contourpy/contourpy/issues/377
Also, see the two examples here: 
https://shapely.readthedocs.io/en/stable/manual.html#validation.make_valid
which I think show the two potential issues we're running into: An bowtie geometry creating two different polygons and an extra linestring hanging around.

rcomer:
Just to note that I tried `make_valid` on @Clarmy's original bad polygon and it took about 50x longer than the buffer approach.  Also `make_valid` only came in at Shapely v1.8 and we currently test against v1.7.

I was also thinking that try/except should be faster than `is_valid` since the vast majority of these contours _are_ valid, but I don't think I actually checked that.

Clarmy:
@greglucas  I have referred to the example in [contourpy/contourpy#377](https://github.com/contourpy/contourpy/issues/377), but unfortunately, based on my actual testing, I found that the `make_valid` function does not guarantee the return type as a shapely `Polygon` object. Depending on the input data, it could also return a `MultiPolygon` or even a `GeometryCollection` type , which obviously results in uncontrollable outcomes and introduces more complexity and unpredictability to the program. Any data type other than `Polygon` could cause subsequent program failures. Additionally, regarding the use of `try/except`, I initially wanted to check each `geom` object with the `.is_valid` function, but I found that the check itself incurs significant time overhead, making the function almost unusable, as @rcomer described. If using `try/except`, it skips the normal polygons and only handles the problematic parts, minimizing the program's additional overhead.

Clarmy:
I'm not sure if I'm misunderstanding the `contour/contourf` function, but I've always thought that each `Path` in the collection object returned by the matplotlib `contour/contourf` function represents a separate polygon, without any disconnected polygons appearing within the same `Path` object (or even if there was a "break" as mentioned earlier, they should theoretically belong to the same polygon, just due to some kind of "coincidence"). So, my approach is not to preserve the separated structure when a MultiPolygon occurs, but rather to think about how to merge them together.

rcomer:
countour/contourf used to create one path per contour as you say but there was a change at Matplotlib v3.8 so they now create one path per level.
https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.8.0.html#contourset-is-now-a-single-collection

Note that this part of Cartopy is quite general and handles anything that is drawn with a path (which I think is most things?)  I think the main other example of needing to sort out internal/external polygons is for the FeatureArtist, hence the code comments about land masses with lakes in.  So at this point we don't actually know if we are working with contours or something else.

It does bother me that, specifically for the contour case the overall process is effectively
* ContourPy creates individual contours
* Matplotlib concatenates the contours and makes paths
* Cartopy splits the paths back down and tries to figure out which contours are inside each other

It would be nice if we could _somehow_ intercept the contours straight out of ContourPy when it is presumably more obvious what the individual contours are.  I did have a couple of attempts studying the relevant parts of Matplotlib to figure that out, but I didn't come up with anything.

Clarmy:
I understand now, so `path_to_geos` is a very general function that handles all Path objects built with matplotlib, converting them into shapely geometric objects for further projection transformations and other tasks. This function can be applied not only to contours but also to national borders, inland water systems, and more. Is my understanding correct?

If so, I believe that the modifications in this PR should not affect other functionalities (assuming they were previously working properly). The issue of generating invalid polygon objects would likely only occur when generating contours, as these are interpolated on the spot from grid data. For other data like those used in FeatureArtist, which are pre-generated, the input graphic paths should theoretically already be free from topological errors and valid.

Of course, I cannot guarantee that my understanding is completely accurate, so please point out any obvious errors if there are any.

Moreover, if we directly obtain the paths from ContourPy, does it imply that this part of the code needs to be refactored?

rcomer:
I think your understanding is correct

> if we directly obtain the paths from ContourPy, does it imply that this part of the code needs to be refactored?

This is really just a vague aspiration and I don't know if it's even possible, so I wouldn't worry about that.

greglucas:
If we are getting 1 `Path` per level now from contouring and that is causing issues (?), would we need to somehow create MultiPolygons per level, while also accounting for holes? It looks like we create MultiLineStrings, but not MultiPolygons later in the function. However, I think that is after the errors that are cropping up, so probably not related.

The +/- buffering seems like a hack to me and may not be suitable for all resolutions / scales. One other thought that comes to mind is that this is due to the `contains` check failing. Is there anything we can say about the contains and set `is_inside = False` if it raises the exception and avoid modifying the geometry?

Clarmy:
> The +/- buffering seems like a hack to me and may not be suitable for all resolutions / scales. One other thought that comes to mind is that this is due to the `contains` check failing. Is there anything we can say about the contains and set `is_inside = False` if it raises the exception and avoid modifying the geometry?

I can start by running some tests to see.

Clarmy:
@greglucas Regarding your suggestions, I have conducted the following work:

1. To understand if it is possible to directly set `is_inside` to False when a `GEOSException` occurs, I checked the actual value of `is_inside` after using `buffer(0)` for repair when the program encounters the exception. I found that in my tests, the actual value is True. This result at least indicates that setting it to False directly when an error occurs could lead to some results that are inconsistent with expectations (even if these inconsistencies might be very slight).
2. After repairing the polygon to a multipolygon using `buffer(0)`, I tried to perform type judgment and then handle the `MultiPolygon` separately based on my understanding, attempting to integrate it back into the original collection. However, this ultimately failed, resulting in significant color loss in the rendered images. My current view is that the order of polygons in the collection is important. If we try to integrate the `MultiPolygon` individually, it will disrupt the original order of the list in the collection, leading to unpredictable graphical issues.
3. Regarding your comment that a single buffer value is not suitable for all scales, I agree. Therefore, I attempted to write some code to dynamically generate a sufficiently small buffer scale parameter based on the actual size of the polygon to address this issue.

rcomer:
I tried @Clarmy's original example with an old environment we have lying around (Cartopy 0.18, Matplotlib 3.3.4) and the error is the same, so I don't think the recent `ContourSet` changes caused this.

Does the `contains` method work on the `MultiPolygon` and `GeometryCollection`?  If so I wonder if we could use the output of `buffer(0)` to define `is_inside`, but not actually replace the original polygon.  We don't see errors when plotting the same data with `contour`, so presumably the actual transformation and plotting steps don't care if the polygon is invalid.  Downside would be that we could end up in the `except` multiple times for a given bad polygon.

[I see see @Clarmy pushed another change while I was typing this so maybe this suggestion is redundant.]

Clarmy:
> If so I wonder if we could use the output of `buffer(0)` to define `is_inside`, but not actually replace the original polygon.

I just tried locally to only judge the value of is_inside without modifying the collection, but in the end, it still throws a GEOSException elsewhere, as you mentioned, which might require multiple exception catches across various modules.

greglucas:
Where does this PR stand? Is it ready for review or were we going to update some of the other areas of code to handle these cases instead?
https://github.com/SciTools/cartopy/pull/2373#issuecomment-2079046227

MHBalsmeier:
Hi all,
I encountered the same problem when plotting wave data from the German CWAM wave model, this is a small script reproducing it (cartopy 0.23.0):

`import cartopy.crs as ccrs
import numpy as np
import netCDF4 as nc

ds = nc.Dataset("cwam_sig_wave_height.nc", "r", format = "NETCDF4")
lat_deg = ds["lat_deg"][:]
lon_deg = ds["lon_deg"][:]
swh = ds["wave_height"][:]
ds.close()

# works
fig = plt.figure()
ax = plt.axes()
ax.contourf(lon_deg, lat_deg, swh)
plt.show()

# does not work
fig = plt.figure()
proj = ccrs.Mercator(central_longitude = 11.8)
ax = plt.axes(projection = proj)
ax.contourf(lon_deg, lat_deg, swh, transform = ccrs.PlateCarree())
`
[cwam_nc.zip](https://github.com/user-attachments/files/15980166/cwam_nc.zip)

Maybe this helps for testing.
It looks like a complex issue, currently I cannot help any further, but maybe I'll find the time to dig into it as well.

greglucas:
Sorry for losing track of this @Clarmy, are you interested in coming back to finish this up?

@rcomer just moved/renamed and refactored this function a bit in #2455 and now we depend on shapely>=2.0 so some things might be able to be simplified here now.

Clarmy:
@greglucas Sorry for not paying attention to this PR for quite some time. It seems that there have been significant changes in Cartopy's code since this PR was submitted. I'm also not sure if the issue that this PR aims to address still exists. I can run some tests with the latest code and the previous data to see if the issue has already been resolved. It might turn out that there's nothing more for me to do.

Clarmy:
@greglucas @rcomer Due to this PR being outdated, there are some conflicts with the latest code as versions have iterated. Therefore, I made modifications based on the latest code and submitted a new PR: #2463.
I have tested the code myself, and it does solve the previous issues, although some of the implementations are not very elegant. It's up to your discretion; if you prioritize solving the problem over the elegance of the code, you can approve and merge this PR.

Clarmy:
New PR: https://github.com/SciTools/cartopy/pull/2463

Issue #2142, opened 2023-03-02
Nightshade coordinates for different times don't change


dchaddock:
### Description

I am playing around with using the Nightshade feature and in need of extracting the lat/lon coordinates from the returned Shapely object, but running into some funky behaviour that is confusing me. I am able to get the nightshade, but when extracting the coordinates, they appear to never change based on the specified date. However, when plotting the feature onto a Cartopy map, the coordinates are clearly changing!

I am using the latest version of Cartopy (0.21.1). Here's a code snippet of how I am extracting the coordinates, along with screenshots of a more complete test.

#### Code to reproduce

```python
date = datetime.datetime(2023, 2, 27, 0, 0)
terminator_daynight1 = Nightshade(date, alpha=0.2, refraction=0)

terminator_polygon1 = None
for poly in terminator_daynight1.geometries():
    terminator_polygon1 = poly
    
pprint.pprint(terminator_polygon1.exterior.coords[100:110])

----
[(-90.0, -79.99444135630907),
 (-90.0, -79.89438576987214),
 (-90.0, -79.79433018343524),
 (-90.0, -79.69427459699833),
 (-90.0, -79.59421901056142),
 (-90.0, -79.49416342412451),
 (-90.0, -79.3941078376876),
 (-90.0, -79.29405225125069),
 (-90.0, -79.19399666481378),
 (-90.0, -79.09394107837687)]
```

![image](https://user-images.githubusercontent.com/10278552/222526318-17cf2500-abac-4f98-8f2c-3d4c5d387e64.png)

![image](https://user-images.githubusercontent.com/10278552/222526343-6d262a30-be4d-4b36-817e-f624a80ed04b.png)

The point here with the screenshots is to show that the Shapely Polygones returned from two Nightshade() calls with different times, give the same coordinates but display different coordinates when plotted.

I noticed while debugging a bit that the Shapely Polygons are weak references, so I wonder if could be something to do with needing to do a `deepcopy` before extracting coordinates?

Any help with this would be great. Thanks Cartopy team.

### Operating system

Behaviour is seen on both Windows 11 and RHEL 8. Using Anaconda on Windows and Miniconda on Linux. Conda venv using Python 3.9.

### Cartopy version

0.21.1, installed using conda.

### conda list

```
# Name                    Version                   Build  Channel
_ipyw_jlab_nb_ext_conf    0.1.0            py39haa95532_0
alabaster                 0.7.12             pyhd3eb1b0_0
[...]

PR #2375, opened 2024-04-25
Update nightshade.py to allow user-defined sub-solar coordinates

dsheward-astro:
This allows for the specification of the sub solar coordinates directly, rather than calculating them using the date. This increases the usefulness of nightshade to those of us using cartopy for non-Earth bodies.

## Rationale

Currently Nightshade produces the shade based off the sub-solar point at a defined time, based on the Earth. As cartopy is used for modelling other solar-system bodies, such as the Moon, which will have differing sub-solar points at a given time, the current use case for Nightshade is exclusive to the Earth. By allowing for user-defined sub-solar coordinates, this use case is extended to any sphere. 

## Implications

Nightshade is now also useable for Non-Earth bodies

### DIFF
--- a/lib/cartopy/feature/nightshade.py
+++ b/lib/cartopy/feature/nightshade.py
@@ -14,7 +14,7 @@ from . import ShapelyFeature
 
 class Nightshade(ShapelyFeature):
     def __init__(self, date=None, delta=0.1, refraction=-0.83,
-                 color="k", alpha=0.5, **kwargs):
+                 color="k", alpha=0.5, sub_solar_coords=None, **kwargs):
         """
         Shade the darkside of the Earth, accounting for refraction.
 
@@ -47,8 +47,12 @@ class Nightshade(ShapelyFeature):
 
         # Returns the Greenwich hour angle,
         # need longitude (opposite direction)
-        lat, lon = _solar_position(date)
+        if sub_solar_coords is None:        #if not specified proceed as usual
+            lat, lon = _solar_position(date)
+        else:
+            lat, lon = sub_solar_coords     #if specified, use given coordinates
         pole_lon = lon
+                     
         if lat > 0:
             pole_lat = -90 + lat
             central_lon = 180


lib/cartopy/feature/nightshade.py
```diff
@@ -14,7 +14,7 @@
 
 class Nightshade(ShapelyFeature):
     def __init__(self, date=None, delta=0.1, refraction=-0.83,
-                 color="k", alpha=0.5, **kwargs):
+                 color="k", alpha=0.5, sub_solar_coords=None, **kwargs):
```
greglucas, line 17: `sub_solar_point`? `coords` seems like it could be an array of coordinates to me.

You will need to add documentation about what this is (tuple of latitude, longitude), the units (degrees), and how it interacts with date (takes precedence if it is provided).

greglucas (commented):
I think this makes sense in general to be able to use the functionality on different bodies. I am not sure the best approach for providing either a date or a sub-solar point, but this seems reasonable.

You should also add a few tests making sure it is producing the values you expect.

lib/cartopy/feature/nightshade.py
```diff
@@ -29,6 +29,11 @@ def __init__(self, date=None, delta=0.1, refraction=-0.83,
         refraction : float
             The adjustment in degrees due to refraction,
             thickness of the solar disc, elevation etc...
+        sub_solar_point : tuple of latitude, longitude
+            The coordinates in degrees of the point at which the sun is directly 
+            overhead. If both sub_solar_point and date are provided, sub_solar_point
+            takes precedence.
+
```
dopplershift, line 36: Extra blank line snuck in here.

dopplershift (changes requested):
Minor nits. Looks like pre-commit is complaining about trailing whitespace (might be the line I pointed out). Feature seems like a nice add.

lgolston:
The documentation should still be adjusted here (main docstring still says "Shade the darkside of the Earth..."; which is no longer true for Nightshade, but is still true that_solar_position is specific to Earth).

lib/cartopy/feature/nightshade.py
```diff
@@ -29,6 +29,11 @@ def __init__(self, date=None, delta=0.1, refraction=-0.83,
         refraction : float
             The adjustment in degrees due to refraction,
             thickness of the solar disc, elevation etc...
+        sub_solar_point : tuple of latitude, longitude
```
lgolston, line 32:
```suggestion
        sub_solar_point : tuple of latitude, longitude, optional
```
I guess we should note all the inputs are optional.

lib/cartopy/feature/nightshade.py
dsheward-astro: I have removed the extra whitespace, thanks.

lib/cartopy/feature/nightshade.py
dsheward-astro: I have implemented the 'optional' tag, thanks.

greglucas (approved):

greglucas:
@dsheward-astro, would you be able to add a quick test for this? Let me know if you want any pointers for where to add it.

I also see that you added the subsolar point as `(lat, lon)`, is that the order we are wanting all of our public facing conventions to be? In all of our plot functions we generally have `(lon, lat)` due to the x, y ordering. But this also seems fine and is documented so I don't have a strong preference either way, maybe a +0.1 towards `(lon, lat)` if I had to choose.

alexstoken:
While testing this on another planetary body, I found that the class likely needs an option to pass an alternative globe (which would then be passed through to the `ccrs.RotatedPole()` in order to support non-Earth shades.

```
rotated_pole = ccrs.RotatedPole(
    pole_latitude=pole_lat,
    pole_longitude=pole_lon,
    central_rotated_longitude=central_lon,
    globe=kwargs.get("globe", None),
)
```

The shade can then be converted to lonlat as follows (from https://github.com/SciTools/cartopy/issues/2142#issuecomment-1452451470) 

```
globe = ccrs.Globe(
    semimajor_axis=BODY_RADIUS_METERS,
    semiminor_axis=BODY_RADIUS_METERS,
    ellipse=None,
)
nightshade = Nightshade(
    refraction=-0.01, sub_solar_point=(ss_lon, ss_lat), globe=globe
)
geom = next(nightshade.geometries())
x, y = np.array(geom.exterior.coords).T

latlon = ccrs.Geodetic(globe)

lons, lats, _ = latlon.transform_points(nightshade.crs, x, y).T
shade = shapely.Polygon([lon, lat] for lon, lat in zip(lons, lats))
```

And users should be careful with the refraction value for other bodies, as the default only applies to Earth.

PR #2376, opened 2024-04-26
CI: add scheduled testing against nightlies

rcomer:
## Rationale

This PR removes `shapely-dev` from the CI test matrix and instead introduces monthly testing against the Shapely, SciPy and Matplotlib nightly builds (note Numpy and ContourPy nightlies are automatically pulled in with Matplotlib).

The [guidance from Scientific Python](https://scientific-python.org/specs/spec-0004/#test-with-nightly-wheels) recommends a weekly job for this.  At https://github.com/SciTools/cartopy/pull/2371#discussion_r1564037949, @greglucas suggested a monthly job.  I also lean towards monthly just because Cartopy doesn't have a lot of resource so with a more frequent run we probably would end up closing lots of duplicate issues that we didn't get around to looking at yet.  The choice of 10th of the month is pretty arbitrary but it seems good to move it away from the beginning of the month so we aren't bothered by it on New Year's Day! I also added a `workflow-dispatch` trigger so we can run it if and when we have any reason to worry about recent updates in any package.

For the approach I roughly followed [Matplotlib's workflow](https://github.com/matplotlib/matplotlib/blob/main/.github/workflows/tests.yml).  It feels a little excessive to run the whole matrix but I couldn't find a neat way to restrict it for just these workflow triggers.  I took the command to install the nightlies from the [Scientific Python page](https://scientific-python.org/specs/spec-0004/#test-with-nightly-wheels).  It's a little different from what Matplotlib does, and what's recommended in [Matplotlib's installation documentation](https://matplotlib.org/stable/users/installing/index.html#install-a-nightly-build) but I don't know the reason for the difference or how to choose between them.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -1,6 +1,12 @@
 name: Tests
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    # 10th of each month
+    - cron: "0 0 10 * *"
+  workflow_dispatch:
 
 jobs:
   build:
@@ -12,13 +18,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: [3.9, '3.10', '3.11', '3.12']
-        shapely-dev: [false]
         use-network: [true]
         include:
-          - os: ubuntu-latest
-            python-version: '3.11'
-            shapely-dev: true
-            use-network: true
           - os: ubuntu-latest
             python-version: '3.11'
             use-network: false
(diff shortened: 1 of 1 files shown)


rcomer:
Put in draft for now as I realised I was getting Matplotlib 3.9rc2 instead of the nightly build.

rcomer:
I now have a run that demonstes installing nightlies of all five expected packages: https://github.com/rcomer/cartopy/actions/runs/8858366152

I have also learned that if you still get failures after re-spin, you get duplicate issues: https://github.com/rcomer/cartopy/issues/12, https://github.com/rcomer/cartopy/issues/13, https://github.com/rcomer/cartopy/issues/14.  I think the flakiness of our tests is another good reason to keep the schedule fairly low frequency.

.github/workflows/ci-testing.yml
```diff
@@ -108,5 +111,30 @@ jobs:
...
         if: failure()
         with:
-          name: image-failures-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.shapely-dev }}
-          path: cartopy_test_output-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.shapely-dev }}
+          name: image-failures-${{ matrix.os }}-${{ matrix.python-version }}
+          path: cartopy_test_output-${{ matrix.os }}-${{ matrix.python-version }}
+
+  # Separate dependent job to only upload one issue from the matrix of jobs
+  create-issue:
+    if: ${{ failure() && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
```
greglucas, line 119: Do we want workflow_dispatch to upload an issue as well, or just the scheduled run?

greglucas (approved):
Looks good, and I agree monthly seems like the right cadence for now.

.github/workflows/ci-testing.yml
rcomer: Fair point.  I guess if you've done workflow_dispatch then you are actively looking at it and don't need the issue.

PR #2380, opened 2024-05-01
MNT: drop py39 support and bump min package versions

rcomer:
## Rationale

According to the [NEP29 Support Table](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) we should now be supporting Python 3.10+ and Numpy 1.23+.  [SPEC0](https://scientific-python.org/specs/spec-0000/) says we should support Matplotlib 3.6+ and Scipy 1.9+.  For other packages I tried to go for something that is a little older than 2 years.  This is more lenient than for the core packages considered by SPEC0, but some of them have pretty erratic release frequencies and it just seemed simpler to go for 2 years+ than try to decide how much less than 2 years is reasonable.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,21 +8,21 @@ name: cartopy-dev
 channels:
   - conda-forge
 dependencies:
-  - cython>=0.29.24
-  - numpy>=1.21
-  - shapely>=1.7.1
+  - cython>=0.29.28
+  - numpy>=1.23
+  - shapely>=1.8
   - pyshp>=2.3
   - pyproj>=3.3.1
+  - packaging>=21
   # The testing label has the proper version of freetype included
-  - conda-forge/label/testing::matplotlib-base>=3.5
+  - conda-forge/label/testing::matplotlib-base>=3.6
 
   # OWS
-  - owslib>=0.24.1
-  - pillow>=6.1.0
+  - owslib>=0.27
+  - pillow>=9.1
   # Plotting
-  - scipy>=1.6.3
+  - scipy>=1.9
   # Testing
-  - packaging>=20
   - pytest
   - pytest-mpl
   - pytest-xdist
--- a/lib/cartopy/tests/mpl/test_crs.py
+++ b/lib/cartopy/tests/mpl/test_crs.py
@@ -7,12 +7,12 @@ import matplotlib.pyplot as plt
 import pytest
 
 import cartopy.crs as ccrs
-from cartopy.mpl import _MPL_36
+from cartopy.mpl import _MPL_37
 
 
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(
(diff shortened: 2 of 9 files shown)


rcomer:
OWSlib 0.26 doesn't play with PyProj 3.3.  So bumped OWSlib further to 0.27. https://github.com/geopython/OWSLib/issues/823

lib/cartopy/tests/mpl/test_crs.py
```diff
@@ -7,12 +7,12 @@
...
 
 import cartopy.crs as ccrs
-from cartopy.mpl import _MPL_36
+from cartopy.mpl import _MPL_37
 
 
 @pytest.mark.natural_earth
 @pytest.mark.mpl_image_compare(
-    filename="igh_land.png", tolerance=0.5 if _MPL_36 else 3.6)
+    filename="igh_land.png", tolerance=0.5 if _MPL_37 else 3.6)
```
rcomer, line 15: We did this [last time too](https://github.com/SciTools/cartopy/pull/2258/files#diff-18c2d8f7b2271b656b2822b5984e1e1c2ee1d768a243c1b47136ee435f98b583), based on failures in the min version test.  I'm starting to think maybe it's not the mpl version making the difference.  I don't think I want to start digging into that today though.

environment.yml
```diff
@@ -8,21 +8,21 @@ name: cartopy-dev
...
 dependencies:
-  - cython>=0.29.24
-  - numpy>=1.21
-  - shapely>=1.7.1
+  - cython>=0.29.28
+  - numpy>=1.23
+  - shapely>=1.8
   - pyshp>=2.3
   - pyproj>=3.3.1
+  - packaging>=20
```
lgolston, line 16: >=21 to be consistent with pyproject.toml

lgolston:
Can also bump ruff target setting to py310 at: https://github.com/rcomer/cartopy/blob/14beb1bb64aac3b1b1a2852cd9918961e9179ca1/pyproject.toml#L101

lgolston:
Maybe drop py39 too from CIBW_BUILD at https://github.com/SciTools/cartopy/blob/14beb1bb64aac3b1b1a2852cd9918961e9179ca1/.github/workflows/release.yml#L93

rcomer:
Thanks for picking those up @lgolston!  Now fixed.

dopplershift:
> Can also bump ruff target setting to py310 at: https://github.com/rcomer/cartopy/blob/14beb1bb64aac3b1b1a2852cd9918961e9179ca1/pyproject.toml#L101

I think we can actually do even better and just remove `target-version` altogether. The [default config](https://docs.astral.sh/ruff/settings/#target-version) will look at the `python-requires` set more broadly in `pyproject.toml`.

lgolston (approved):

PR #2386, opened 2024-05-17
Removed dict annotation from GeoAxes.gridlines

BhavyeMathur:
## Rationale

Just something that had been bothering me for a bit. The `**kwargs` parameter in `GeoAxes.gridlines` is itself a `dict`, but shouldn't be annotated as `dict` because that implies that it accepts only dictionary keyword arguments. For example, the following code is highlighted with a warning:

```python
ax.gridlines(..., linewidth=0.15)  # syntax highlighter expects linewidth to be a dictionary
```

## Implications

Better syntax highlighting. Warning is no longer displayed for code that is correct.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -1473,7 +1473,7 @@ class GeoAxes(matplotlib.axes.Axes):
 
         Keyword Parameters
         ------------------
-        **kwargs: dict
+        **kwargs:
             All other keywords control line properties.  These are passed
             through to :class:`matplotlib.collections.Collection`.
 


BhavyeMathur:
Some of the tests have failed due to 404 errors. I don't think those are related to my changes however.

rcomer (approved):
Hi @BhavyeMathur, thanks for this.

Cartopy follows the numpydoc style and according to the [guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters), including the `dict` type is correct.  However, I also looked up [a](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) [few](https://numpy.org/doc/stable/reference/generated/numpy.arccos.html) [examples](https://numpy.org/doc/stable/reference/generated/numpy.bitwise_xor.html) [in](https://numpy.org/doc/stable/reference/generated/numpy.deg2rad.html) [numpy](https://numpy.org/doc/stable/reference/generated/numpy.greater.html) [itself](https://numpy.org/doc/stable/reference/generated/numpy.isnan.html), and only one of them show a type for `**kwargs`, which is `any`.

I cannot think of a scenario where stating this is a `dict` is useful information for the user, so if it causes a problem, let's take it out.  Out of interest which editor do you use?

Although the doc build is broken, the [relevant docstring renders fine](https://output.circle-artifacts.com/output/job/9fa2f125-9f1f-4e95-88b3-6a36fde56b16/artifacts/0/docs/build/html/reference/generated/cartopy.mpl.geoaxes.GeoAxes.html#cartopy.mpl.geoaxes.GeoAxes.gridlines).

BhavyeMathur:
Hi, thanks for merging the pull request. 

Interesting, I wonder why NumPy recommends those hints for `*args` and `**kwargs`. Noticed the issue while using JetBrains PyCharm and DataSpell, but looks good now.

PR #2330, opened 2024-02-22
FIX: gridliner auto-updates by default (and always in future)

rcomer:
## Rationale

As suggested at https://github.com/SciTools/cartopy/issues/1792#issuecomment-1954609875, updating gridlines and their labels when the figure is re-drawn (or tight_bbox re-calculated) is now the default.  Also deprecate the option to not update.

Closes #1792 

## Implications

There is one change in an image test result, where the zero labels on the equator have disappeared from the top right plot.  I'm not 100% clear why this is.  There is some logic that makes labels invisible based on how much space is available so my best guess is that the calculation after `tight_layout` puts us just the other side of the threshold for "not enough space", though any changes in positioning are imperceptible.

I guess if anyone tells us they have a use-case where `auto_update=False` matters, we can remove the deprecation.

PR #2387, opened 2024-05-19
Make Gridliner deprecation warning show up

rcomer:
## Rationale

Follow up to #2330.  The deprecation warning is currently not showing for normal use.  I thought I had checked but perhaps I only looked in test output.  I've now set the stacklevel to 3 for when `Gridliner` is created via `ax.gridlines`, and to 2 for the possible (but unlikely?) cases where a `Gridliner` is created directly.

I have checked locally that this now gives the desired results
```python
In [1]: import matplotlib.pyplot as plt
   ...: import cartopy.crs as ccrs
   ...: fig = plt.figure()
   ...: ax = fig.add_subplot(projection=ccrs.PlateCarree())
   ...: ax.coastlines()
Out[1]: <cartopy.mpl.feature_artist.FeatureArtist at 0x756842d2d1c0>

In [2]: ax.gridlines(draw_labels=True, auto_update=True)
<ipython-input-2-499091cee3ba>:1: DeprecationWarning: The auto_update parameter was deprecated at Cartopy 0.23.  In future the gridlines and labels will always be updated.
  ax.gridlines(draw_labels=True, auto_update=True)
Out[2]: <cartopy.mpl.gridliner.Gridliner at 0x75688bcc5790>

In [3]: from cartopy.mpl.gridliner import Gridliner

In [4]: test = Gridliner(ax, ccrs.PlateCarree(), auto_update=True)
<ipython-input-4-422209944b61>:1: DeprecationWarning: The auto_update parameter was deprecated at Cartopy 0.23.  In future the gridlines and labels will always be updated.
  test = Gridliner(ax, ccrs.PlateCarree(), auto_update=True)
```

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -448,6 +448,7 @@ class Gridliner(matplotlib.artist.Artist):
         if auto_update is None:
             auto_update = True
         else:
+            # Note #2394 should be addressed before this deprecation expires.
             calling_module = inspect.stack()[1].filename
             warnings.warn(
                 "The auto_update parameter was deprecated at Cartopy 0.23.  In future "
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -9,6 +9,7 @@ location information to the plots.
 
 """
 
+import inspect
 import itertools
 import operator
 import warnings
@@ -447,10 +448,13 @@ class Gridliner(matplotlib.artist.Artist):
         if auto_update is None:
             auto_update = True
         else:
+            calling_module = inspect.stack()[1].filename
             warnings.warn(
                 "The auto_update parameter was deprecated at Cartopy 0.23.  In future "
                 "the gridlines and labels will always be updated.",
-                DeprecationWarning)
+                DeprecationWarning,
+                stacklevel=(3 if calling_module.endswith('cartopy/mpl/geoaxes.py')
+                            else 2))
         self._auto_update = auto_update
 
     def has_labels(self):


rcomer:
#2394 suggests we have a bad interaction between the label placement logic and `bbox_inches='tight'`.  I have added a comment that that should be addressed before we expire the deprecation.

lib/cartopy/mpl/gridliner.py
```diff
@@ -447,10 +448,13 @@ def __init__(self, axes, crs, draw_labels=False, xlocator=None,
         if auto_update is None:
             auto_update = True
         else:
+            # Note #2394 should be addressed before this deprecation expires.
             warnings.warn(
                 "The auto_update parameter was deprecated at Cartopy 0.23.  In future "
                 "the gridlines and labels will always be updated.",
-                DeprecationWarning)
+                DeprecationWarning,
+                stacklevel=(3 if sys._getframemodulename(1) == 'cartopy.mpl.geoaxes'
```
greglucas, line 456: I'm not sure we should be using the private method here. The documentation says it may not be available in all reference Python implementations.
https://docs.python.org/3/library/sys.html#sys._getframemodulename

I think you could possibly use the inspect module instead, but it looks like in other code locations we've just explicitly set the stacklevel to 2, is that enough to handle this and not special case different stack levels?

rcomer: Thanks - I didn't know `inspect` had something for that.  Now updated.  If we leave the stacklevel at 2 then the warning won't show for users who are calling `ax.gridlines`.  I was semi-following Matplotlib's `warn_external` which [uses sys._getframe](https://github.com/matplotlib/matplotlib/blob/b02aa82f3d3ba7f55e8a3fd431bc8763e3ae5c5c/lib/matplotlib/_api/__init__.py#L369).

greglucas (approved):

PR #2392, opened 2024-05-24
Use fewer network tests

lgolston:
## Rationale

This builds further on the approach from @rcomer in #2371. Now the tests involving network access are only run twice (once on minimum dependencies, and once on latest dependencies. It should be noted, the documentation/gallery also will pick up on some network related issues).
This also removes a shapely-dev remnant.

## Implications

Network and non-network tests are fully separated, making it easier to track problems that come up.

External network usage is reduced since the tests are only run on a subset of the full testing matrix.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -18,11 +18,14 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12']
-        use-network: [true]
+        use-network: [false]
         include:
           - os: ubuntu-latest
-            python-version: '3.11'
-            use-network: false
+            python-version: '3.10'
+            use-network: true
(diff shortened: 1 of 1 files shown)


lgolston:
Not sure what has changed, to cause problem with line continuation on Windows - will check tomorrow.

rcomer:
A lot of the tests that use natural earth are not actually testing the natural earth functionality but just happen to add coastlines or similar when testing something else.  Arguably that wasn’t a great choice when writing those tests, but I wonder if we want to restrict the test coverage that much.

greglucas:
I think having a go at rebasing/updating https://github.com/SciTools/cartopy/pull/1250 to try and save the network activity would be useful as an alternative to reducing the number of tests we are running in CI.

lgolston:
The VCR alternative, with occasional live network checks, seems reasonable so I will go ahead and close this out.

PR #2401, opened 2024-06-20
Fix failing gouraud shading mask creation in pccolormesh

e-k-m:
## Rationale

Using pccolormesh with gouraud shading fails on wrap around mask creation, since numpy array mask is created with negative dimension. 

Details: Fails here https://github.com/SciTools/cartopy/blob/main/lib/cartopy/mpl/geoaxes.py#L1896 because 
data_shape equals (Nx, Ny, -1) or (Ny - 1, Nx - 1, -1). 

## Implications

As far as I can see there are no implications of the proposed change.

### DIFF
--- a/lib/cartopy/tests/mpl/test_mpl_integration.py
+++ b/lib/cartopy/tests/mpl/test_mpl_integration.py
@@ -654,6 +654,17 @@ def test_pcolormesh_single_column_wrap():
     return fig
 
 
+def test_pcolormesh_wrap_gouraud_shading_failing_mask_creation():
+    x_range = np.linspace(-180, 180, 50)
+    y_range = np.linspace(90, -90, 50)
+    x, y = np.meshgrid(x_range, y_range)
+    data = ((np.sin(np.deg2rad(x))) / 10. + np.exp(np.cos(np.deg2rad(y))))
+
+    fig = plt.figure(figsize=(10, 6))
+    ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mercator())
+    ax.pcolormesh(x, y, data, transform=ccrs.PlateCarree(), shading='gouraud')
+
+
 def test_pcolormesh_diagonal_wrap():
     # Check that a cell with the top edge on one side of the domain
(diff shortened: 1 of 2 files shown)


e-k-m:
Docs generation seems to fail given an unrelated issue to this PR. Ready for review. THX.

lib/cartopy/crs.py
greglucas, line 1: Could you revert the changes to this file? They are unrelated to your other updates.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -654,6 +654,21 @@ def test_pcolormesh_single_column_wrap():
...
+    x, y = np.meshgrid(x_range, y_range)
+    data = ((np.sin(np.deg2rad(x))) / 10. + np.exp(np.cos(np.deg2rad(y))))
+
+    fig = plt.figure(figsize=(10, 6))
+    ax = fig.add_subplot(1, 1, 1, projection=ccrs.Mercator())
+    try:
+        ax.pcolormesh(x, y, data, transform=ccrs.PlateCarree(), shading='gouraud')
+    except ValueError as e:
+        if str(e) == 'negative dimensions are not allowed':
+            pytest.fail("Unexpected ValueError('negative dimensions are not allowed')")
```
greglucas, line 669: pytest will fail for us, so no need to explicitly catch and fail.
```suggestion
    ax.pcolormesh(x, y, data, transform=ccrs.PlateCarree(), shading='gouraud')
```

greglucas (commented):
Thanks for updating this! Looks like that was just overlooked in https://github.com/SciTools/cartopy/pull/2166

lib/cartopy/crs.py
e-k-m: Yes, done.

lib/cartopy/tests/mpl/test_mpl_integration.py
e-k-m: Yes, done. THX.

greglucas (approved):

PR #2402, opened 2024-06-22
CI: Fix image upload path to remove shapely-dev

greglucas:
There was an extra "-" added at the end of our CI uploads that prevented artifacts from being uploaded. We don't use shapely-dev anymore, so that can be removed here.


rcomer (approved):

rcomer:
CI failures are clearly unrelated.

PR #2404, opened 2024-06-22
DOC: Remove expected NASA failing images because they work now

greglucas:
In the title, try to get circleCI builds working again. Ii chose to leave the comment in there rather than removing it altogether because they are bound to fail again and it is easier to just comment/uncomment those lines of code.


rcomer (approved):

rcomer:
Woo!  `main` is green again.  Thanks for sorting those out @greglucas

PR #2407, opened 2024-06-29
FIX/MNT: Simplify project geometry handling

### LINKED ISSUE #2406: Projection._project_multiline() returns empty list rather than empty sgeom.MultiLineString
### Description

When providing a MultiLineString to the [crs.Projection.project_geometry()](https://scitools.org.uk/cartopy/docs/latest/_modules/cartopy/crs.html#Projection) function, if an empty MultiLineString is provided or if the projection returns no geometries, an empty list is returned, rather than an empty MultiLineString. In all other _project_* functions, an empty geometry is returned if the geoms list is empty.

This is a problem for me as I am projecting global LineStrings/MultiLineStrings from Plate Carree to Mercator, and some of these objects may be north/south of the limits of the Mercator projection, and so return nothing when projected. This leads to crs.Projection.project_geometry() returning an empty list rather than an empty MultiLineString, which caused my downstream code to fail as it expected the returned object to have  a `geoms` attribute. I can bypass this issue with an `isinstance(geom, list)` condition, but the underlying issue should probably be fixed.

#### Code to reproduce

```
>>> import shapely.geometry as sgeom
>>> ccrs.Mercator().project_geometry(sgeom.MultiLineString([sgeom.LineString([[0, 0], [1, 1]])]), ccrs.PlateCarree())
<MULTILINESTRING ((0 0, 111319.491 110579.965))>
>>> type(ccrs.Mercator().project_geometry(sgeom.MultiLineString([sgeom.LineString([[0, 0], [1, 1]])]), ccrs.PlateCarree()))
<class 'shapely.geometry.multilinestring.MultiLineString'>
>>> ccrs.Mercator().project_geometry(sgeom.MultiLineString(), ccrs.PlateCarree())
[]
>>> type(ccrs.Mercator().project_geometry(sgeom.MultiLineString(), ccrs.PlateCarree()))
<class 'list'>
>>> ccrs.Mercator().project_geometry(sgeom.MultiLineString([sgeom.LineString([[0, 88], [1, 88]])]), ccrs.PlateCarree()) 
[]
>>> type(ccrs.Mercator().project_geometry(sgeom.MultiLineString([sgeom.LineString([[0, 88], [1, 88]])]), ccrs.PlateCarree()))
<class 'list'>
```

greglucas:
The MultiLineString return type was a plain list, but it should be an empty MultiLineString to be consistent with the other return types. Additionally, all geometry constructors take empty lists, so just use that rather than special-casing the returns.

Closes #2406

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -918,10 +918,7 @@ class Projection(CRS, metaclass=ABCMeta):
         geoms = []
         for geom in geometry.geoms:
             geoms.append(self._project_point(geom, src_crs))
-        if geoms:
-            return sgeom.MultiPoint(geoms)
-        else:
-            return sgeom.MultiPoint()
+        return sgeom.MultiPoint(geoms)
 
     def _project_multiline(self, geometry, src_crs):
         geoms = []
@@ -929,10 +926,7 @@ class Projection(CRS, metaclass=ABCMeta):
             r = self._project_line_string(geom, src_crs)
             if r:
                 geoms.extend(r.geoms)
-        if geoms:
-            return sgeom.MultiLineString(geoms)
-        else:
-            return []
+        return sgeom.MultiLineString(geoms)
 
     def _project_multipolygon(self, geometry, src_crs):
         geoms = []
@@ -940,11 +934,7 @@ class Projection(CRS, metaclass=ABCMeta):
             r = self._project_polygon(geom, src_crs)
             if r:
                 geoms.extend(r.geoms)
-        if geoms:
-            result = sgeom.MultiPolygon(geoms)
-        else:
-            result = sgeom.MultiPolygon()
-        return result
+        return sgeom.MultiPolygon(geoms)
 
     def _project_polygon(self, polygon, src_crs):
         """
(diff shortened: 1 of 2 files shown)


greglucas:
Interesting, the `geom_type` has changed from `GeometryCollection` to `MultiLineString`. So previously, this would have failed before even making it into the transformation because we don't know how to project GeometryCollections. Should we ignore this test with `shapely<2`, or other suggestions for how to handle it?

dopplershift:
I think it's important to know whether any of those classes accepted empty lists on Shapely < 2? It's not clear to me whether the test is checking code paths that we could even hit before in practice.

Essentially, after this change: do we actually still support Shapely < 2?

greglucas:
Yes, we still have a minimum test that pulls in Shapely 1.8.

https://github.com/SciTools/cartopy/actions/runs/9726193384/job/26844355302#step:4:1

Shapely 2.0 changes the geom_type return to be more correct. Not a regression for us, just additional things that will get projected properly now that we get for free. It just impacts the test where this fails without shapely 2.0 and was also failing before this update.

dopplershift (approved):
This looks good to me then.

PR #2426, opened 2024-08-23
Support shading=None in pcolor(mesh)

rcomer:
## Rationale

If `pcolor`/`pcolormesh` receives `shading=None`, that should be treated the same as if _shading_ was not set.  My colleague was using `pyplot.pcolormesh` on a geoaxes, and pyplot [explicitly defaults the shading to `None`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pcolormesh.html#matplotlib.pyplot.pcolormesh).  This caused a hard-to-debug artefact!


rcomer:
Thanks @dopplershift!

PR #2441, opened 2024-09-24
FIX: `_trigger_patch_reclip` takes axes instance

rcomer:
## Rationale

According to the [Axes documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.html), the callbacks attached to an axes "will be called with func(ax) where ax is the `Axes` instance."  So changed the `_trigger_patch_reclip` function to expect the axes rather than the event.  This fix was prompted by [this StackOverflow question](https://stackoverflow.com/questions/79018815/keyerror-geo-when-attempting-to-add-geoaxes-using-axes-inset-axes) and the new test adapted from that.  Note that [the axes property of a normal axes is itself, but the axes property of an inset axes is its parent](https://github.com/matplotlib/matplotlib/blob/613c9c996c20e034de67126e3e1cc42d19835b1f/lib/matplotlib/axes/_base.py#L2277-L2279).  Unfortunately the test coverage of this callback is minimal: when I removed it completely, the only test failure I got was [here](https://github.com/SciTools/cartopy/blob/d57fa1328c701d40edffbd5882a12fb191c5ccf9/lib/cartopy/tests/mpl/test_gridliner.py#L598), where `labels` became an empty list.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -2302,13 +2302,12 @@ GeoAxesSubplot = matplotlib.axes.subplot_class_factory(GeoAxes)
 GeoAxesSubplot.__module__ = GeoAxes.__module__
 
 
-def _trigger_patch_reclip(event):
+def _trigger_patch_reclip(axes):
     """
     Define an event callback for a GeoAxes which forces the background patch to
     be re-clipped next time it is drawn.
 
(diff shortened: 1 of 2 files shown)


QuLogic (approved):
Seems right to me.

dopplershift (approved):

PR #1561, opened 2020-05-15
Implemented IGH Oceanic View

jkrasting:
- Option to select 'land' or 'ocean' emphasis
- Requires an [update to the PROJ library](https://github.com/OSGeo/PROJ/pull/2226) and use of
  a new projection (imh_o)
- Ocean view works best with central_longitude=-160.

IGH Default View:

```
proj = ccrs.InterruptedGoodeHomolosine()
```

imh_default_view

IGH New Optional Ocean View:

```
proj = ccrs.InterruptedGoodeHomolosine(central_longitude=-160,emphasis='ocean')
```

imh_ocean_view

## Rationale

This update provides new functionality for oceanographic plots that was not previously possible with the land-only focus of the current IGH projection.

## Implications

None.

PR #2442, opened 2024-09-25
Added winkel tripel projection class

AgentOxygen:
## Rationale
Winkel-Tripel is a popular projection that would be used by multiple users (my group at UT and folks at NCAR). It is also the primary projection for the National Geographic Society.

## Implications
I have created a new CRS projection object without modifying any code by utilizing existing class objects and the pyproj backend.

## Checklist
1. The new class object can be utilized in similar fashion to the other projection classes. 
2. I am unsure how to integrate this into the existing test suite, however I imagine it would undergo the same testing as other projections. I am happy to assist with that integration if  I could be pointed in the right direction.

This is my first pull request! I am open to input how on to properly contribute.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -2493,6 +2493,33 @@ class Robinson(_WarpedRectangularProjection):
         return result
 
 
+class WinkelTripel(_WarpedRectangularProjection):
+    """
+    A Winkel-Tripel projection.
+
+    Compromise modified azimuthal projection that is less distorted
+    and more area-accurate. It is comparable to the Robinson
+    projection in both appearance and popularity.
+
+    The National Geographic Society uses the Winkel-Tripel projection
+    for most of the maps they produce.
+    """
+
+    def __init__(self, central_longitude=0.0, globe=None):
+        globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
+        proj4_params = [('proj', 'wintri'),
+                        ('lon_0', central_longitude),
+                        ('lat_0', 0.0)]
+
+        super(WinkelTripel, self).__init__(proj4_params,
+                                           central_longitude,
+                                           globe=globe)
+
+    @property
+    def threshold(self):
+        return 1e4
+
+
 class InterruptedGoodeHomolosine(Projection):
     """
     Composite equal-area projection emphasizing either land or


lib/cartopy/crs.py
```diff
@@ -2493,6 +2493,33 @@ def transform_points(self, src_crs, x, y, z=None, trap=False):
...
+    for most of the maps they produce.
+    """
+
+    def __init__(self, central_longitude=0.0, globe=None):
+        globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
+        proj4_params = [('proj', 'wintri'),
+                        ('lon_0', central_longitude),
+                        ('lat_0', 0.0)]
+
+        super(WinkelTripel, self).__init__(proj4_params,
```
greglucas, line 2514:
```suggestion
        super().__init__(proj4_params,
```

lib/cartopy/crs.py
```diff
@@ -2493,6 +2493,33 @@ def transform_points(self, src_crs, x, y, z=None, trap=False):
...
+
+    The National Geographic Society uses the Winkel-Tripel projection
+    for most of the maps they produce.
+    """
+
+    def __init__(self, central_longitude=0.0, globe=None):
+        globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
+        proj4_params = [('proj', 'wintri'),
+                        ('lon_0', central_longitude),
+                        ('lat_0', 0.0)]
```
greglucas, line 2512: Do we need to pass in the `lat_0` parameter even if 0 is the default, and does that even work, it looks like it is `lat_1` in the documentation: https://proj.org/en/9.4/operations/projections/wintri.html

lib/cartopy/crs.py
```diff
@@ -2493,6 +2493,33 @@ def transform_points(self, src_crs, x, y, z=None, trap=False):
         return result
 
 
+class WinkelTripel(_WarpedRectangularProjection):
```
greglucas, line 2496: Do we need to do any of the nan handling here like we are doing in the Robinson projection? i.e. should this subclass `Robinson` instead if it is similar to that?

greglucas (commented):
Nice, this looks like a great start.

Regarding the tests: Probably add a test similar to one of the other crs's here containing some transform points and initialization expectations: https://github.com/SciTools/cartopy/tree/main/lib/cartopy/tests/crs

Add the projection to the parameterization list here and add a comparison image.
https://github.com/SciTools/cartopy/blob/d57fa1328c701d40edffbd5882a12fb191c5ccf9/lib/cartopy/tests/mpl/test_mpl_integration.py#L161-L200

A PR you can look at for updating one of the other projections is https://github.com/SciTools/cartopy/pull/1561

lib/cartopy/crs.py
AgentOxygen: I was able to generate maps containing maps with no problem. Are there other tests for nans that I would need to do?

greglucas:
ping @AgentOxygen did you want to try and get this in still? It looks like it needs tests and a possible update to the keyword argument.

lib/cartopy/crs.py
AgentOxygen: `lat_0` is required and `lat_1` doesn't work. Maybe this is a typo in proj? I'm also not very familiar with that library so maybe I'm hacking a parameter.

AgentOxygen:
Apologies for the delay, I just finished up my master's thesis! I've added the test and an image for comparison. I noticed several of the tests failed, though I'm not sure how to proceed. What are the next steps I should take to get this integrated.

greglucas:
> Apologies for the delay, I just finished up my master's thesis!

🎉 Congratulations, @AgentOxygen!

lib/cartopy/crs.py
```diff
@@ -2539,6 +2539,33 @@ def transform_points(self, src_crs, x, y, z=None, trap=False):
...
+
+    The National Geographic Society uses the Winkel-Tripel projection
+    for most of the maps they produce.
+    """
+
+    def __init__(self, central_longitude=0.0, globe=None):
+        globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
+        proj4_params = [('proj', 'wintri'),
+                        ('lon_0', central_longitude),
+                        ('lat_0', 0.0)]
```
greglucas, line 2558:
```suggestion
                        ('lon_0', central_longitude)]
```

lib/cartopy/crs.py
```diff
@@ -2539,6 +2539,33 @@ def transform_points(self, src_crs, x, y, z=None, trap=False):
...
+
+        super().__init__(proj4_params,
+                                           central_longitude,
+                                           globe=globe)
```
greglucas, line 2562:
```suggestion
        super().__init__(proj4_params, central_longitude, globe=globe)
```

greglucas:
@AgentOxygen, it looks like the image comparison is failing now. I pushed the formatting fix, but otherwise this looks good to go to me once you can update the images to the expected content.

PR #2450, opened 2024-10-07
c: Fix pin for cibuildwheel when building matrix

QuLogic:
## Rationale

We used to install cibuildwheel manually in order to generate the Python-version matrix. This means that dependabot does not notice or update it. This is normally fine, but when cibuildwheel adds new Python versions, they won't be in the generated matrix until the pin is updated.

Instead of the manual install, use the action to install cibuildwheel, but don't have it run anything.

Also, expand the `jq` arguments so it's a bit clearer what it does.

## Implications

I'm not 100% certain this implementation will work, but it should add Python 3.13 wheels to the matrix.


QuLogic:
I see I haven't contributed in a long while considering I signed the CLA when it was v4.

greglucas (approved):
:( Sorry, I could have sworn I checked 3.13 wheels earlier but I must have been looking at the test matrix instead. Thanks for the really quick investigation and fix here!

I wonder if this is more fancy than necessary and we should just do the serial builds on each OS like most other projects...

Looks like all tests and builds are green. Can you merge and make a new release when you're ready?

QuLogic:
Yes, this does do a bit of a weird thing in that it assumes the location that cibuildwheel gets installed. Builds take about 0.5-2 minutes (unlike, e.g., Matplotlib's 10-60 for all versions combined in its matrix), so maybe it's just easier to simplify to per-arch/platform?

QuLogic:
Do we want to enable free-threading wheels as well? Or possibly PyPy too?

greglucas:
> Do we want to enable free-threading wheels as well? Or possibly PyPy too?

* free-threading, I tried building locally and I don't think we can quite yet because Shapely and Pyproj both need to be built from source for it at the moment, so probably not worth doing until they have wheels we can build with (we use them within the Cython code, so need it for building our extension).
* PyPy, I don't use personally, so if others would find it useful then sure 🤷 , but I would wait until we get a request for it before spending much time getting that to work.

All in all, I'd suggest just going with your minimal change here. I like the simplification of the build too, thanks for consolidating that!

QuLogic:
> Looks like all tests and builds are green. Can you merge and make a new release when you're ready?

I'm just running tests on Fedora now, so won't release yet, just in case there's some followup there.

PR #2452, opened 2024-10-11
MNT: Deprecate unused clip_path module

greglucas:
None of these functions are used anywhere within Cartopy and they can easily be implemented by others downstream if needed.


QuLogic (approved):
I also could not find any mention of `clip_path` usage.

rcomer (approved):

PR #2453, opened 2024-10-11
MNT: Switch to (lon, lat) for all argument and return ordering

greglucas:
There are more usages of (lon, lat) currently and only a few private versions of (lat, lon) so it is easy enough to switch everything over to (lon, lat) for consistency. This matches what we are expecting from transform_point(x, y) as well.

### DIFF
--- a/lib/cartopy/tests/test_crs.py
+++ b/lib/cartopy/tests/test_crs.py
@@ -36,7 +36,7 @@ class TestCRS:
         ll = ccrs.Geodetic()
 
         # results obtained by nearby.org.uk.
-        lat, lon = np.array([54.5622169298669, -5.54159863617957],
+        lon, lat = np.array([-5.54159863617957, 54.5622169298669],
                             dtype=np.double)
         east, north = np.array([359000, 371000], dtype=np.double)
 
@@ -65,7 +65,7 @@ class TestCRS:
         ll = ccrs.Geodetic()
 
         # results obtained by streetmap.co.uk.
-        lat, lon = np.array([50.462023, -3.478831], dtype=np.double)
+        lon, lat = np.array([-3.478831, 50.462023], dtype=np.double)
         east, north = np.array([295132.1, 63512.6], dtype=np.double)
 
         # note the handling of precision here...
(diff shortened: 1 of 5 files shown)


lib/cartopy/tests/test_crs.py
```diff
@@ -36,7 +36,7 @@ def test_osni(self, approx):
         ll = ccrs.Geodetic()
 
         # results obtained by nearby.org.uk.
-        lat, lon = np.array([54.5622169298669, -5.54159863617957],
+        lon, lat = np.array([-5.54159863617957, 54.5622169298669],
```
QuLogic, line 39: Why is this even put through `np.array` in the first place if it's split immediately?

greglucas: 🤷 I was wondering the same thing, but really didn't want to go refactoring more than this PR title suggests.

QuLogic (approved):

dopplershift (approved):
Curious what the failure cause is?

greglucas:
> Curious what the failure cause is?

Looks like an issue with a corrupted shapefile download or something that Fiona is causing errors from which leads to a later error.

```
type = <class 'weakref.KeyedRef'>, ob = None
callback = <function WeakValueDictionary.__init__.<locals>.remove at 0x000001D0FE0E63E0>
key = <cartopy.mpl.feature_artist._GeomKey object at 0x000001D0FFA4D590>

    def __new__(type, ob, callback, key):
>       self = ref.__new__(type, ob, callback)
E       TypeError: cannot create weak reference to 'NoneType' object

C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\weakref.py:348: TypeError
------------------------------ Captured log call ------------------------------
ERROR    fiona._env:shapereader.py:217 Error in fread() reading object of size 248 at offset 332 from .shp file
ERROR    fiona._env:shapereader.py:217 Error in fread() reading object of size 1128 at offset 580 from .shp file
ERROR    fiona._env:shapereader.py:217 Error in fread() reading object of size 792 at offset [170](https://github.com/SciTools/cartopy/actions/runs/11298952884/job/31428955719?pr=2453#step:9:171)8 from .shp file
ERROR    fiona._env:shapereader.py:217 fread(27) failed on DBF file.
```

I reran the job as this seems like something intermittent and unrelated.

PR #2455, opened 2024-10-15
Introduce `path_to_shapely` and `shapely_to_path`

### LINKED ISSUE #2447: Make `geos_to_path` return one path per geometry
Should this logic be moved up into the `geos_to_path()` function though? That currently returns a list of paths, but it seems like if you're passing in one geometry that should be a compound path like you have here.

_Originally posted by @greglucas in https://github.com/SciTools/cartopy/pull/2325#discussion_r1484944534_
            
I have been meaning to look into this but not got around to it, so creating an issue before I forget completely!

rcomer:
## Rationale

Closes #2447.  Introduce `shapely_to_path` which maps a single Shapely geometry or collection to a single Matplotlib path, and `path_to_shapely`, which does the opposite.  To take full advantage of this in our transforms code, also modify `project_geometry` so that it returns a `GeometryCollection` from either a `GeometryCollection` or a `LinearRing`.  Previously `GeometryCollection` was not supported and `LinearRing` was mapped to a tuple.

Note the first commit just duplicates the existing `path_to_geos` and `geos_to_path` and all the material changes are in the second commit.  I rearranged the logic in `path_to_shapely` because it seemed simpler to separate the geometry types into different lists up front.

I also went ahead and started a whatsnew for the next release.  I think that was suggested a couple of releases ago and it does seem simpler to me to add the entries as we go.

## Implications

The new behaviour of `project_geometry` from a `LinearRing` could be a breaking change but, since the docstring of `project_geometry` states that it returns a Shapely geometry, I claim this one as a bug fix.


lib/cartopy/mpl/patch.py
```diff
@@ -302,79 +322,83 @@ def path_to_shapely(path, force_ccw=False):
...
         else:
-            geom = sgeom.LineString(path_verts)
-
-        # If geom is a Polygon and is contained within the last geom in
-        # collection, it usually needs to be an interior to that geom (e.g. a
-        # lake within a land mass).  Sometimes there is a further geom within
-        # this interior (e.g. an island in a lake, or some instances of
-        # contours).  This needs to be a new external geom in the collection.
-        if geom.is_empty:
-            pass
```
rcomer: These lines were introduced at #560, but the test added then has since been updated to expect [a list of points](https://github.com/SciTools/cartopy/blob/f8fee7ff052e6ca7672b4e85edc15014009a08d1/lib/cartopy/tests/mpl/test_patch.py#L26).  So I don't think this loop is relevant any more.

lib/cartopy/tests/mpl/test_caching.py
```diff
@@ -122,7 +122,8 @@ def test_contourf_transform_path_counting():
     gc.collect()
     initial_cache_size = len(cgeoaxes._PATH_TRANSFORM_CACHE)
 
-    with mock.patch('cartopy.mpl.patch.path_to_geos') as path_to_geos_counter:
+    with mock.patch('cartopy.mpl.patch.path_to_shapely',
+                    wraps=cartopy.mpl.patch.path_to_shapely) as path_to_shapely_counter:
```
rcomer, line 126: This was failing because `project_geometry` received a `MagicMock` that it didn't know what to do with.  I'm confused why that didn't also happen previously with `path_to_geos`.

rcomer:
I also tested this with my user code that generates many contour plots.  I got identical results with my branch and with `main`.

rcomer:
The Ubuntu py3.10 failures look relevant but I have no idea why they would specifically affect that OS and python version.

rcomer:
Most of the tests were fixed by bumping minimum Shapely to v2.0.  I think that’s reasonable to do since v2 will be 2 years old in December.  However, I think the remaining test failure might need a workaround for lack of https://github.com/matplotlib/matplotlib/pull/25252.  It’s too soon to drop support for MPL 3.7.

docs/source/whatsnew/v0.25.rst
```diff
@@ -0,0 +1,19 @@
+Version 0.25 (Date TBD)
+=======================
+
```
greglucas, line 3: Add a note about the minimum version of Shapely being bumped

lib/cartopy/crs.py
```diff
@@ -821,6 +821,10 @@ def project_geometry(self, geometry, src_crs=None):
             raise TypeError('Source CRS must be an instance of CRS'
                             ' or one of its subclasses, or None.')
         geom_type = geometry.geom_type
+        if geom_type == 'GeometryCollection':
+            return sgeom.GeometryCollection(
+                [self.project_geometry(geom, src_crs) for geom in geometry.geoms])
```
greglucas, line 826: Perhaps add this to the `_method_map` and add the logic to a new function `_project_geometry_collection` to keep it consistent with the others.

lib/cartopy/crs.py
```diff
@@ -915,7 +919,7 @@ def _project_linear_ring(self, linear_ring, src_crs):
         if rings:
             multi_line_string = sgeom.MultiLineString(line_strings)
 
-        return rings, multi_line_string
+        return sgeom.GeometryCollection([*rings, multi_line_string])
```
greglucas, line 922: Update the docstring of this function to indicate that it is returning a GeometryCollection now.

lib/cartopy/mpl/patch.py
```diff
@@ -53,13 +57,16 @@ def geos_to_path(shape):
...
+       Use `shapely_to_path` instead.
+
     Parameters
     ----------
     shape
         A list, tuple or single instance of any of the following
         types: :class:`shapely.geometry.point.Point`,
         :class:`shapely.geometry.linestring.LineString`,
-        :class:`shapely.geometry.linestring.LinearRing`,
+        :class:`shapely.geometry.polygon.LinearRing`,
```
greglucas, line 69: I think these are all exported in the top-level shapely namespace so could we reference them as `shapely.LinearRing` directly now and not worry about the nested modules.

lib/cartopy/mpl/patch.py
greglucas, line 1: A few thoughts I'll throw out here to get other's feedback on whether they would be good/bad.

1. Change this to `mpl/path.py` as there isn't anything to do with matplotlib patches at the moment, it is all path handling.
2. An idea I had while reading this is that could have a `GeoPath(Path)` that subclasses matplotlib's Path and then add class methods on that `GeoPath.to_shapely()`, `GeoPath.from_shapely()` and the `_ensure_closed` function could be attached to that as well. I'm not sure if it makes sense to make a subclass of Matplotlib's Path or not, but thought I'd throw that idea out there in case it sparks something to anyone else and it fits with our subclass of `GeoAxes(Axes)`.

lib/cartopy/mpl/patch.py
```diff
@@ -249,3 +265,183 @@ def not_zero_poly(geom):
     result = list(filter(not_zero_poly, geom_collection))
 
     return result + other_result_geoms
+
+
+def path_to_shapely(path, force_ccw=False):
+    """
+    Create a Shapely geometric objects from a :class:`matplotlib.path.Path`.
```
greglucas, line 272:
```suggestion
    Create a Shapely geometric object from a :class:`matplotlib.path.Path`.
```

lib/cartopy/mpl/patch.py
```diff
@@ -249,3 +265,183 @@ def not_zero_poly(geom):
...
+
+    Parameters
+    ----------
+    path
+        A :class:`matplotlib.path.Path` instance.
+
+    Other Parameters
+    ----------------
+    force_ccw
```
greglucas, line 281: Are we able to remove this parameter? I couldn't find any uses of `force_ccw` in Cartopy.

lib/cartopy/mpl/patch.py
```diff
@@ -249,3 +265,183 @@ def not_zero_poly(geom):
...
+        :class:`shapely.geometry.polygon.Polygon`,
+        :class:`shapely.geometry.linestring.LineString`
+        :class:`shapely.geometry.point.Point`,
+        :class:`shapely.geometry.multipolygon.MultiPolygon`
+        :class:`shapely.geometry.multilinestring.MultiLineString`
+        :class:`shapely.geometry.multipoint.MultiPoint`
+        :class:`shapely.geometry.collection.GeometryCollection`.
+
+    """
+    # Convert path into numpy array of vertices (and associated codes)
```
greglucas, line 298: Are we able to leverage any of the recent contour logic / path handling you were updating in Matplotlib?

https://github.com/matplotlib/matplotlib/blob/25a05f2399d6b63bbaad5a9380a8a89cec821c50/lib/matplotlib/contour.py#L795

There are a bunch of path handling routines for stitching the previous individual paths into a compound path with the recent refactor that might give some inspiration here as well.

lib/cartopy/mpl/patch.py
```diff
@@ -302,79 +322,83 @@ def path_to_shapely(path, force_ccw=False):
...
+            if len(linestrings) > 1:
+                return sgeom.MultiLineString(linestrings)
+            else:
+                return linestrings[0]
+    else:
+        if not linestrings and not points:
+            if len(polygons) > 1:
+                return sgeom.MultiPolygon(polygons)
+            else:
+                return polygons[0]
```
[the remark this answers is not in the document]
greglucas: There is a lot of logic encapsulated here, so I think you might only be able to get it slightly simpler.

Maybe you could put these cases into a pattern matching syntax if we are only supporting 3.10+ for the next release... That might not be any easier/cleaner, so just throwing that out there as a possible idea. I think this looks OK currently.

lib/cartopy/mpl/patch.py
```diff
@@ -249,3 +265,183 @@ def not_zero_poly(geom):
...
+                                   shape.interiors], 1).T
+        codes = np.concatenate([poly_codes(shape.exterior)] +
+                               [poly_codes(ring) for ring in shape.interiors])
+        return Path(vertices, codes)
+    elif isinstance(shape, (sgeom.MultiPolygon, sgeom.GeometryCollection,
+                            sgeom.MultiLineString, sgeom.MultiPoint)):
+        paths = []
+        for geom in shape.geoms:
+            path = shapely_to_path(geom)
+            if _MPL_38 or path.vertices.size > 0:
```
greglucas, line 440: Why do we need `_MPL_38` here? Might be good to put a quick comment in here about that.

lib/cartopy/mpl/patch.py
```diff
@@ -249,3 +265,183 @@ def not_zero_poly(geom):
...
+                paths.append(path)
+        return Path.make_compound_path(*paths)
+    elif hasattr(shape, '_as_mpl_path'):
```
greglucas, line 443: What shapes/objects have `_as_mpl_path` that we haven't handled above? This seems a bit odd to have the function `shapely_to_path` but then accept other things besides shapely geometries.

lib/cartopy/tests/mpl/test_patch.py
```diff
@@ -4,13 +4,15 @@
...
 from matplotlib.path import Path
+import pytest
 import shapely.geometry as sgeom
 
 import cartopy.mpl.patch as cpatch
 
 
-class Test_path_to_geos:
-    def test_empty_polygon(self):
+@pytest.mark.parametrize('legacy', [False, True])
```
greglucas, line 13: Maybe be a bit more explicit in the variable name about what is legacy here.
```suggestion
@pytest.mark.parametrize('use_legacy_path_to_geos', [False, True])
```

docs/source/whatsnew/v0.25.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Deprecations and Removals
+-------------------------
+
+* `~cartopy.mpl.patch.path_to_geos` and `~cartopy.mpl.patch.geos_to_path` are
+  deprecated.  Use `~cartopy.mpl.path.path_to_shapely` and
+  `~cartopy.mpl.path.shapely_to_path` instead.
+
+* `cartopy.mpl.patch.path_segments` has moved to `cartopy.mpl.path.path_segments`.
```
rcomer, line 24: I'm wondering if we actually want to privatise `path_segments`?  Since it's only two lines it's not hard for downstream to vendor, and it would mean we have more freedom to do what we want with it in future.

lib/cartopy/tests/mpl/test_path.py
```diff
@@ -0,0 +1,92 @@
+# Copyright Crown and Cartopy Contributors
```
rcomer, line 1: I used `git mv` to change this from `test_patch.py` so I'm not clear why it's showing as a deletion and new file :confused:

lib/cartopy/mpl/patch.py
rcomer: ContourSet's `allsegs` and `allkinds` use the private `Path._iter_connected_components`.  We could raise a feature request to have that made public.  However, I made a quick test swapping out our path splitting logic here for `_iter_connected_components` and got some spectacular image test failures.  I didn't dig too deep, but it seems some of the `_iter_connected_components` paths have an extra `STOP` on the end which our own splitting logic doesn't produce.

Or have I misunderstood what you meant?

lib/cartopy/tests/mpl/test_path.py
rcomer: I moved the rename to its own commit, so it's now easier to see what I actually changed.  It also restored the blame that had gone missing, so I propose to keep that separate commit at merge.

greglucas (approved):

PR #2458, opened 2024-10-23
ENH: Add HEALPix and rHEALPix projections

### LINKED ISSUE #882: Projection extension - Gnomonic cubed-sphere

### LINKED ISSUE #2133: Cubic map projection
### Description

I have a need to utilize a cubic map projection (which as an example is supported by [ArcGIS](https://pro.arcgis.com/en/pro-app/latest/help/mapping/properties/cube.htm).
image

Would this be a desired feature to have added to Cartopy? If so, I'd be willing to work on it but would need significant assistance as I've not worked with map projections before.

greglucas:
## Rationale

This adds the HEALPix and rHEALPix projections from PROJ to Cartopy.
https://proj.org/en/9.4/operations/projections/healpix.html
https://proj.org/en/9.4/operations/projections/rhealpix.html

This replaces #1008 and allows for flexible globes/radii as well.

closes #882
closes #1008
closes #2133


greglucas:
For class names, do we want `Healpix()` or `HEALPix()`? Similarly for the rectangular version, do we want `Rectangular` spelled out, or is `RHEALPix` as I've put it ok?

rcomer:
Also closes #2133 I think.

lgolston:
Always enjoy seeing new projections be added! For the name, I would vote for HEALPix() and rHEALPix().

docs/make_projection.py
```diff
@@ -162,7 +165,14 @@ def prj_class_sorter(cls):
             # Get instance arguments and number of plots
             instance_args = SPECIFIC_PROJECTION_KWARGS.get(prj, {})
 
-            prj_inst, instance_repr = create_instance(prj, instance_args)
+            try:
```
greglucas, line 168: Note that I added this here because I don't have PROJ 9.6 installed locally, so this didn't work when instantiated. Maybe there is a better way to approach this?

greglucas:
> Always enjoy seeing new projections be added! For the name, I would vote for HEALPix() and rHEALPix().

I think I prefer the capitalization of "R" to indicate it is a class in Python by common convention, even if PROJ does lowercase it. Any other votes/preferences on naming?

I think this is ready for review again, and I added in a review comment about the doc build process when someone doesn't have the latest PROJ installed.

QuLogic:
I rebased and fixed some conflicts (mostly just `sgeom` -> `shapely`).

rcomer:
Cycling to pick up #2723.

QuLogic:
It seems that RTD didn't pick up the new merge commit, so I rebased again.

greglucas:
I had missed a globe pass-through in RHEALPix and I had used a hard-coded number as the fallback instead of WGS84 constant that we use elsewhere in the file. I've updated both of those issues now and pushed the new branch.

The RHEALPix map blue/red lines look quite bad to me, but I don't think there is much we can do about that currently. I'm hoping that the adaptive resampling with a finer sampling would improve it, but we should leave that as a follow-up. I think the current issue is that we might be snapping to a boundary too soon instead of interpolating along the line. This might be a really good test case for adaptive resampling to see how it does with this kind of square/box-y projection.

lib/cartopy/crs.py
```diff
@@ -1873,6 +1873,116 @@ def __init__(self):
...
+        w = (self.ellipsoid.semi_major_metre or WGS84_SEMIMAJOR_AXIS) * np.pi
+        h = w / 2
+        self.bounds = [-w, w, -h, h]
+        self.threshold = w / 1e4
+
+        self._boundary = shapely.LinearRing(
+            [(-w, h / 2), (-3 / 4 * w, h), (-w / 2, h / 2), (-w / 4, h), (0, h / 2),
+             (w / 4, h), (w / 2, h / 2), (3 / 4 * w, h), (w, h / 2), (w, -h / 2),
+             (3 / 4 * w, -h), (w / 2, -h / 2), (w / 4, -h), (0, -h / 2), (-w / 4, -h),
+             (-w / 2, -h / 2), (-3 / 4 * w, -h), (-w, -h / 2), (-w, h / 2)])
```
QuLogic: Are you sure this boundary is correct? For the default parameters, I get these gridlines:

Image

But if I set `globe=ccrs.Globe(semimajor_axis=1000, semiminor_axis=500)`, then I get this, which seems to have too large a boundary:

Image

lib/cartopy/crs.py
```diff
@@ -1873,6 +1873,116 @@ def __init__(self):
...
+                        ('lon_0', central_longitude)]
+        super().__init__(proj4_params, globe=globe)
+
+        w = (self.ellipsoid.semi_major_metre or WGS84_SEMIMAJOR_AXIS) * np.pi
+        h = 3 / 4 * w
+        box_h = w / 4
+        self.bounds = [-w, w, -h, h]
+        self.threshold = w / 1e4
+
+        self._boundary = shapely.LinearRing([
```
QuLogic: A similar problem occurs here:

Default:
Image

vs 1000/500 globe:
Image

lib/cartopy/crs.py
QuLogic: Also, looking at it again, the longitude lines are bit wrong even with the defaults. Based on the lower plot, I guess that the 0/180 lines are travelling all the way up the triangle to the point, and their labels are getting picked with higher priority over the other lines that are actually visible.

greglucas: The boundary was defined correctly, the "w" wasn't calculated correctly. The width isn't dependent on semi-major axis apparently. I've made it so we calculate the "w" based on a proj transform directly (which maybe we should do in more places to get more precise boundary widths).

This should be fixed now.

greglucas: > Also, looking at it again, the longitude lines are bit wrong even with the defaults. Based on the lower plot, I guess that the 0/180 lines are travelling all the way up the triangle to the point, and their labels are getting picked with higher priority over the other lines that are actually visible.

I think this is an issue with our gridliner and what it chooses to display and we should defer that (I have another major overhaul of the gridliner in the works). But yes this is an issue because there are many possible choices at those locations due to everything meeting at a point.

lib/cartopy/crs.py
greglucas: Same update as above to the REALPix.

QuLogic:
I guess you aren't in a rush to get this into 0.26 since that's been tagged.

greglucas:
Yes, I was more interested in getting the release out than getting this PR in. This PR had already sat for a while. We can focus on getting more projections added for the next release, this should be ready for a review with all comments addressed.

Issue #2030, opened 2022-03-29
AzimuthalEquidistant Landarea Projection issue in the southern hemisphere


lsawade:
### Description

Hi Team,

I have been encountering an issue that may be related to other issues here concerning the accurate plotting of land areas when providing the keywords `central_longitude=??` and `central_latitude=??`. I attached a working and a failing example. There is no error, but the plotting is weird. I'm running on macOS with a fresh `conda` environment with only `cartopy` and `matplotlib` installed like so:

```bash
conda create -n cartopy python matplotlib cartopy
```

When creating a `AzimuthalEquidistant` projection instance using 

```python
central_longitude=-66
central_latitude=-25
```

The land areas fill the entire map:

![Screen Shot 2022-03-29 at 9 31 51 AM](https://user-images.githubusercontent.com/23224303/160622864-876a8d27-9372-471a-81ff-32f47fb0c2c5.png)

However, if I use a latitude just slightly closer to the equator the projection is completely fine:

```python
central_longitude=-66
central_latitude=-22
```

![Screen Shot 2022-03-29 at 9 30 51 AM](https://user-images.githubusercontent.com/23224303/160622655-97b68beb-7644-43ae-a92c-6b0e91d25a1b.png)

##### Failing example

`central_longitude=-66, central_latitude=-25`

```
from cartopy.crs import AzimuthalEquidistant, PlateCarree
import matplotlib.pyplot as plt
import cartopy

fig = plt.figure()

projection = AzimuthalEquidistant(central_longitude=-66, central_latitude=-25)


ax = plt.axes(projection=projection)
ax.set_global()

# only fails for very certain central_longitude=????, central_latitude=????
ax.add_feature(cartopy.feature.LAND, zorder=0, edgecolor='k',
                       linewidth=0.5, facecolor=(0.8, 0.8, 0.8))
plt.show()
```

##### Working as expected example

`central_longitude=-66, central_latitude=-25`

```
from cartopy.crs import AzimuthalEquidistant, PlateCarree
import matplotlib.pyplot as plt
import cartopy

fig = plt.figure()

projection = AzimuthalEquidistant(central_longitude=-66, central_latitude=-25)


ax = plt.axes(projection=projection)
ax.set_global()

# only fails for very certain central_longitude=????, central_latitude=????
ax.add_feature(cartopy.feature.LAND, zorder=0, edgecolor='k',
                       linewidth=0.5, facecolor=(0.8, 0.8, 0.8))
plt.show()
```

[...]

Issue #1076, opened 2018-06-08
some contourf levels not drawn in Orthographic projection (are drawn by contour)


keatonb:
### Description

I'm trying to draw spherical harmonic contours on an orthographic projection of a sphere.  I see the correct contours when calling contour(), but many are not filled by contourf(). I think this is a bug, perhaps related to [Issue 1024](https://github.com/SciTools/cartopy/issues/1024) (but I couldn't view their example figures).

#### Code to reproduce

```
import matplotlib.pyplot as plt
import numpy as np
import scipy.special as sp
import cartopy
import cartopy.crs as ccrs

# Compute spherical harmonic pattern
l=4; m=3
lon = np.linspace(0,2*np.pi,200)
lat = np.linspace(-np.pi/2,np.pi/2,500)
colat = lat+np.pi/2
d = np.zeros((len(lon),len(colat)),dtype = np.complex64)
meshed_grid = np.meshgrid(lon, lat)
lat_grid = meshed_grid[1]
lon_grid = meshed_grid[0]
for j, yy in enumerate(colat):
    for i, xx in enumerate(lon):
        d[i,j] = sp.sph_harm(m,l,xx,yy)
drm = np.transpose(np.real(d))

#Plot example with contour() and contourf()
fig, (ax1,ax2) = plt.subplots(1, 2, subplot_kw={'projection': ccrs.Orthographic(0, 30)})

#contour() works as expected
ax1.contour(lon*180/np.pi,lat*180/np.pi,drm,51,
             transform=ccrs.PlateCarree(),cmap='seismic')

ax1.relim()
ax1.autoscale_view()

#contourf() doesn't
ax2.contourf(lon*180/np.pi,lat*180/np.pi,drm,51,
             transform=ccrs.PlateCarree(),cmap='seismic')

ax2.relim()
ax2.autoscale_view()

plt.tight_layout()
plt.savefig('sphericalcontours.png')
```

![sphericalcontours](https://user-images.githubusercontent.com/6413923/41180702-5823f560-6b6f-11e8-8eb8-d218d01d321c.png)
#### Traceback 

No error reported.

<details>
 <summary>Full environment definition</summary>

### Operating system
OS X 10.10.5

### Cartopy version
0.16.0

### conda list

```
# packages in environment at /Users/keatonb/anaconda:
#
# Name                    Version                   Build  Channel
_license                  1.1                      py27_0    <unknown>
abstract-rendering        0.5.1                np19py27_0    <unknown>
[...]

PR #1739, opened 2021-02-26
Treatment of interior rings from transform between feature and projection domains

kdpenner:
## Rationale

Fixes #1685, #1613, #1149, #1076, #1723, #955, #1367, #1617, and a bunch of other unfiled issues. NB a few of them were incorrectly marked as duplicates. Methodology applicable to #1362. Potentially solves #323 and #1449?

Some of the test suite failures are expected.

This PR is a work in progress. Debugging junk is everywhere and more work is needed. I'm submitting now as a heads-up; the changes will need extensive testing. In particular I've done no debugging of the sampling of non-Plate Carree feature CRSs and have written no unit tests.

A summary of the discoveries this PR solves:

 1) points that project to infinity are improperly included (-> changes to `FeatureArtist`)
 2) `_project_linear_rings` sometimes creates small interior rings. `_rings_to_multi_polygon` inverts the rings, which leads to polygons covering almost the entire domain (-> changes to `_rings_to_multi_polygon`)
 3) OTOH `_project_linear_rings` sometimes creates exterior rings encircling the entire domain or almost the entire domain. If two of these rings are created the interior ring removal logic leads to polygons covering almost the entire domain (-> changes to `_rings_to_multi_polygon`)
 4) correcting 2) or 3) invalidates the `MultiPolygon` creation process (-> changes to `_project_polygon` and `_project_multipolygon`
 5) `_project_linear_rings` sometimes flips the orientation of the rings. i.e. land polygons are filled when ocean polygons are passed

## Implications

Performance is reduced, severely for large polygons. If invalid in their native CRS they have to be buffered for a difference, then buffered again in the projection CRS. But the following plot is successful.

PR #2470, opened 2024-11-02
FIX: create a single inverted polygon when no exteriors found

### LINKED ISSUE #1149: Broken ocean polygon reprojection with LambertAzimuthalEqualArea
> May I add `LambertAzimuthalEqualArea` to the list or reproducers with current master + matplotlib2

```python
proj = cartopy.crs.LambertAzimuthalEqualArea(central_latitude=45,
                                             central_longitude=-100)
ax.add_feature(cartopy.feature.OCEAN)
ax.coastlines()
ax.set_extent([-140, -70, 20, 60])
```

test

_Originally posted by @akrherz in https://github.com/SciTools/cartopy/issues/803#issuecomment-333909742_

### LINKED ISSUE #1449: Error plotting data with Orthographic() map projection
I am trying to plot a netcdf file, which can be found here:
[testcube.zip](https://github.com/SciTools/cartopy/files/4149382/testcube.zip)

The issue is that the Orthographic projection incorrectly plots the data. Seemingly, all negative values are set to 0 during plotting. This is not the case for other projections. My code is below:

#### Code to reproduce

```
import iris
import cartopy
import cartopy.crs as ccrs
import iris.plot as iplt
print(cartopy.__version__)

cube=iris.load_cube("testcube.nc")
print(cube)
print(cube.coord("latitude"))
print(cube.coord("longitude"))

fig,axes=plt.subplots(4)
clevs=np.linspace(-abs(cube.data).max(),abs(cube.data).max(),21)
projs=[ccrs.PlateCarree(),ccrs.Mollweide(),ccrs.Orthographic(0,90),ccrs.NearsidePerspective(central_latitude=50,central_longitude=-20)]
for i,proj in enumerate(projs):
    ax=plt.subplot(4,1,i+1,projection=proj)
    ax.coastlines()
    plot=iplt.contourf(cube, levels=clevs, cmap=plt.cm.RdBu_r,axes=ax)
    plt.colorbar(mappable=plot)

```
Which produces the following output:
```
'0.17.0'

unknown / (unknown)                 (latitude: 73; longitude: 144)
     Dimension coordinates:
          latitude                           x              -
          longitude                          -              x
     Scalar coordinates:
          mean cluster composites: 0
DimCoord(array([ 90. ,  87.5,  85. ,  82.5,  80. ,  77.5,  75. ,  72.5,  70. ,
        67.5,  65. ,  62.5,  60. ,  57.5,  55. ,  52.5,  50. ,  47.5,
        45. ,  42.5,  40. ,  37.5,  35. ,  32.5,  30. ,  27.5,  25. ,
        22.5,  20. ,  17.5,  15. ,  12.5,  10. ,   7.5,   5. ,   2.5,
         0. ,  -2.5,  -5. ,  -7.5, -10. , -12.5, -15. , -17.5, -20. ,
       -22.5, -25. , -27.5, -30. , -32.5, -35. , -37.5, -40. , -42.5,
       -45. , -47.5, -50. , -52.5, -55. , -57.5, -60. , -62.5, -65. ,
       -67.5, -70. , -72.5, -75. , -77.5, -80. , -82.5, -85. , -87.5,
       -90. ], dtype=float32), standard_name='latitude', units=Unit('degrees'), var_name='latitude')
DimCoord(array([  0. ,   2.5,   5. ,   7.5,  10. ,  12.5,  15. ,  17.5,  20. ,
        22.5,  25. ,  27.5,  30. ,  32.5,  35. ,  37.5,  40. ,  42.5,
        45. ,  47.5,  50. ,  52.5,  55. ,  57.5,  60. ,  62.5,  65. ,
        67.5,  70. ,  72.5,  75. ,  77.5,  80. ,  82.5,  85. ,  87.5,
        90. ,  92.5,  95. ,  97.5, 100. , 102.5, 105. , 107.5, 110. ,
       112.5, 115. , 117.5, 120. , 122.5, 125. , 127.5, 130. , 132.5,
       135. , 137.5, 140. , 142.5, 145. , 147.5, 150. , 152.5, 155. ,
       157.5, 160. , 162.5, 165. , 167.5, 170. , 172.5, 175. , 177.5,
       180. , 182.5, 185. , 187.5, 190. , 192.5, 195. , 197.5, 200. ,
       202.5, 205. , 207.5, 210. , 212.5, 215. , 217.5, 220. , 222.5,
       225. , 227.5, 230. , 232.5, 235. , 237.5, 240. , 242.5, 245. ,
       247.5, 250. , 252.5, 255. , 257.5, 260. , 262.5, 265. , 267.5,
       270. , 272.5, 275. , 277.5, 280. , 282.5, 285. , 287.5, 290. ,
       292.5, 295. , 297.5, 300. , 302.5, 305. , 307.5, 310. , 312.5,
       315. , 317.5, 320. , 322.5, 325. , 327.5, 330. , 332.5, 335. ,
       337.5, 340. , 342.5, 345. , 347.5, 350. , 352.5, 355. , 357.5],
      dtype=float32), standard_name='longitude', units=Unit('degrees'), var_name='longitude', circular=True)
```
testfig

### LINKED ISSUE #1674: Inaccurate distinction between ocean and land when plotting the Arctic
Now, I'm ready to plot the map for the Arctic, with land and ocean in different colors. 
However, when I add_feature to the land and ocean, the color of ocean will overlap the land (or the continents near the Arctic). That is, the cartopy may be treat the ocean inaccurately in the Arctic.

The code for this map is shown as follows:

```python
import cartopy.crs as ccrs
fig = plt.figure(figsize=[5,5])
ax1 = fig.add_subplot(1, 1, 1, projection=ccrs.NorthPolarStereo(central_longitude=0))
ax1.set_extent([-180, 180,60, 90], ccrs.PlateCarree())
ax1.gridlines()
ax1.add_feature(cfeature.LAND)
ax1.add_feature(cfeature.OCEAN)
plt.show()
```

However, if I exchange the order for the two sentences, write the "ax1.add_feature(cfeature.OCEAN)"  first and then write "ax1.add_feature(cfeature.LAND)", it seems works.

#### Code to reproduce

```

```

#### Traceback 

```

```

<details>

### conda list

```
```

### pip list

```
```

</details>

### LINKED ISSUE #2160: Inconsistency drawing LAND/OCEAN cfeature for PolarStereo maps
### Description

Hi all, 

I wanted to report an inconsistency in adding features for the North/SouthPolarStereo maps. 

I created a NorthPolarStereo map and added the LAND and OCEAN cfeatures. I was surprised that the LAND feature was not plotted.

```python
import cartopy.crs as ccrs
```
```python
ax = plt.axes(projection=ccrs.NorthPolarStereo())

ax.coastlines()
ax.add_feature(cfeature.LAND)
ax.add_feature(cfeature.OCEAN)
```
image

It actually appears the LAND feature is drawn "under" the ocean. If I swap the order LAND and OCEAN are drawn, then I get what I expected on the right.

```python
fig, (ax1, ax2) = plt.subplots(1,2, subplot_kw={'projection':ccrs.NorthPolarStereo()})

ax1.coastlines()
ax1.add_feature(cfeature.LAND)
ax1.add_feature(cfeature.OCEAN)
ax1.set_title('North ax1')

ax2.coastlines()
ax2.add_feature(cfeature.OCEAN) #<-- Swap order
ax2.add_feature(cfeature.LAND)
ax2.set_title('North ax2')
```

image

This behavior is inconsistent with adding these features on `SouthPolarStereo`, where it doesn't matter the order LAND and OCEAN are drawn; LAND is always on top of OCEAN. This is the behavior I expected NorthPolarStereo to also have.

```python
fig, (ax1, ax2) = plt.subplots(1,2, subplot_kw={'projection':ccrs.SouthPolarStereo()})

ax1.coastlines()
ax1.add_feature(cfeature.LAND)
ax1.add_feature(cfeature.OCEAN)
ax1.set_title('South ax1')

ax2.coastlines()
ax2.add_feature(cfeature.OCEAN) #<-- Swap order
ax2.add_feature(cfeature.LAND)
ax2.set_title('South ax2')
```
image

### Matplotlib version 3.7.1

rcomer:
## Rationale

When `_rings_to_multipolygon` finds interior rings without associated exteriors, for each of these interiors it creates a new polygon as the difference between ~the whole map and that interior.  Intuitively, if there are multiple interiors, then all of those whole map polygons will overlay each other and you won't see the holes.  Instead it should create a single ~whole map polygon which uses all of those left over interiors.

I also switched out use of `.buffer(0)` for `shapely.make_valid` as that seems like the more official way to do that.

Fixes #1149, fixes #1674, fixes #2160 (all ocean feature examples) and fixes #1449 (`contourf` example)
Also fixes the original case from #2030 (but not the second example so not linking that issue)
Also fixes the `contourf` examples at https://github.com/SciTools/cartopy/issues/1076#issuecomment-1774329796 and https://github.com/SciTools/cartopy/issues/1076#issuecomment-591926911

I turned #1149 into an image test.  Really there should be something more specific in `test_polygon.py`, but I'm at a loss how to create an artificial example for this.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -21,6 +21,7 @@ import numpy as np
 import pyproj
 from pyproj import Transformer
 from pyproj.exceptions import ProjError
+import shapely
 import shapely.geometry as sgeom
 from shapely.prepared import prep
 
@@ -1214,10 +1215,26 @@ class Projection(CRS, metaclass=ABCMeta):
             y3 -= by
             x4 += bx
             y4 += by
+
+            interior_polys = []
+
             for ring in interior_rings:
-                # Use shapely buffer in an attempt to fix invalid geometries
-                polygon = sgeom.Polygon(ring).buffer(0)
-                if not polygon.is_empty and polygon.is_valid:
+                polygon = shapely.make_valid(sgeom.Polygon(ring))
+                if not polygon.is_empty:
+                    if isinstance(polygon, sgeom.Polygon):
+                        interior_polys.append(polygon)
+                    elif isinstance(polygon, sgeom.MultiPolygon):
+                        interior_polys.extend(polygon.geoms)
+                    elif isinstance(polygon, sgeom.GeometryCollection):
+                        for geom in polygon.geoms:
+                            if isinstance(geom, sgeom.Polygon):
+                                interior_polys.append(geom)
+                            elif isinstance(geom, sgeom.MultiPolygon):
+                                interior_polys.extend(geom.geoms)
+                    else:
+                        # make_valid may produce some linestrings.  Ignore these
+                        continue
+
                     x1, y1, x2, y2 = polygon.bounds
                     bx = (x2 - x1) * 0.1
                     by = (y2 - y1) * 0.1
@@ -1225,23 +1242,27 @@ class Projection(CRS, metaclass=ABCMeta):
                     y1 -= by
                     x2 += bx
                     y2 += by
-                    box = sgeom.box(min(x1, x3), min(y1, y3),
-                                    max(x2, x4), max(y2, y4))
 
-                    # Invert the polygon
-                    polygon = box.difference(polygon)
+                    x3 = min(x1, x3)
+                    x4 = max(x2, x4)
+                    y3 = min(y1, y3)
+                    y4 = max(y2, y4)
 
-                    # Intersect the inverted polygon with the boundary
-                    polygon = boundary_poly.intersection(polygon)
+            box = sgeom.box(x3, y3, x4, y4, ccw=is_ccw)
 
-                    if not polygon.is_empty:
-                        polygon_bits.append(polygon)
+            # Invert the polygons
+            polygon = box.difference(sgeom.MultiPolygon(interior_polys))
 
-        if polygon_bits:
-            multi_poly = sgeom.MultiPolygon(polygon_bits)
-        else:
-            multi_poly = sgeom.MultiPolygon()
-        return multi_poly
+            # Intersect the inverted polygon with the boundary
+            polygon = boundary_poly.intersection(polygon)
+
+            if not polygon.is_empty:
+                if isinstance(polygon, sgeom.MultiPolygon):
+                    polygon_bits.extend(polygon.geoms)
+                else:
+                    polygon_bits.append(polygon)
+
+        return sgeom.MultiPolygon(polygon_bits)
 
     def quick_vertices_transform(self, vertices, src_crs):
         """
--- a/lib/cartopy/tests/test_polygon.py
+++ b/lib/cartopy/tests/test_polygon.py
@@ -446,6 +446,18 @@ class TestHoles(PolygonTests):
         self._assert_bounds(polygon.interiors[0].bounds,
                             - 1.2e7, -1.2e7, 1.2e7, 1.2e7, 1e6)
 
+    def test_inverted_poly_multi_hole(self):
+        # Adapted from 1149
+        proj = ccrs.LambertAzimuthalEqualArea(
+            central_latitude=45, central_longitude=-100)
+        poly = sgeom.Polygon([(-180, -80), (180, -80), (180, 90), (-180, 90)],
+                             [[(-50, -50), (-50, 0), (0, 0), (0, -50)]])
+        multi_polygon = proj.project_geometry(poly)
+        # Should project to single polygon with multiple holes
+        assert len(multi_polygon.geoms) == 1
+        assert len(multi_polygon.geoms[0].interiors) >= 2
+
+
     def test_inverted_poly_clipped_hole(self):
         proj = ccrs.NorthPolarStereo()
         poly = sgeom.Polygon([(0, 0), (-90, 0), (-180, 0), (-270, 0)],


lib/cartopy/crs.py
```diff
@@ -1214,34 +1215,52 @@ def _rings_to_multi_polygon(self, rings, is_ccw):
...
 
-                    if not polygon.is_empty:
-                        polygon_bits.append(polygon)
+            # Invert the polygons
+            polygon = box.difference(sgeom.MultiPolygon(interior_polys))
 
-        if polygon_bits:
-            multi_poly = sgeom.MultiPolygon(polygon_bits)
-        else:
-            multi_poly = sgeom.MultiPolygon()
```
rcomer, line 1243: Note that `MutilPolygon` will work fine with an empty list.

lib/cartopy/crs.py
```diff
@@ -1214,34 +1215,52 @@ def _rings_to_multi_polygon(self, rings, is_ccw):
...
-                # Use shapely buffer in an attempt to fix invalid geometries
-                polygon = sgeom.Polygon(ring).buffer(0)
-                if not polygon.is_empty and polygon.is_valid:
+                polygon = shapely.make_valid(sgeom.Polygon(ring))
+                if not polygon.is_empty:
+                    if isinstance(polygon, sgeom.Polygon):
+                        interior_polys.append(polygon)
+                    elif isinstance(polygon, sgeom.MultiPolygon):
+                        interior_polys.extend(polygon.geoms)
+                    elif isinstance(polygon, sgeom.GeometryCollection):
```
rcomer, line 1228: We do get ` GeometryCollection` for the #1149 case.

rcomer:
I note that `_rings_to_multipolygon` is called within `_project_polygon`.  If you start with a `MultiPolygon` it will just loop through the polygons so we might plausibly create multiple overlapping whole map polygons that way.  So (plausibly) we might fix more cases by applying this logic at the `MultiPolygon` level.  I propose to get this fix in first and dig into the `MultiPolygon` question later though.

rcomer:
I appear to be reinventing part of #1739...

greglucas:
> Really there should be something more specific in test_polygon.py, but I'm at a loss how to create an artificial example for this.

I agree. Simplifying these down would really help us debug future issues too because I feel like it is confusing what the _root_ cause of the issue is sometimes. A few thoughts in case this sparks anything.

* All geometries causing issues are coming from the default PlateCarree, so we should be able to create something in degrees.
* Most of the issues seem to be around Lambert* transforms, so going to an r/theta coordinate frame.
* Are there situations where the bottom left corner of a box gets transformed into the upper right corner of a box when you are right around the central longitude / maximum latitude? (possibly a rhombus/diamond shape where one point is on the opposite side of a boundary)

> I appear to be reinventing part of https://github.com/SciTools/cartopy/pull/1739...

That has a lot going on. Maybe this is a small piece of that and we can keep doing incremental updates?

rcomer:
OK here's the simplest I have so far by studying https://github.com/SciTools/cartopy/issues/1149#issuecomment-430163998

```python
import cartopy.crs as ccrs

sudo_ocean = sgeom.Polygon([(-180, -80), (180, -80), (180, 90), (-180, 90)],
                           [[(-50, -50), (-50, 0), (0, 0), (0, -50)]])

fig = plt.figure(figsize=(6, 8))

ax2 = fig.add_subplot(2, 1, 2, projection=ccrs.LambertAzimuthalEqualArea(
                      central_latitude=45, central_longitude=-100))

for ax in [ax1, ax2]:
    ax.add_geometries(sudo_ocean, crs=ccrs.PlateCarree())
    ax.set_global()
    
plt.show()
```
`main`:
image

branch:
image

lib/cartopy/tests/test_polygon.py
```diff
@@ -446,6 +446,18 @@ def test_inverted_poly_simple_hole(self):
...
+    def test_inverted_poly_multi_hole(self):
+        # Adapted from 1149
+        proj = ccrs.LambertAzimuthalEqualArea(
+            central_latitude=45, central_longitude=-100)
+        poly = sgeom.Polygon([(-180, -80), (180, -80), (180, 90), (-180, 90)],
+                             [[(-50, -50), (-50, 0), (0, 0), (0, -50)]])
+        multi_polygon = proj.project_geometry(poly)
+        # Should project to single polygon with multiple holes
+        assert len(multi_polygon.geoms) == 1
+        assert len(multi_polygon.geoms[0].interiors) >= 2
```
rcomer, line 458: We actually get 36 holes.  I assume most of them are responsible for the seam down the dateline visible in https://github.com/SciTools/cartopy/pull/2470#issuecomment-2453463888

Against main, this test fails at the first `assert` since we have two polygons in our multipolygon.

rcomer (commented):
Should I remove the image test?

lib/cartopy/tests/test_polygon.py
greglucas: I would remove the image test since it is downloading the natural earth ocean for every runner, or replace it with this example as an image comparison test that doesn't require external downloading.

The 36 holes is interesting and worth investigating IMO. For your example here, it looks like there are many small loops around the boundary points that are created. (your image has a small white curve/line in it protruding from the south pole area which are these linear rings I think)
```python

ax = plt.figure().add_subplot()

interior = multi_polygon.geoms[0].interiors
for i in range(len(interior)):
    ax.plot(*interior[i].xy)

plt.show()
```

I wonder if these extra loops/rings are what is slowing down some of our projecting of geometries even 🤔 

I'm also happy to defer these investigations to future PRs because this is already a pretty substantial improvement I think.

greglucas:
I dug into the multiple rings issue a bit and I think the main issue is projecting the seam along the dateline. On one side of the boundary we go up and the other we go down, and because our intermediate point bisection routine is not the same going from A->B as B->A. A quick example with just line segments being plotted and then zooming in you can see that the points aren't exactly aligned and then the segments overlap one another making invalid polygons.

```python
import cartopy.crs as ccrs

# Adapted from 1149
proj = ccrs.LambertAzimuthalEqualArea(
    central_latitude=45, central_longitude=-100)
x = 179
l0 = sgeom.LineString([(-180, -80), (-180, 80)])
l1 = sgeom.LineString([(-180, 80), (-180, -80)])
p0 = proj.project_geometry(l0).geoms[0]
p1 = proj.project_geometry(l1).geoms[0]

ax = plt.figure().add_subplot()
ax.plot(p0.xy[0], p0.xy[1])
ax.plot(p1.xy[0], p1.xy[1])
plt.show()
```

I thought there might be an easy way of modifying `project_segment()` and swapping the start/end points so that it is always going the same direction regardless of inputs. But, new lines are started/stopped within that routine so the reversal isn't as trivial as I was hoping it would be... I think ultimately we need a new interpolation/projection routine which is a much bigger task.

rcomer:
I had a go at https://github.com/SciTools/cartopy/pull/2470#issuecomment-2453372909 with https://github.com/rcomer/cartopy/commit/f4408721b16ae000bde70892ef638f8a3cc12d57.  It did not break any exiting tests but also did not fix any issue.  So back to the drawing board!  Also it does make sense that the projection of a multipolygon should give the same result as projecting those polygons individually, so if it had worked it would have been more of a workaround than an actual fix I think.

greglucas (approved):
I think this is a good improvement and fixes many of the cases. We shouldn't let perfect be the enemy of good here. There is room for follow-up PRs after this to keep improving the situation here.

Thanks @rcomer for digging into all of this!

Issue #2319, opened 2024-01-29
Key Error using Natural Earth POV files


rcomer:
### Description

When trying to use [Natural Earth's POV shapefiles](https://www.naturalearthdata.com/about/disputed-boundaries-policy/), the unzipping step fails because they do not include a `.cpg` file.  I wonder if we just need a try-except around the `getinfo` call.  Having said that, out of the list

https://github.com/SciTools/cartopy/blob/ec891b0f119a5a21f7c404b49e8cc2f84a50ea38/lib/cartopy/io/shapereader.py#L342

we only appear to use `.shp` and `.shx` in other parts of the library.  So is there a reason we are unpacking all of them?

#### Code to reproduce

```python
import cartopy
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

mpov = cartopy.feature.NaturalEarthFeature('cultural', 'admin_0_countries_arg', '10m')

fig, ax = plt.subplots(subplot_kw={"projection": ccrs.PlateCarree()})
ax.add_feature(mpov, facecolor="none")

plt.show()
```

#### Traceback

On first run, when Cartopy is downloading and unzipping the files I get

```
[script_path]/natural_earth.py:3: UserWarning: Treat the new Tool classes introduced in v1.5 as experimental for now; the API and rcParam may change in future versions.
  import matplotlib.pyplot as plt
[git_clone_path]/cartopy/lib/cartopy/io/__init__.py:241: DownloadWarning: Downloading: https://naturalearth.s3.amazonaws.com/10m_cultural/ne_10m_admin_0_countries_arg.zip
  warnings.warn(f'Downloading: {url}', DownloadWarning)
Exception in Tkinter callback
Traceback (most recent call last):
  File "[conda_env_path]/cartopy-dev/lib/python3.12/tkinter/__init__.py", line 1962, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "[conda_env_path]/cartopy-dev/lib/python3.12/tkinter/__init__.py", line 861, in callit
    func(*args)
  File "[conda_env_path]/cartopy-dev/lib/python3.12/site-packages/matplotlib/backends/_backend_tk.py", line 274, in idle_draw
    self.draw()
  File "[conda_env_path]/cartopy-dev/lib/python3.12/site-packages/matplotlib/backends/backend_tkagg.py", line 10, in draw
    super().draw()
[...]

PR #2477, opened 2024-11-13
MNT: Extract all NE resources from the zip file

greglucas:
## Rationale

The various Natural Earth datasets can have different file members within them. Rather than explicitly iterating through the list of expected files, lets just extract all of them. This adds a `datasetname.README.txt` and `datasetname.VERSION.txt` which could be helpful for debugging purposes too and the sizes are very small for the extra files.

closes https://github.com/SciTools/cartopy/issues/2319

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -31,6 +31,7 @@ import io
 import itertools
 from pathlib import Path
 from urllib.error import HTTPError
+from zipfile import ZipFile
 
 import shapefile
 import shapely.geometry as sgeom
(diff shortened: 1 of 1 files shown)


greglucas:
This is failing because we allow the renaming of file contents when extracting an archive. The content within the zipfile is: `ne_110m_rivers_lake_centerlines.shp`, but the download template wants to rename that to: `{category}_{resolution}_{name}.shp`. I'm not sure I fully agree with the ability to rename a file coming from an archive, versus making the reader handle different formatted filenames, but I do sort of see why you may want to do that to try and organize things your own way too.

bjlittle:
@greglucas Just a thought ... is there a compromise here to extract everything in the zipfile as you've implemented, but also as a convenience create links to the original assets following the existing `cartopy` naming convention?

Or will that just muddy the waters?

greglucas:
> Just a thought ... is there a compromise here to extract everything in the zipfile as you've implemented, but also as a convenience create links to the original assets following the existing cartopy naming convention?
>
> Or will that just muddy the waters?

@bjlittle, that is an interesting thought. I'm not opposed to it, but I do wonder if we need the renaming or if we can just get rid of that capability and simplify our lives a bit...

greglucas:
superseded by https://github.com/SciTools/cartopy/pull/2534 so closing this version.

PR #2485, opened 2024-11-26
Capitalise and protect "Python" and "Matplotlib" in suggested BibTeX citation

warrickball:
## Rationale

"Python" and "Matplotlib" are not capitalised in the example BibTeX citation and it is being copied verbatim, propagating the error.

## Implications

Editors will no longer have to fix the capitalisation of the citation downstream.

### DIFF
--- a/docs/source/citation.rst
+++ b/docs/source/citation.rst
@@ -22,7 +22,7 @@ For example::
 
  @manual{Cartopy,
  author = {{Met Office}},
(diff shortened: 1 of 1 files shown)


rcomer (approved):
Thanks for this @warrickball.

I suspect we should further review this citation as I'm not sure the years and address are still what we want, but there's no need to hold up this fix for that.

Issue #2480, opened 2024-11-20
`imshow` fails when transform projection is different that the axes projection


edsaac:
### Description

`imshow` works fine when both axes and raster data projections are the same (Figure 1), but it fails when the axes has a projection that is different from the data (Figure 2). As a workaround, the data extent has to be projected independently to the axes projection and pass that as the extent to `imshow` (Figure 3)

|Figure|Map|
|---|---|
|1|![image](https://github.com/user-attachments/assets/102a134f-0399-4271-b845-b6eb83fbbfde)|
|2|![image](https://github.com/user-attachments/assets/d11f1bd9-12ec-4ae4-a255-e7aaabc665f2)|
|3|![image](https://github.com/user-attachments/assets/9bf4659d-c446-4ccc-8ea8-29b22dc12de9)|

#### To reproduce

<details>
 <summary> Code </summary>

```py
import numpy as np
import matplotlib.pyplot as plt
import cartopy.crs as ccrs  
import cartopy.feature as cf
from shapely import Point

map_extent = (-88, -86, 41.2, 42.2)

# Some dummy raster data
raster = np.array([[1,2], [3,4]])  
raster_extent = (-87.6, -86.4, 41.4, 42.0) # (left, right, bottom, top)
raster_proj = ccrs.PlateCarree()

# Corners of the raster data
points = (raster_extent[:2], raster_extent[2:])
points_proj = ccrs.PlateCarree()


#----- Figure 1 ------------
## Map configuration
axes_proj = ccrs.PlateCarree()
fig, ax = plt.subplots(subplot_kw={"projection": axes_proj}, figsize=(6, 6))

## Add some map features
ax.add_feature(cf.LAKES, alpha=0.3)
ax.add_feature(cf.STATES)

## Add my raster data
ax.imshow(
    raster,
    extent=raster_extent,
    transform=raster_proj,
    origin="lower",
    cmap="spring"
)

## Add my point data
ax.scatter(
    *points, 
    transform=ccrs.PlateCarree(),
    s=100, c='k',
    marker="x",
    label="imshow extent"
)

## Final touches
ax.set_extent(map_extent, crs=ccrs.PlateCarree())
ax.gridlines(draw_labels=True)
ax.set_title("PlateCarree")
ax.legend()
plt.show()


#----- Figure 2 ------------
## Map configuration
axes_proj = ccrs.LambertConformal(central_longitude=-87, central_latitude=41.7)

[...]

Issue #1468, opened 2020-02-18
Inconsistent imshow interpolation


dopplershift:
### Description

The output plotting projected data depends on when `set_extent` is called. This code:
```python
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import numpy as np

data2 = np.full((10, 10), 5)
ext = (-110, -45, 50, 10)
data_proj = ccrs.PlateCarree()

proj = ccrs.LambertConformal()
bounds = (-119.0, -56.0, 19.0, 47.0)

fig = plt.figure(figsize=(20, 10))

ax1 = fig.add_subplot(2, 1, 1, projection=proj)
ax1.set_extent(bounds, ccrs.PlateCarree())
ax1.imshow(data2, extent=ext, transform=data_proj)

ax2 = fig.add_subplot(2, 1, 2, projection=proj)
ax2.imshow(data2, extent=ext, transform=data_proj)
ax2.set_extent(bounds, ccrs.PlateCarree())
```
produces this output for me (with matplotlib 3.1.3 and cartopy 0.17.0):

<img width="531" alt="image" src="https://user-images.githubusercontent.com/221526/74766709-36954b00-524b-11ea-971a-432a19a626a3.png">

QuLogic:
`imshow` is eagerly warped https://github.com/SciTools/cartopy/blob/master/lib/cartopy/mpl/geoaxes.py#L1284 and original data is lost, so this would take some work with new wrappers that update at draw/extent change.

dopplershift:
At the very least it would be great to have this ***well-documented*** so that no one else has to dig into this like I did to understand what's going on. It's definitely a break from standard matplotlib behavior.

PR #2496, opened 2025-01-28
DOC: Add a raster reprojection example

edsaac:
## Rationale

Plotting a raster image with `ax.imshow` requires data to be warped and reprojected. This does not works well if the map's extent is not set in advance. I am adding an additional example to the documentation to cover that common pitfall: 

- https://github.com/SciTools/cartopy/issues/2480
- https://github.com/SciTools/cartopy/issues/1468

## Implications

This is just an additional example to put in the documentation; no functionality in the package is modified. Below is a screenshot of rebuilding the docs with sphinx:

```sh
sphinx-build -M html docs/source/ docs/build
```

image

### DIFF
--- a/examples/scalar_data/raster_reprojections.py
+++ b/examples/scalar_data/raster_reprojections.py
@@ -5,10 +5,11 @@ Raster reprojections
 When plotting raster data onto a map with `imshow`, we need to first set the
 extend of the map so the reprojection is done correctly.
 
-In this example, we have a raster data stored as a numpy array that is
+In this example, we have some raster data stored as a numpy array that is
 referenced to a rectangular coordinate system. Cartopy reproject these
 data to match the map's coordinate system, however, the map extent
-must be set *before* `imshow` is called.
+must be set *before* `imshow` is called, otherwise, the reprojection
+will not succeed.
 
 """
 
--- a/examples/scalar_data/raster_reprojections.py
+++ b/examples/scalar_data/raster_reprojections.py
@@ -1,0 +1,98 @@
+"""
+Raster reprojections
+====================
+
+When plotting raster data onto a map with `imshow`, we need to first set the
+extend of the map so the reprojection is done correctly.
+
+In this example, we have a raster data stored as a numpy array that is
+referenced to a rectangular coordinate system. Cartopy reproject these
+data to match the map's coordinate system, however, the map extent
+must be set *before* `imshow` is called.
+
+"""
+
+from matplotlib.patches import Rectangle
+import matplotlib.pyplot as plt
+import numpy as np
+
+import cartopy.crs as ccrs
+import cartopy.feature as cf
+
+
+def main():
+    # Generate raster data as a numpy array
+    img = np.linspace(0, 1, 10_000).reshape(100, 100)
+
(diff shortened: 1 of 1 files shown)


examples/scalar_data/raster_reprojections.py
```diff
@@ -0,0 +1,99 @@
+"""
+Raster reprojections
+====================
+
+When plotting raster data onto a map with `imshow`, we need to first set the
+extend of the map so the reprojection is done correctly.
```
greglucas, line 6:
```suggestion
extent of the map so the reprojection is done correctly.
```

examples/scalar_data/raster_reprojections.py
```diff
@@ -0,0 +1,99 @@
...
+
+In this example, we have some raster data stored as a numpy array that is
+referenced to a rectangular coordinate system. Cartopy reproject these
```
greglucas, line 9:
```suggestion
referenced to a rectangular coordinate system (PlateCarree). Cartopy reprojects the
```

examples/scalar_data/raster_reprojections.py
```diff
@@ -0,0 +1,99 @@
...
+data to match the map's coordinate system, however, the map extent
+must be set *before* `imshow` is called, otherwise, the reprojection
+will not succeed.
```
greglucas, line 12:
```suggestion
data to match the map's coordinate system based on the currently set map limits.
This means that the map extent/boundary must be set *before* `imshow` is called.
```

examples/scalar_data/raster_reprojections.py
```diff
@@ -0,0 +1,99 @@
...
+    map_extent = (-88.1, -86.1, 41.2, 42.2)
+    map_proj = ccrs.RotatedPole(pole_longitude=120.0, pole_latitude=70.0)
+
+    fig, axs = plt.subplots(
+        nrows=1,
+        ncols=2,
+        figsize=(12, 5),
+        subplot_kw={"projection": map_proj},
+        sharex=True,
+        sharey=True,
```
greglucas, line 52:
```suggestion
        sharey=True,
        layout="constrained",
```

examples/scalar_data/raster_reprojections.py
```diff
@@ -0,0 +1,99 @@
...
+                facecolor="None",
+                linewidth=3,
+                label="Raster data bounds",
+            )
+        )
+
+        ax.legend()
+        ax.gridlines(draw_labels=True, x_inline=False, dms=True)
+
+    fig.tight_layout()
```
greglucas, line 94:
```suggestion
```

greglucas (commented):
Sorry this as lingered @edsaac! I think this is a good quick example of a possible papercut while using Cartopy and how to avoid it. It would be nice if we could fix the underlying cause and not be dependent on the order of calling `set_extent`, but this is a good first step of highlighting the issue.

I have a few nits on wordsmithing, feel free to modify those even further.

edsaac (commented):
@greglucas, thanks for reviewing!

greglucas (approved):

greglucas:
Thanks @edsaac!

PR #2500, opened 2025-02-17
Fix handling of PosixPath in cartopy.io fh_getter function

jolivetr:
## Rationale

fh_getter fails when fh is a PosixPath because it attempts at checking filename, which has not been initialized.

## Fix

Initialize filename to None

## Implications

Minimal

### DIFF
--- a/lib/cartopy/io/__init__.py
+++ b/lib/cartopy/io/__init__.py
@@ -41,6 +41,8 @@ def fh_getter(fh, mode='r', needs_filename=False):
     if mode != 'r':
         raise ValueError('Only mode "r" currently supported.')
 
+    filename = None
+
     if isinstance(fh, str):
         filename = fh
         fh = open(fh, mode)


jolivetr:
A very tiny fix in fh_getter for cartopy.io.
A variable was tested and not initialized.
Fix: Intitialize the variable

dopplershift (approved):

QuLogic:
Please update the commit message/PR title to be clearer about the change.

jolivetr:
Done. Let me know if this is ok

jolivetr:
> I like the fix, I'm wondering if we can add a test. What were you doing to trigger this error in the first place?

I am not sure. I am using cartopy to plot SRTM data. I think this error shows up with recent versions of Python (12 and 13 for sure). It was not failing with 3.8. 

Anyway, the variable filename was not initialized in the first place, so that should fix it.

Issue #1498, opened 2020-03-27
align a cartopy 2D map plot with a 1D line plot


miniufo:
I need to align a cartopy 2D map subplot with a 1D line subplot (zonal-mean or meridional mean of the 2D data), using the following code:
```python
import xarray as xr
import cartopy.crs as ccrs
import matplotlib.pyplot as plt


air = xr.tutorial.open_dataset('air_temperature').air[0]

fig = plt.figure(figsize=(14,5))

ax1 = fig.add_subplot(121,
                 projection=ccrs.PlateCarree(central_longitude=180))
air.plot.pcolormesh(ax=ax1,
               transform=ccrs.PlateCarree(),
               cbar_kwargs={'orientation':'horizontal'})
ax1.coastlines()
ax1.set_extent((air.lon[0], air.lon[-1], air.lat[-1], air.lat[0]),
               crs=ccrs.PlateCarree())

ax2 = fig.add_subplot(122)
air.mean('lon').plot(ax=ax2, y='lat')

plt.tight_layout()
```
This gives [the plot as this](https://i.stack.imgur.com/5y0Dv.png).

Both subplots have the same y- (latitude-) range and I expect the boxes of the two axes could align with each other.  Is there a simple way to do this?  This has been proposed on [stackoverflow](https://stackoverflow.com/questions/60807254/align-a-cartopy-2d-map-plot-with-a-1d-line-plot).

QuLogic:
I have not tested, but I would guess that you would have to set the aspect ratio of the plain Axes to something that would match the Cartopy Axes.

trexfeathers:
[This example in the Iris gallery](https://scitools-iris.readthedocs.io/en/stable/generated/gallery/general/plot_zonal_means.html) demonstrates how Cartopy and Matplotlib can be used to achieve this.

rcomer:
I would use compressed layout rather than tight layout.  I added an answer on StackOverflow.

miniufo:
Thanks for your suggestions.  I've already got this resolved using `proplot` as a workaround.

However, if one uses a projection (e.g., `Robin`) other than `PlateCarree`, then the latitudes would be slightly stretched.  In this case can we still have a equivalent zonal-mean plot aligned with the projection?

trexfeathers:
Just tested with `Mollweide` and this is indeed a problem

heroldn:
[...]

PR #2502, opened 2025-02-24
DOC: add example aligning Cartopy and Matplotlib axes

rcomer:
## Rationale

Added a new gallery example adapted from https://github.com/SciTools/cartopy/issues/1498#issuecomment-2672281574.  I switched to using `inset_axes` as that reliably vertically aligns the two axes, which didn't always happen with the subplots approach.  Also re-ordered the code a bit and reduced the number of projections.  I have no strong opinion on how many and which projections to include, although I couldn't figure out how to get the right numbers of latitude lines for the Orthographic case without manually specifying them, and I'd rather not get into that kind of detail.

I'm not 100% sure which category this belongs under.  For now it is under "gridlines and labels" because they are part of what we are lining up.

### DIFF
--- a/examples/gridlines_and_labels/map_latitudes.py
+++ b/examples/gridlines_and_labels/map_latitudes.py
@@ -1,0 +1,82 @@
+"""
+Aligning Cartopy and Matplotlib axes
+====================================
+
+Sometimes it is useful to line up a standard Matplotlib axes with a Cartopy map, for
+example when plotting zonal mean data next to a global contour plot.  This example
+demonstrates one way to do that for projections where latitude lines are horizontal.
+
+"""
+
+# sphinx_gallery_multi_image = "single"
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import cartopy.crs as ccrs
+
+
+PC_PROJ = ccrs.PlateCarree()
+
+def map_and_latitudes(map_proj):
+    """
+    Create a map with the given projection, together with a rectangular axes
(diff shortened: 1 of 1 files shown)


rcomer:
[Renders like this](https://output.circle-artifacts.com/output/job/3b4917a6-e072-4cc4-a1a6-ed7356a9db8c/artifacts/0/docs/build/html/gallery/gridlines_and_labels/map_latitudes.html#sphx-glr-gallery-gridlines-and-labels-map-latitudes-py)

greglucas:
One thing that might be nice is to use the data from one of the scalar_data examples and plot what the zonal mean is of that actual function. Would it be worth adding a meridional mean below the axes as well, or would that add too much complexity with multiple to/from functions then?

greglucas:
@rcomer, I see you moved this to draft. Do you still want to add this in? I think we should add more helpful examples like this!

rcomer:
Thanks @greglucas yes, I had intended to follow up on your previous comment, but creative juices are not really flowing at the moment.  I'm not sure when I'm going to get to it.

greglucas:
:+1: Sounds good. I am personally happy with the way this is now, and don't think we need to get hung up on getting everything perfect / implemented right away. So I'd be fine merging this and then when your creativity strikes again updating this example or adding another one.

greglucas:
Thanks @rcomer! We can follow up with further improvement as inspiration strikes, no rush here :)

PR #2504, opened 2025-02-27
ENH: Azimuthal rotation for Orthographic projection

kenhira:
## Rationale

This update adds `azimuth` argument to `ccrs.Orthographic()`, with a corresponding test. It will enable users to rotate the projection around the central point. It uses new Proj parameter `alpha`, which was added to `Orthographic` in Proj version 9.5.0. 
https://proj.org/en/stable/operations/projections/ortho.html

An equivalent `azimuth` argument has already been implemented to `ccrs.ObliqueMercator()`.

## Implications

Users will become able to rotate the map as they prefer. For example, one can orient a satellite image along the solar direction to emphasize which direction the sun glint and shadow appear.
This update is a minimal update, adding just one argument to the function.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -2093,9 +2093,20 @@ class Orthographic(Projection):
     _handles_ellipses = False
 
     def __init__(self, central_longitude=0.0, central_latitude=0.0,
-                 azimuth=0.0, globe=None):
-        proj4_params = [('proj', 'ortho'), ('lon_0', central_longitude),
-                        ('lat_0', central_latitude), ('alpha', azimuth)]
+                 azimuth=None, globe=None):
+        if pyproj.__proj_version__ >= '9.5.0':
+            if azimuth is None:
+                azimuth = 0.0
+            proj4_params = [('proj', 'ortho'), ('lon_0', central_longitude),
+                            ('lat_0', central_latitude), ('alpha', azimuth)]
+        else:
+            proj4_params = [('proj', 'ortho'), ('lon_0', central_longitude),
+                            ('lat_0', central_latitude)]
+            if azimuth is not None:
+                warnings.warn(
+                    'Setting azimuth is not supported with PROJ versions < 9.5.0. '
+                    'Assuming azimuth=0.'
+                    'Current PROJ version: %s' % pyproj.__proj_version__)
         super().__init__(proj4_params, globe=globe)
 
         # TODO: Let the globe return the semimajor axis always.
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -2093,16 +2093,14 @@ class Orthographic(Projection):
     _handles_ellipses = False
 
     def __init__(self, central_longitude=0.0, central_latitude=0.0,
-                 azimuth=None, globe=None):
+                 azimuth=0.0, globe=None):
         if pyproj.__proj_version__ >= '9.5.0':
-            if azimuth is None:
-                azimuth = 0.0
             proj4_params = [('proj', 'ortho'), ('lon_0', central_longitude),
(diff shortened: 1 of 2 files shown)


greglucas:
I'm guessing you'll need to put a version gate around this parameter so that we only use it if the underlying PROJ library >= 9.5. Here is an example in our tests

https://github.com/SciTools/cartopy/blob/0c51cbb269c7909ab283e62978fe47cb59512e0c/lib/cartopy/tests/test_crs.py#L95

What happens if we pass this parameter to the projection in earlier versions, do we get a warning or do we need to issue a warning on our side that the parameter has no effect and will be ignored?

greglucas:
ping @kenhira , are you able to add the version gate on this parameter?

kenhira:
Apologies for my slow response. I have implemented the version gate (`pyproj.__proj_version__ >= '9.5.0'`). 
> What happens if we pass this parameter to the projection in earlier versions, do we get a warning or do we need to issue a warning on our side that the parameter has no effect and will be ignored?

I tested on my end with `pyproj=3.4.1` and `proj=9.1.1`, and it seems that we do not get a warning by passing `alpha` parameter to proj even though it is not an expected input. Therefore I added a warning on cartopy side that the azimuth input will be ignored.

kenhira:
Here is what the output is like:
With the recent proj version, the azimuthal rotation works.
test_plot_9 5 1

However, with the older proj (tested with `proj=9.1.1`), the rotation is not effective. With my implementation, the user will get the following warning
```
.../cartopy/lib/cartopy/crs.py:2100: UserWarning: Setting azimuth is not supported with PROJ versions < 9.5.0. Assuming azimuth=0. Current PROJ version: 9.1.1
  warnings.warn(
```
test_plot_9 1 1

Here is the test code used for plots above:
```python
import cartopy.crs as ccrs
from cartopy.geodesic import Geodesic

if __name__ == "__main__":
    anchorage_lon, anchorage_lat = -149.9003, 61.2181
    denver_lon, denver_lat = -104.9903, 39.7392
    geod = Geodesic()
    heading = geod.inverse((anchorage_lon, anchorage_lat), (denver_lon, denver_lat))[0][1]
    
    fig = plt.figure(figsize=(8, 4))

    ax1 = fig.add_subplot(1, 2, 1, 
        projection=ccrs.Orthographic(central_longitude=anchorage_lon, 
                                     central_latitude=anchorage_lat))
    ax1.coastlines()
    ax1.set_global()
    ax1.plot([anchorage_lon, denver_lon], [anchorage_lat, denver_lat], 
             transform=ccrs.Geodetic(), label="Anchorage to Denver")
    ax1.plot(anchorage_lon, anchorage_lat, 'o', transform=ccrs.Geodetic(), label="Anchorage")
    ax1.plot(denver_lon, denver_lat, 'o', transform=ccrs.Geodetic(), label="Denver")
    ax1.set_title("azimuth=0")

    ax2 = fig.add_subplot(1, 2, 2,
        projection=ccrs.Orthographic(central_longitude=anchorage_lon,
                                     central_latitude=anchorage_lat,
                                     azimuth=heading))
    ax2.coastlines()
    ax2.set_global()
    ax2.plot([anchorage_lon, denver_lon], [anchorage_lat, denver_lat], 
             transform=ccrs.Geodetic())
    ax2.plot(anchorage_lon, anchorage_lat, 'o', transform=ccrs.Geodetic(), label="Anchorage")
    ax2.plot(denver_lon, denver_lat, 'o', transform=ccrs.Geodetic(), label="Denver")
    ax2.set_title("azimuth=%.2f deg" % heading)
    ax2.legend(loc='lower right')

    fig.tight_layout()
    plt.savefig('test_plot.png')
    plt.show()
```

greglucas (approved):

greglucas:
Thanks @kenhira! That all looks good to me.

PR #1686, opened 2020-11-30
add python 3.9 support

eumiro:
see title

PR #2508, opened 2025-03-09
Use pathlib.Path's read and write methods

eumiro:
## Rationale

Code refactoring: use `pathlib.Path`'s read and write methods.

## Implications

You're already using `pathlib.Path` objects everywhere in file manipulation. With this changes, the code should be more straightforward to read. No implications on the functionality.


eumiro:
Signed CLA in #1686

greglucas (approved):
Looks good, thanks @eumiro!

Issue #2514, opened 2025-03-28
Using Cartopy with JupyterLite and handling data downloads


rpwagner:
Hi,

I'm trying to use Cartopy within a [JupyterLite](https://jupyterlite.readthedocs.io/en/stable/) notebook and I'm running into an issue with downloads still happening in the background while the plots are still being generated. I haven't found a place or configuration to block until the downloads complete, such as using an `await` statement.

I'm using the [simple maps example](https://scitools.org.uk/cartopy/docs/latest/matplotlib/intro.html#beautifully-simple-maps) of how to use Cartopy with Matplotib
```
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

ax = plt.axes(projection=ccrs.PlateCarree())
ax.coastlines()

# Save the plot by calling plt.savefig() BEFORE plt.show()
plt.savefig('coastlines.pdf')
plt.savefig('coastlines.png')

plt.show()
```

When I run this, I first receive a `DownloadWarning` 
```
/lib/python3.12/site-packages/cartopy/io/__init__.py:241: DownloadWarning: Downloading: https://naturalearth.s3.amazonaws.com/110m_physical/ne_110m_coastline.zip
  warnings.warn(f'Downloading: {url}', DownloadWarning)
```

This then produces the `BlockingIOError` error (full error at the bottom):
```
Error in callback <function _draw_all_if_interactive at 0x677fab8> (for post_execute), with arguments args (),kwargs {}:
...
URLError: <urlopen error [Errno 26] Operation in progress>
```

Is there a workaround or way to block on the IO, or to prefetch the data? I recognize this could be difficult in the JupyterLite context. If needed, I could prestage some of the data when building the JupyterLite site, but this limits the types of plots users can make when doing exploration.

Thanks

For reference, here are the versions of the packages in this environment:
```
Pyodide version: 0.27.3
Pyodide kernel version: 0.5.2
Cartopy version: 0.24.1
Matplotlib version: 3.8.4
```

Full error:
```
---------------------------------------------------------------------------
BlockingIOError                           Traceback (most recent call last)
File /lib/python312.zip/urllib/request.py:1344, in AbstractHTTPHandler.do_open(self, http_class, req, **http_conn_args)
   1343 try:
-> 1344     h.request(req.get_method(), req.selector, req.data, headers,
   1345               encode_chunked=req.has_header('Transfer-encoding'))
   1346 except OSError as err: # timeout error

[...]

PR #2515, opened 2025-03-29
Replace urllib.request with urllib3 to support WebAssembly

rpwagner:
As discussed in #2514, `urllib.request` is [not supported in WebAssembly](https://docs.python.org/3/library/urllib.request.html). This replaces `urllib.request` where needed with `urllib3` which does work with WebAssembly. I've tried to keep the code changes to a minimum and to reuse existing semantic patterns.

The linter did introduce new lines in one of the imports. I found it odd but left it in place.

## Rationale

This enables the use of Cartopy is [Pyodide](http://pyodide.org/) and [JupyterLite](http://jupyterlite.readthedocs.io/). Without it, downloading of shape files, etc., fail. See #2514.

## Implications

I've run the unit tests and nothing new fails. It's also been tested in JupyterLite and works. There is a chance that the download code path hasn't been fully tested and something may still need to be changed to support `urllib3`.

This also introduces a new dependency, even though `urllib3` is widely available. It is also an existing package in Pyodide.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,6 +14,7 @@ dependencies:
   - pyshp>=2.3
   - pyproj>=3.3.1
   - packaging>=21
+  - urllib3>=2.3
   # The testing label has the proper version of freetype included
   - conda-forge/label/testing::matplotlib-base>=3.6
 
--- a/lib/cartopy/io/__init__.py
+++ b/lib/cartopy/io/__init__.py
@@ -228,6 +228,7 @@ class Downloader:
         response = self._urlopen(url)
 
         target_path.write_bytes(response.read())
+        response.release_conn()
 
         return target_path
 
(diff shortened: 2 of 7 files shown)


greglucas:
It is still open, but needs a champion that is interested in merging it in. I'm not in favor of it because it adds a new dependency that seems somewhat unncessary except for a WASM environment, but I won't block if someone else thinks this is the way to go.

See the comment here and the thread for workarounds/suggestions on how to approach this: https://github.com/SciTools/cartopy/issues/2514#issuecomment-2763101149

MMesch:
The workaround is reasonably easy and works for me. Thanks for pointing to it.

On the other hand, "except for a WASM environment" seems like a very good reason to me. Especially considering that WASM is a pretty universal standard these days rapidly gaining popularity through platforms such as JupyterLite. But I'm not neutral.

dopplershift (approved):
I've come around in favor of this approach for a few reasons:
* I'm interested in WASM
* We're already depending on requests conditionally through owslib
* urllib3 being a dependency of requests makes it 99% likely it's already installed in someone's environment and not a risky add.

I'll wait to merge a bit for any impassioned counterpoints, but IMO this is a reasonable merge.

Issue #2176, opened 2023-05-08
TypeError: 'GeometryCollection' object is not subscriptable (cartopy calling shapely library)


durack1:
### Description
I have hit a repeatable, but confusing error with ~20 timesteps in a long ~1800 timestep global array. The error appears to be sensitive to the levels specified in the optional `levels` argument, and is reproduced below. It triggers a `GeometryCollection` error in the shapely library call (see traceback below).

I have included a working example below, which loops through two timesteps, the first works great plotting cleanly, and the second bombs, triggering the `GeometryCollection` bug. The error continues on subsequent timesteps, after ~20 or so, it starts to work again.

I would appreciate any pointers to resolving this issue, or tips at identifying a workaround to get me plotting through the problem ~20 timesteps.

#### Code to reproduce

```python
# %% imports
import cartopy.crs as ccrs
import matplotlib.pyplot as plt
import numpy as np
from xarray import open_dataset

# %% load dataset
f = "cartopyLevelBugTwoTimesteps.nc"  # (file will be dropped below)
ds = open_dataset(f)
da = ds.siconc

# %% setup plot
levs1 = list(np.arange(-20, 121, 10))  # siconc
centralLon = 202

# %% step through working and failing timesteps
for time in np.arange(0, 2):
    print("Step:", time)
    ar = da.isel(time=time)
    fig = plt.figure(figsize=(10, 15))

    # Start subplots
    ax1 = fig.add_subplot(3, 1, 1,
                          projection=ccrs.Robinson(
                              central_longitude=centralLon,
                              globe=None,
                              false_easting=None,
                              false_northing=None,
                          ),
                          )
    cs1 = ax1.contourf(da.lon.data, da.lat.data, ar.data,
                       levs1,
                       transform=ccrs.PlateCarree(),
                       cmap="RdBu_r"
                       )
    plt.show()
    # time = 1; Should bomb with "TypeError: 'GeometryCollection' object is not subscriptable"

```
#### Data to reproduce
[cartopyLevelBugTwoTimesteps.nc.gz](https://github.com/SciTools/cartopy/files/11425547/cartopyLevelBugTwoTimesteps.nc.gz)

#### Traceback
```python
Traceback (most recent call last):
  File ~/mambaforge/envs/xcd050xes071spy543mat371car0211ipy8132/lib/python3.10/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)
[...]

PR #2521, opened 2025-04-23
FIX: make hole multipolygon valid before removing from projected polygon

rcomer:
## Rationale

When a polygon with multiple holes is projected, the projected holes may become so close that they can't form a valid MultiPolygon as they [touch along a line](https://shapely.readthedocs.io/en/stable/manual.html#MultiPolygon).  In this case, we can still create the holes one by one.  Taking the first example from https://github.com/SciTools/cartopy/issues/2176#issuecomment-2772601094, and adding coastlines for reference, 

```python
import cartopy.crs as ccrs

crs_proj = ccrs.LambertAzimuthalEqualArea(central_latitude=-90)
fig, ax = plt.subplots(1, 1, subplot_kw={"projection": crs_proj})
ax.add_feature(cfeature.OCEAN)
ax.coastlines()
plt.show()
```
I now get
image

Setting the central latitude to -30 still results in a failure, but this is very sensitive to the exact number and the user will be able to work around it by setting central latitude to e.g. -29.9999:

image

## Implications

Making the holes one by one is slow compared with using the MultiPolygon:  I tried removing the try-except and always removing them in the loop.  This benchmark took 18 times longer:

```python
import cartopy.crs as ccrs


def ocean_fig():
    fig = plt.figure()
    ax = fig.add_subplot(111, projection=ccrs.NorthPolarStereo())

    ax.add_feature(cfeature.OCEAN.with_scale('10m'))
    
    return fig


import timeit

setup = """
from __main__ import ocean_fig
fig = ocean_fig()
"""

print(timeit.timeit('fig.draw_without_rendering()', setup=setup, number=5))  
```

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1251,7 +1251,8 @@ class Projection(CRS, metaclass=ABCMeta):
             box = sgeom.box(x3, y3, x4, y4, ccw=is_ccw)
 
             # Invert the polygons
-            polygon = box.difference(sgeom.MultiPolygon(interior_polys))
+            multi_poly = shapely.make_valid(sgeom.MultiPolygon(interior_polys))
+            polygon = box.difference(multi_poly)
 
             # Intersect the inverted polygon with the boundary
             polygon = boundary_poly.intersection(polygon)
--- a/lib/cartopy/tests/test_polygon.py
+++ b/lib/cartopy/tests/test_polygon.py
@@ -457,6 +457,14 @@ class TestHoles(PolygonTests):
         assert len(multi_polygon.geoms) == 1
         assert len(multi_polygon.geoms[0].interiors) >= 2
 
+    def test_inverted_poly_merged_holes(self):
+        proj = ccrs.LambertAzimuthalEqualArea(central_latitude=-90)
+        pc = ccrs.PlateCarree()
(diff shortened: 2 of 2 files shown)


rcomer:
CI failures are in `test_wfs_france` so not my fault!

lib/cartopy/crs.py
```diff
@@ -1251,7 +1251,14 @@ def _rings_to_multi_polygon(self, rings, is_ccw):
             box = sgeom.box(x3, y3, x4, y4, ccw=is_ccw)
 
             # Invert the polygons
-            polygon = box.difference(sgeom.MultiPolygon(interior_polys))
+            try:
+                polygon = box.difference(sgeom.MultiPolygon(interior_polys))
```
greglucas, line 1255: Is it possible to call `make_valid()` before doing the `box.difference()` call here? Or would that be just as slow potentially...

greglucas (approved):
Seems reasonable to change something that is broken for something that is slow and works :)

greglucas:
> CI failures are in test_wfs_france so not my fault!

I noticed that on main too. We should probably xfail that for now.

lib/cartopy/crs.py
rcomer: Actually that seems to take no time at all.  Let's do that :+1:

greglucas (approved):

PR #2522, opened 2025-04-25
TST: xfail France test due to data server issues

greglucas:
This test is failing due to a timeout with the server. I am not sure if the server has changed locations and needs a new address or if it is deprecated now, but this is the quickest fix to get tests passing again.

@mcuntz do you have any insight into this server since you added the test?

### DIFF
--- a/lib/cartopy/tests/mpl/test_features.py
+++ b/lib/cartopy/tests/mpl/test_features.py
@@ -83,6 +83,7 @@ def test_wfs():
                     reason='OWSLib and at least one of pykdtree or scipy is required')
 @pytest.mark.xfail(raises=(ParseError, AttributeError),
                    reason="Bad XML returned from the URL")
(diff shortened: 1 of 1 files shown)


rcomer (approved):

rcomer:
Any idea what the "build and upload to PyPI" failure is about?

Merging anyway since that can be handled separately, and it is obviously useful to get pytest passing.

greglucas:
> Any idea what the "build and upload to PyPI" failure is about?

No, I've noticed that job gets cancelled due to a higher priority job a lot, but I don't know what events are causing multiple jobs to trigger. Maybe we should just remove that piece and let all jobs proceed. They get skipped after ~1-2 seconds.
https://github.com/SciTools/cartopy/blob/ebdd75ab8205371f0234b0b39e835fbfba37936b/.github/workflows/release.yml#L6

PR #2526, opened 2025-04-30
pep639: setuptools license and license-files fields

bjlittle:
This pull-request mops up some tech debt to comply with [PEP639](https://peps.python.org/pep-0639/) for `setuptools`, also see [license and license-files](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files).

Additionally avoids the following `setuptools` deprecation warning:
```
!!
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:
          License :: OSI Approved :: BSD License
          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
!!
```
See [Deprecated license classifiers](https://peps.python.org/pep-0639/#deprecate-license-classifiers).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools >= 40.6.0",
+    "setuptools >= 77.0.3",
     "Cython >= 0.29.24",
     # numpy requirement for wheel builds for distribution on PyPI - building
     # against 2.x yields wheels that are also compatible with numpy 1.x at
@@ -10,7 +10,7 @@ requires = [
     # redistributors can do this by installing the numpy version they like and
     # disabling build isolation.
     "numpy>=2.0.0rc1",
-    "setuptools_scm >= 7.0.0",
+    "setuptools_scm >= 8.0.0",
 ]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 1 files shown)


greglucas:
I've run into issues on other projects with uploads to pypi failing due to bad metadata when I've bumped this. Do you know how we can fix that? (I have either pinned to a lower setuptools, or set the flag to ignore the metadata check on upload)

Checking out this branch and building with `python -m build .` and then running `twine check dist/*` I get the following error:
```
Checking dist/cartopy-0.24.2.dev70+g5fc8006f-cp312-cp312-macosx_10_9_universal2.whl: ERROR    InvalidDistribution: Invalid distribution metadata: unrecognized or malformed field 'license-expression';
         unrecognized or malformed field 'license-file'
```

pyproject.toml
```diff
@@ -10,7 +10,7 @@ requires = [
     # redistributors can do this by installing the numpy version they like and
     # disabling build isolation.
     "numpy>=2.0.0rc1",
```
greglucas, line 12: We should probably bump this to be `>=2.0.0` without the rc option.

bjlittle:
> Checking out this branch and building with `python -m build .` and then running `twine check dist/*` I get the following error:

@greglucas I can't replicate this. Checking out this branch, building the wheel and then checking with `twine` "just works".

In my environment I've got the following top-level packages:

```
# Name                    Version                   Build  Channel
build                     1.2.2.post1              pypi_0    pypi
pip                       25.1               pyh8b19718_0    conda-forge
twine                     6.1.0              pyh29332c3_0    conda-forge
python                    3.12.10         h9e4cc4f_0_cpython    conda-forge
setuptools                80.1.0             pyhff2d567_0    conda-forge
```

pyproject.toml
bjlittle: Done :+1:

greglucas (approved):
Interesting... I am not sure what happened on my end when testing. Sorry for the noise there! I just created a new environment and it looks good to me now too. Thanks for following up Bill!

Issue #1376, opened 2019-10-05
Projection request: Spilhaus projection (world ocean map)


AJueling:
The Spilhaus projection displays the world ocean as a continuous water mass around Antarctica with the other (highly distorted) continents acting as the map's boundary. I think it would be a great addition to cartopy's functionality especially for oceanographers.

Here is a nice write up of the history of the projection with potential starting points for the implementation: [ArcGIS storymap on Spilhaus projection](https://storymaps.arcgis.com/stories/756bcae18d304a1eac140f19f4d5cb3d)

![Spilhaus projection image from https://storymaps.arcgis.com/stories/756bcae18d304a1eac140f19f4d5cb3d](https://user-images.githubusercontent.com/20947797/66254702-50416480-e77a-11e9-9e78-ac26ce996fbd.png)
(from aforementioned ArcGIS link)

dopplershift:
CartoPy relies on PROJ extensively, so all of this is going to largely depend on what's in PROJ.

AJueling:
Thank you, that's what I thought after looking through the code. If I opened a issue with PROJ, would cartopy need the rotated projection (with the square standing on its side, as shown above) in addition to the original _Adams World in a Square II_ projection (where the square rests on a corner)? Spilhaus is the same projection with different parameters (e.g. the poles are in Asia and  South America) and that is subsequently rotated.
<img width="609" alt="Screenshot 2019-10-08 at 12 46 05" src="https://user-images.githubusercontent.com/20947797/66390156-5af23880-e9ca-11e9-842e-e452305dce40.png">
(from page 95 of Snyder, J. P. and Voxland, P. M. (1989). _An Album of Map Projections._ https://doi.org/10.3133/pp1453)

dopplershift:
Yes, the rotation would need to be handled by PROJ.

AJueling:
There is an open issue over at PROJ: OSGeo/PROJ#1851

mdaeron:
Based on [this comment](https://github.com/OSGeo/PROJ/issues/1851#issuecomment-1868220725) there seems to be a Python implementation [here](https://github.com/rtlemos/spilhaus/tree/main). Would this be a suitable starting point to get this projection added to cartopy? (Pale)oceanographers would rejoice :)

dopplershift:
[...]

PR #2529, opened 2025-05-10
Spilhaus projection 

MaceKuailv:
## Rationale

Implementing a cool-looking map projection for global ocean called the Spilhaus Adams world in a square II map projection. 
This is in response to projection request #1376 
## Example

```
import cartopy.crs as ccrs

fig = plt.figure(figsize=(8, 8))

rots = [-45,45,-135,135]
for i in range(4):
    ax = plt.subplot(2,2,i+1,projection = ccrs.Spilhaus(rotation=rots[i]))
    ax.add_feature(cfeature.OCEAN,color = 'royalblue')
#     ax.axis('off')
    
plt.tight_layout()
plt.subplots_adjust(wspace=0.0, hspace=0.0)
```
image

Many thanks to @greglucas.


MaceKuailv:
Some notes:

1. I have tried both wkt and proj4 initialization, they creates the identical map projection, and they are both very slow (about 30 seconds).
2. The boundary listed [here](https://epsg.io/54099) is very far off. That boundary is a tangential line to 180. I then use some optimization algorithm to find the actual boundary. The meta pole is not exactly 115E 30N, but very slightly off [115.00000024,  30.00000036]. The projected domain is also not a perfect square, and the aspect ratio is off by about 1/10,000. 
3. The tests are all failing because the map projection is too new. All of the tests pass on my local machine with PROJ 9.6.0 and pyproj 3.7.1.

MaceKuailv:
https://github.com/SciTools/cartopy/blob/1e1583334587ae16644ebb31f1b311b2ef36f23f/.github/workflows/ci-testing.yml#L43C9-L45C1
I think this minimum requirement basically means the new tests I wrote can never pass, and if I use "skipif" the pytest simply would never run.

rcomer:
The minimum version specification only affects one of the test runners.  The others should use the latest pyproj from PyPI, though it looks like that is [older than Proj 9.6](https://pypi.org/project/pyproj/#history).  So maybe we just need to wait for a pyproj release.  Did you create your local environment with conda?  [Conda-forge does its own re-building](https://github.com/conda-forge/pyproj-feedstock/pull/158) independent of the package maintainers.

MaceKuailv:
@rcomer Thanks for clarifying. Yes, I guess we will have to wait for a pip release from pyproj. Yes, I built a new environment using `conda env create -f environment.yml.

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,59 @@ def x_limits(self):
     def y_limits(self):
         return self._y_limits
 
+class Spilhaus(Projection):
+    """
+    Spihaus-Adams World Ocean Map in a Square II
```
greglucas, line 3224: Small typo missing the L in Spilhaus. Where does the II come from? Is there a good reference for this, or link out to the PROJ documentation?

https://proj.org/en/stable/operations/projections/spilhaus.html

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,59 @@ def x_limits(self):
...
+
+    """
+    def __init__(self,greenland_at = 0,false_easting = 0.0,false_northing = 0.0):
```
greglucas, line 3227: We don't have any other projections with "location_at" I don't think. I don't think we want that since it doesn't really have a direct relevance to the underlying projection. I think you can expose the `rotation=45` parameter here and pipe that through for rotation directly and let users handle the "greenland_at" logic with a wrapper function if they want that.

Do we want `central_longitude` and `central_latitude` to be exposed here too? They have different default values than typical though, so we could just leave them as default and not let the user have the option.

Note the formatting/linting of the arguments here too, there should be a space after a comma.

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,59 @@ def x_limits(self):
...
+        if isinstance(greenland_at, int) and (0<=greenland_at<4):
+            rot_spil = 45+(greenland_at%4)*90
+        else:
+            raise ValueError('The orientation of Greenland must be one'
+                ' of {allowed_ori}, or integer [0,1,2,3]')
+        proj4_params = [('proj', 'spilhaus'),
+                        ('lat_0', -49.56371678),
+                        ('lon_0', 66.94970198),
+                        ('azi',40.17823482),
+                        ('k_0', 1.4142135623730951),
```
greglucas, line 3256: We shouldn't hard-code any of the defaults being passed into PROJ, if they choose to change those defaults later I think we'd want to just go along with them. So in the proj4_params we should only pass in the values that the user has the options to change.

lib/cartopy/tests/crs/helpers.py
```diff
@@ -11,4 +11,4 @@
 def check_proj_params(name, crs, other_args):
     expected = other_args | {f'proj={name}', 'no_defs'}
     proj_params = set(crs.proj4_init.lstrip('+').split(' +'))
-    assert expected == proj_params
+    assert expected == proj_params, expected.symmetric_difference(proj_params)
```
greglucas, line 14: I think this might be less informative because it would just print the string of the values that were different but you wouldn't know which one was in the left vs right?

lib/cartopy/tests/crs/test_spilhaus.py
```diff
@@ -0,0 +1,52 @@
...
+    'datum=WGS84',
+    'ellps=WGS84',
+    'k_0=1.414213562373095',
+    'lat_0=-49.56371678',
+    'lon_0=66.94970198',
+    'no_defs',
+    'units=m',
+}
+@pytest.mark.skipif(
+    parse_version(pyproj.__version__) <= parse_version("3.7.0"),
```
greglucas, line 27: I think this is actually dependent on the PROJ version like you mentioned, so can we get that out of pyproj somehow instead? i.e. I think I can build 3.7.0 with PROJ 9.6 and this would still work.

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,59 @@ def x_limits(self):
...
+    """
+    def __init__(self,greenland_at = 0,false_easting = 0.0,false_northing = 0.0):
+        """
+        Parameters
+        ----------
+        greenland_at: optional
+            Orientation of the map. 0,1,2,3, stand for Greenland at upper left,
+            upper right, lower right and lower left, respecivly. Defaults to 0.
+            Strings like 'upper left' are also allowed.
+        false_easting: optional
```
greglucas, line 3235: There should be a space before and after the ":" in these docstrings for sphinx to format it properly.
```suggestion
        false_easting : optional
```

greglucas (commented):
Nice! Glad you're having fun with this one :) I've got a few suggestions to mostly make this simpler on the Cartopy maintenance side and then users can extend it later with the extra names if they want.

lib/cartopy/crs.py
MaceKuailv: Yes, it is essentially this map projection with a different pole https://proj.org/en/stable/operations/projections/adams_ws2.html. I think it will be clearer if I just remove the "Adams" and the "II".

lib/cartopy/crs.py
MaceKuailv: I have now fully exposed the rotation parameter. I don't entirely understand why exposing central_lat lon "not let the user have the option", but I have also exposed those two variables.

lib/cartopy/tests/crs/helpers.py
MaceKuailv: Sorry, I forgot to remove this...

lib/cartopy/tests/crs/test_spilhaus.py
MaceKuailv: Yeah, this is now implemented. I haven't try older version of pyproj, I will remove the condition and try again.

lib/cartopy/crs.py
MaceKuailv: I ended up deciding not to expose central_longitude and central_latitude.

lib/cartopy/tests/crs/test_spilhaus.py
MaceKuailv: You are right. Even old version of pyproj would work.

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,42 @@ def x_limits(self):
     def y_limits(self):
         return self._y_limits
 
+class Spilhaus(Projection):
+    """
+    Spilhaus World Ocean Map in a Square
+
+    """
+    def __init__(self,rotation = 45, false_easting = 0.0, false_northing = 0.0):
```
greglucas, line 3227: The other classes in this module use a default `globe=None` which then allows the datum to be overridden.

It also looks like every other class has a `central_longitude` parameter as well, but it is always 0 in the other projections and would be 66.94970198 here to correspond to the default lon0 in PROJ. So I'm not sure whether we want to expose that parameter or not for consistency. I can go either way on this one, so fine with how you have it for now.

```suggestion
    def __init__(self, rotation=45, false_easting=0.0, false_northing=0.0, globe=None):
```

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,42 @@ def x_limits(self):
...
+        false_northing : optional
+            Y offset from the planar origin in metres. Defaults to 0.0.
+
+        """
+        proj4_params = [('proj', 'spilhaus'),
+                        ('rot',rotation),
+                        ('x_0', false_easting),
+                        ('y_0', false_northing),
+                        ('datum', 'WGS84'),
+                        ('units', 'm')]
```
greglucas, line 3244:
```suggestion
                        ('y_0', false_northing)]
```

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,42 @@ def x_limits(self):
...
+
+        super().__init__(proj4_params)
```
greglucas, line 3246:
```suggestion
        super().__init__(proj4_params, globe=globe)
```

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,42 @@ def x_limits(self):
...
+        # The boundary on https://epsg.io/54099 are wrong
+        # The following bounds are calculated based on
+        #[-65.00000012, -29.99999981]
+        # and [115.00000024,  30.00000036]
+        self.bounds = [
+            -11802684.083372328,
+            11802683.949222516,
+            -11801129.925928915,
+            11801129.925928915
+        ]
```
greglucas, line 3256: Should these be an actual square and therefore the same value in all positions?

Is there a pyproj way to get these exactly with `pyproj_crs.area_of_use.bounds`?

lib/cartopy/tests/crs/test_spilhaus.py
```diff
@@ -0,0 +1,40 @@
...
+proj_version = parse_version(pyproj.proj_version_str)
+common_arg = {
+    'datum=WGS84',
+    'ellps=WGS84',
+    'no_defs',
+    'units=m',
+}
+@pytest.mark.skipif(
+    (proj_version < parse_version("9.6.0")),
+    reason="Requires pyproj > 3.7.0"
```
greglucas, line 25: I think you could add this mark at the module level since all of these tests require PROJ>9.6.
https://docs.pytest.org/en/stable/example/markers.html

The text of the reason needs to be updated to PROJ specific in all of the reasons.
```suggestion
    reason="Requires PROJ >= 9.6.0"
```

lib/cartopy/tests/crs/test_spilhaus.py
```diff
@@ -0,0 +1,40 @@
+# Copyright Crown and Cartopy Contributors
+#
+# This file is part of Cartopy and is released under the BSD 3-clause license.
+# See LICENSE in the root of the repository for full licensing details.
+"""
+Tests for the HEALPix projection.
```
greglucas, line 6:
```suggestion
Tests for the Spilhaus projection.
```

lib/cartopy/crs.py
MaceKuailv: There are some linter that can change this kind of formatting. I can add it to .pre-commit-config (not sure if I am spelling this correctly) if you want. 

I will add the globe argument, sorry for the ommision.

lib/cartopy/crs.py
MaceKuailv: Well, they should be a actual square, but it is off by about 0.01%. I think this slight inaccuracy stem from upstream. The feature I love the most about this projection is that it is a square and can be used to tile the entire surface. 

The Spilhaus projection is a global projection, but (-180,-90) and (180,90) are not at the corners. The transformed bounds defined [here](https://epsg.io/54099) also does not produce the correct result some how...Theoretically the points furthest away from the center should be (115,30) but it is actually slightly off. My strategy was to use scipy minimize to find the actual pole, but I don't think it make sense to run the optimization every time we open the projection object. 

It is a bit unsatisfying indeed. I don't know the math going into the map projection well enough to calculate the correct lon_0,lat_0, and azimuth, but may be I could design a optimization process to find them. Either way, we will have to pass a hard coded number in to pyproj.

philippemiron:
Should we also add this projection to `docs/source/reference/projections.rst` ? Ps: thanks for taking the lead on that @MaceKuailv !!

MaceKuailv:
> Should we also add this projection to `docs/source/reference/projections.rst` ? Ps: thanks for taking the lead on that @MaceKuailv !!

Thank you! @philippemiron 

Yeah, we can definitely do that. My only concern is that between we merge this PR and making a public release, users will see this projection in the documentation and no be able to use it without cloning the repo. I could open another PR just to add this to the documentation. Let me know what the preferred work flow is.

greglucas:
> Yeah, we can definitely do that. My only concern is that between we merge this PR and making a public release, users will see this projection in the documentation and no be able to use it without cloning the repo. I could open another PR just to add this to the documentation. Let me know what the preferred work flow is.

The documentation is only released when we make a code release, so no problem there. You can run that `make_projections` script here or in a follow-up PR. But, I agree it would be good to run that before the next release. 👍

lib/cartopy/crs.py
greglucas: What happens if you do the `area_of_use` approach above?
`CRS.from_proj4("+proj=...").area_of_use.bounds`

I agree we don't want to run this on every initialization and hardcoding some values would be useful. My overall comment is that you've chosen some values that are _close_ to being a square, so do we want to just make those a square arbitrarily since you are somewhat choosing arbitrary initial points anyways.

I wonder if the values being different from epsg.io are due to what you mention with the lon_0, lat_0 not being 0 as is typical in many other projections.

lib/cartopy/tests/crs/test_spilhaus.py
```diff
@@ -0,0 +1,40 @@
...
+    reason="Requires PROJ >= 9.6.0"
+)
+def test_defaults():
+    crs = ccrs.Spilhaus()
+    expected = {'rot=45','x_0=0.0','y_0=0.0'} | common_arg
+    check_proj_params('spilhaus', crs, expected)
+
+@pytest.mark.skipif(
+    (proj_version < parse_version("9.6.0")),
+    reason="Requires pyproj > 3.7.0"
```
greglucas, line 34:
```suggestion
    reason="Requires PROJ >= 9.6.0"
```

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -173,6 +177,10 @@ def test_simple_global():
                  id='InterruptedGoodeHomolosine'),
     ccrs.LambertCylindrical,
     ccrs.LambertZoneII,
+    pytest.param(ccrs.Spilhaus,marks=pytest.mark.skipif(
+            (proj_version < parse_version("9.6.0")),
+            reason="Spilhaus not supported in pyproj ≤ 3.7.0"
```
greglucas, line 182:
```suggestion
            reason="Requires PROJ >= 9.6.0"
```

greglucas (commented):
Overall looks good :+1: a few more doc updates.

lib/cartopy/crs.py
MaceKuailv: I just get a None object if I pass in the proj4 string. We can make it a perfect square, but that either leave some empty space or make the projection of some points outside the bounds. If we were to make it a square, I suggest we use the largest absolute value (11802684.083372328). However, I still much prefer that we just leave it slightly imperfect. 

It is unclear what the bounds in epsg.io comes from. In any case, the correct transformed value should look something like a square. This is what it looks like when I apply the bounds on epsg.io: 
image

This is what it looks like when I set lon_0, and lat_0 to zero under the same bounds
image

docs/make_projection.py
```diff
@@ -95,7 +95,8 @@ def utm_plot():
                   'OSGB': 4, 'LambertZoneII': 4.1, 'EuroPP': 5,
                   'Geostationary': 6, 'NearsidePerspective': 7,
                   'EckertI': 8.1, 'EckertII': 8.2, 'EckertIII': 8.3,
-                  'EckertIV': 8.4, 'EckertV': 8.5, 'EckertVI': 8.6}
+                  'EckertIV': 8.4, 'EckertV': 8.5, 'EckertVI': 8.6,
+                  'Spilhaus':9}
```
greglucas, line 99:
```suggestion
                  'Spilhaus': 9}
```

lib/cartopy/crs.py
```diff
@@ -3219,6 +3219,40 @@ def x_limits(self):
     def y_limits(self):
         return self._y_limits
 
+class Spilhaus(Projection):
+    """
+    Spilhaus World Ocean Map in a Square
```
greglucas, line 3224: I think we should add a link to the story about this projection which explains some of the ideas here too and is a nice historical reference.
https://storymaps.arcgis.com/stories/756bcae18d304a1eac140f19f4d5cb3d

In particular it explains why those two points you chose below are special. Maybe even put those antipodal points int he docstring here too?

greglucas (approved):
👍 Looks good to me. One more minor formatting change and a request to add a link to the ArcGIS story or something similar to explain the projection a little more.

greglucas (approved):
LGTM, thanks for incorporating this!

philippemiron:
@greglucas do you know when/how new version of cartopy are released? Cheers and have a good weekend!

greglucas:
> @greglucas do you know when/how new version of cartopy are released?

We don't have any set schedule, but it has been every roughly 6 months - 1 year. It looks like we are about due for another release and have quite a few things ready to go. I opened this issue https://github.com/SciTools/cartopy/issues/2532 to track any features folks want to see get in. Feel free to leave a comment there if you have anything else you'd like to see in!

rcomer:
@guidocioni I suggest posting your example as a new issue.

guidocioni:
> @guidocioni I suggest posting your example as a new issue.

I wanted to but does it make sense as this has not yet been published into a release? Otherwise I'm happy to do so :)

MaceKuailv:
@guidocioni I ran into this issue as well. It is essential the same issue when you try to plot across datelines. Since the border here is quite a bit harder to define than “180W”, this makes the plotting tricky.

I have a solution that involves figuring out the boundaries and masking the boundary grid points with NaN. It is a pretty ad hoc solution, but it works OK for high resolution simulations. I can share an example in your new issue.

Issue #2532, opened 2025-05-22
Prepare for v0.25 release


greglucas:
It has been a little over 6 months since our last release and we have had quite a few nice improvements in that time. We should look at releasing those features soon. Let us know if there are any issues/PRs that people are hoping to get in before the next release and we can try to prioritize looking at those.

ybmy001:
When will the version 0.25 of cartopy be released?

greglucas:
I will plan on tagging a new release Friday 8/1/2025. If anyone has any issues with that date, feel free to comment now.

greglucas:
Release was published today!

PR #2539, opened 2025-05-25
DOC: Add v0.25 whats new entries

greglucas:
This is in preparation for the v0.25 release. I went through all PRs since the previous release and added my quick description to the notes with authors. I added a new section "Maintenance" to try and add some credit to the helpful work many people did during this release that aren't really features, but they keep the project moving forward. Let me know if there is a better place for that, or if people think that is too much to add here. It looks like previously we were putting these into the Features heading, but only some of them.

Preparing for https://github.com/SciTools/cartopy/issues/2532

### DIFF
--- a/docs/source/whatsnew/v0.25.rst
+++ b/docs/source/whatsnew/v0.25.rst
@@ -1,21 +1,91 @@
 Version 0.25 (Date TBD)
 =======================
 
+This version contains a lot of maintenance work and updating internal
+portions of the code to use more modern Python features and methods.
+There was a lot of work done in path transformations to provide more
+robust Shapely to Matplotlib Path and vice versa transformations. Additionally,
+there are several new projections that were added in this release.
+
 The new minimum supported versions of dependencies that have been updated are:
 
 * Shapely 2.0
 
 
-Features
---------
+🎉 Features
+-----------
 
 * Ruth Comer introduced `~cartopy.mpl.path.shapely_to_path` and
   `~cartopy.mpl.path.path_to_shapely` which map a single Shapely geometry or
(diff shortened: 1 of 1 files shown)


docs/source/whatsnew/v0.25.rst
```diff
@@ -13,6 +19,62 @@ Features
...
+  projected incorrectly. (:pull:`2470`)
+
+* Greg Lucas replaced the deprecated ``utcnow()`` calls.
+
+* Warrick Ball updated the BibTex citation. (:pull:`2485`)
+
+* Greg Lucas fixed gridline handling of offset central longitudes.
+  The gridlines now extend to the edges of the maps. (:pull:`2489`)
+
+* Edwin added an example demonstrating how to reproject a raster image.
```
rcomer, line 52: Maybe link out to the example here (and below)?

rcomer (commented):
Wow there is a lot more here than I realised!

Iris tries to include all PRs in the whatsnew and [has 8 different categories](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.12.html).  I'm not suggesting should follow Iris exactly but maybe there are enough documentation changes and bug fixes to give those their own categories?

greglucas:
> Iris tries to include all PRs in the whatsnew and [has 8 different categories](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.12.html). I'm not suggesting should follow Iris exactly but maybe there are enough documentation changes and bug fixes to give those their own categories?

I like this idea and the extra formatting associated with it. I'll try to reorganize them into a few more categories then.

The other question I've had for release notes formatting: Currently I look up people's names in the GitHub profiles and include those names in the notes. Would we prefer the Github handles directly instead?

PR #2540, opened 2025-05-26
MNT: Apply ruff formatter

greglucas:
Currently we use the ruff linter for style checks, but we don't have any explicit formatter. I'm proposing here that we also use the ruff formatter (drop-in replacement for black). I personally have learned to appreciate the opinionated auto-formatters because then as a reviewer I don't have to comment on minor formatting issues and as a coder I can just let the auto-formatter do its thing.

Most of the changes are single-quote to double-quote and adding a space after import lines / docstring lines. In general, I would say that most of the updates are improvements, there are a few that I maybe wouldn't have done but in my opinion that is OK and we can just let the formatter do its thing and handle the wrapping etc for us (they are still easy enough for a human to parse what goes where).


greglucas:
> ruff can be set to single quotes, if that simplifies the diff (and is my personal preference, tbh.)

It does seem like we have more single-quote to begin with so it probably does shrink it a little bit, but not a crazy amount. I'm fine doing it either way and for this project it probably does make more sense because there are JSON strings in geometry files that use double-quotes where it is nicer to not escape those.

One interesting thing is that the sort __all__ pre-commit hook didn't like this change because it must auto-format its sort with only double quotes by default and then the ruff formatter came back and tried to change them to single-quotes again. I've added that RUF022 rule instead for consistency.

PR #2545, opened 2025-06-15
Create django.yml

Addmingbig2:


QuLogic:
Django is not used here; I don't know what the point of this PR is.

PR #2556, opened 2025-08-01
WHL: enable cp314 wheels

neutrinoceros:
## Rationale
I missed the opportunity to do this in time for 0.25.0, but I just tried this on my fork and saw no compilation error, so I'm proposing this change now.
I notice the wheels are not immediately tested, so I'm adding 3.14 to CI too (but I haven't tried that part on my fork yet, so, opening as a draft)

## Implications
Removing `CIBW_BUILD` has the effect of letting `cibuildwheel` take the wheel (pun intended) and select target versions for you, where
- the oldest Python version is either the oldest supported by `cibuildwheel` itself (currently 3.8) or the oldest version compatible with `project.requires-python`, as specified in `pyproject.toml`, whichever is newer.
- the newest Python version selected depends on your pinned version of `cibuildwheel`, and usually include all ABI stable versions, so currently this goes up to 3.14

I'm however also adding a filter to `CIBW_SKIP`, otherwise `cp314t` wheels are *also* included by default, and I don't know what your plans are in this area.


neutrinoceros:
Ah, of course `pyproj` doesn't trivially build from source so it's still a bit early to add 3.14 to CI.

greglucas:
Yeah, I figured that would be the case, so I didn't try too hard on the 3.14 aspect since we are still in rc's on many projects. Lets not wait another year to make a release though (note to self ;) ) so once our dependent projects release their wheels we should try to build and publish ours as well.

neutrinoceros:
I might forget about this PR when the time comes, so please don't hesitate to do what you please with it if it's any help in the future !

neutrinoceros:
Thanks ! I'll try to come back to this in a month or so and see where things are then !

neutrinoceros:
Looks like pyproj 3.7.2 came early with cp314 wheels, so let's try this one more time

neutrinoceros:
the current blocker is now shapely

djhoese:
Dang you're quick. I just released pyproj a couple hours ago.

neutrinoceros:
current blocker: fiona
this one is only an *optional* dependency though, so it should be possible to avoid it.

QuLogic (approved):

.github/workflows/release.yml
```diff
@@ -98,9 +98,8 @@ jobs:
         with:
           package-dir: dist/${{ needs.build_sdist.outputs.SDIST_NAME }}
         env:
-          CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-*"
           # Skip 32 bit builds and musllinux due to lack of numpy wheels
-          CIBW_SKIP: "*-win32 *_i686 *-musllinux*"
+          CIBW_SKIP: "*-win32 *_i686 *-musllinux* cp3*t-"
```
QuLogic, line 102: Hmm, actually, this does nothing, because the freethreaded one is missing the trailing `*`.

neutrinoceros: thank you. Fixed

adamjstewart:
Anything I can do to help get this merged and released?

QuLogic (approved):

PR #2558, opened 2025-08-05
Add readthedocs.yml and remove Circle-CI config files

rcomer:
## Rationale

First attempt at making a readthedocs config file.


QuLogic:
Maybe we can install `certifi` for newer root certificates, if we aren't already.

rcomer:
Well I've no idea how, but it looks like switching to conda fixed it.  It seems faster that Circle-CI too.

greglucas (approved):

greglucas (approved):
I went into CircleCI and clicked the buttons "stop building" and "delete trigger", so hopefully that was enough 🤷

PR #2560, opened 2025-08-07
pre-commit ruff fix

rcomer:
## Rationale

Pre-commit is failing on `main` because ruff says this string is a "legacy alias".  From the [docs](https://docs.python.org/3/library/functions.html#open) the string is valid, albeit it is the default and therefore redundant.

### DIFF
--- a/lib/cartopy/io/shapereader.py
+++ b/lib/cartopy/io/shapereader.py
@@ -141,8 +141,8 @@ class BasicReader:
         # Try to set the CRS from the .prj file if it exists
         self.crs = None
         try:
-            with open(Path(filename).with_suffix('.prj'), 'rt') as fobj:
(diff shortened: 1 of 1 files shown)


rcomer:
The three failing tests were (most recent attempt first)

```
FAILED lib/cartopy/tests/mpl/test_caching.py::test_shapefile_transform_cache - struct.error: unpack requires a buffer of 11088 bytes
FAILED lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[PlateCarree] - struct.error: unpack requires a buffer of 8 bytes
FAILED lib/cartopy/tests/mpl/test_features.py::test_gshhs - http.client.IncompleteRead: IncompleteRead(344446 bytes read, 148813399 more expected)
```

PR #2561, opened 2025-08-07
Backport PR #2558 to v0.25.x: Add readthedocs.yml and remove Circle-CI config files

rcomer:
## Rationale

Backport #2558, adding the readthedocs config, so we can make a stable docs build.


greglucas (approved):

greglucas:
Care to tag it how you think it needs to be @rcomer? I'm not sure what the tag should be. Maybe we move the v0.25.0 tag so that the docs build as v0.25.0? Or add a `.post1` qualifier, but then the docs would be slightly odd to show that I think.

QuLogic:
> Maybe we move the v0.25.0 tag so that the docs build as v0.25.0?

No, no, no, never move a tag. Then it won't correspond with PyPI or any downstream builds that have been made, and won't match on anyone's forks/clones that had the original.

> OK I updated the v0.25.0 tag as I agree that will look better in the switcher. I don't _think_ I broke anything 😅

Now if I try to fetch tags, I get a conflict:
```
$ git fetch upstream --tags
From https://github.com/SciTools/cartopy
 ! [rejected]          v0.25.0    -> v0.25.0  (would clobber existing tag)
```
I may have to force push the original one back, especially since it wasn't used by RTD.

greglucas:
>> Maybe we move the v0.25.0 tag so that the docs build as v0.25.0?

> No, no, no, never move a tag. Then it won't correspond with PyPI or any downstream builds that have been made, and won't match on anyone's forks/clones that had the original.

Doh! Was not thinking through everything while typing quickly there. I guess that is why they say not to trust a tag and rely on the hashes instead. @rcomer, I take full blame for this one. Thank you for going through the effort of doing it in the first place!

When I go look at the tags now it is showing me both tags were 9 hours ago, but your comment was 2 hours ago. Was it not fixed when you pulled, or were you somehow able to update the time of the previous tag? Any suggestions for how to go about this in the future? I still see that the stable documentation gets a `0.1...` version identifier for some reason. https://cartopy.readthedocs.io/stable/reference/index.html

QuLogic:
> When I go look at the tags now it is showing me both tags were 9 hours ago, but your comment was 2 hours ago. Was it not fixed when you pulled, or were you somehow able to update the time of the previous tag?

I already had the original tag from last week, 9 hours ago is the new one. I can't fetch the new one (without explicitly deleting stuff.)

> Any suggestions for how to go about this in the future? I still see that the stable documentation gets a `0.1...` version identifier for some reason. https://cartopy.readthedocs.io/stable/reference/index.html

If I run `git describe`, I get back some random result:
```
$ git show --quiet
commit 2264d277a5b2900b010244d11f80bb523df349dd (HEAD -> main, tag: v0.25.0)
Author: Greg Lucas <greg.m.lucas@gmail.com>
Date:   Fri Aug 1 06:37:05 2025 -0600

    CI: Update release runner version (#2555)

$ git describe
v0.20.1-705-g2264d277
```

From the `git tag --help` docs:
> Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG signature. Whereas a "lightweight" tag is simply a name for an object (usually a commit object).
>
> Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. For this reason, some git commands for naming objects (like git describe) will ignore lightweight tags by default.

So based on that, it seems like all the recent tags are lightweight tags. And this would be a longstanding "bug" with GitHub: https://github.com/orgs/community/discussions/4924 (and from there, see https://github.com/pypa/setuptools-scm/issues/521) So the best way to get proper tags is to not use GitHub for it. I'm not sure though why RTD is not fetching the tags that _are_ annotated; it should have at least been as new as mine.

QuLogic:
> I'm not sure though why RTD is not fetching the tags that _are_ annotated; it should have at least been as new as mine.

Thinking about it a bit more, they probably do a shallow clone, which may not fetch _other_ annotated tags, and since this one is lightweight, it doesn't count either.

QuLogic:
> When I go look at the tags now it is showing me both tags were 9 hours ago, but your comment was 2 hours ago.

I pushed the old tag back, since it wasn't used by RTD anyway. Now it says last week instead of earlier today.

QuLogic:
Also, it does look like @rcomer did use an annotated tag for the post release:
```
$ git show v0.25.0.post1 
tag v0.25.0.post1
Tagger: Ruth Comer <ruth.comer@metoffice.gov.uk>
Date:   Thu Aug 7 17:35:42 2025 +0100

New tag for RTD

commit 1e8acc400c12d198a3a7c23ac1aa7b8429ba58a1 (tag: v0.25.0.post1, upstream/v0.25.x)
Merge: 2264d277 40098e8d
Author: Greg Lucas <greg.m.lucas@gmail.com>
Date:   Thu Aug 7 09:44:55 2025 -0600

    Merge pull request #2561 from rcomer/backport-rtd-config
    
    Backport PR #2558 to v0.25.x: Add readthedocs.yml and remove Circle-CI config files
```
but looking at builds in RTD, it doesn't look like this triggered anything? Maybe it only builds semver tags?

QuLogic:
I found https://github.com/readthedocs/readthedocs.org/issues/10715 and opened #2563

PR #2569, opened 2025-08-13
ENH: support autolim keyword for `add_feature`/`add_geometries`

rcomer:
## Rationale

Inspired by [this StackOverflow question](https://stackoverflow.com/q/79733680/3501128).  The automatic update of the axes extent does not currently work because `Collection.get_datalim` relies on the [get_paths](https://github.com/matplotlib/matplotlib/blob/780e66c7975ccf27abd104f105bf773f1e6cd0d2/lib/matplotlib/collections.py#L272-L275) method.  Since the `FeatureArtist` paths are created on-the-fly at draw, the rest of the time `get_paths` returns an empty list.  This PR modifies `get_paths` to return the same paths that `draw` would use.  It also adds the _autolim_ keyword to `add_feature` and `add_geometries`, defaulting to `False` for backward compatibility and because that is almost certainly what you want if using something like a Natural Earth feature.

We can now do

```python
import cartopy.crs as ccrs
from shapely.geometry import Point

f = plt.figure()
ax = f.add_subplot(projection=ccrs.Mollweide())

# plot a normal artist
ax.plot([-150000,2000,150000, 250000], 
        [-50000, 2000, 50000, -250000], 
        lw=5, c="g")

# add some geometries
for x, y in zip([10, 20, 40], [12, 34, 46]):
    p = Point((x, y)).buffer(3)
    ax.add_geometries(p, fc="r",crs=ccrs.PlateCarree(), autolim=True)

ax.autoscale_view()

plt.show()
```
The call to `autoscale_view` will not be necessary from Matplotlib v3.11 because of https://github.com/matplotlib/matplotlib/pull/29958

Figure_1

### DIFF
--- a/lib/cartopy/tests/mpl/test_feature_artist.py
+++ b/lib/cartopy/tests/mpl/test_feature_artist.py
@@ -114,7 +114,7 @@ def test_feature_artist_draw_styler(feature):
             return {'facecolor':'blue'}
 
     fig, ax = robinson_map()
-    ax.add_feature(feature, facecolor='red', styler=styler)
+    ax.add_feature(feature, facecolor='grey', styler=styler)
 
     return fig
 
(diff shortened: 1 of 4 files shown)


lib/cartopy/tests/mpl/test_feature_artist.py
```diff
@@ -114,7 +114,7 @@ def styler(geom):
             return {'facecolor':'blue'}
 
     fig, ax = robinson_map()
-    ax.add_feature(feature, facecolor='red', styler=styler)
+    ax.add_feature(feature, facecolor='grey', styler=styler)
```
rcomer, line 117: Changing this to 'grey' helps distinguish between the failures 'styler was not applied' and 'styler was applied wrongly'.

PR #2585, opened 2025-10-23
DOC: new example, recreation of The Simpsons Monorail Map

sadielbartholomew:
## Rationale

For fun (and love of said TV Series), I have recreated the Monorail Map from The Simpsons using Cartopy - the challenge was to get as close to likeness as I could using the standard Python visualisation ecosystem - matplotlib etc, in a small-ish script, and Cartopy made this simple. 

Whist I did this mostly with the motivation of making high-res print outs for myself, I think it would make a good example for your examples gallery. As I summarise in the top-level docstring:

> This example demonstrates how to create a minimal outline map of a defined area of land such as a continent, with optional labels at specified locations within the region,

Moreover, other vis libraries with examples galleries have humorous examples - notably the XKCD examples - so I think i is nice to add (another) one to the Cartopy examples gallery.

From other PRs adding new examples I have found from a quick search, it seems that I just need to add the new script in the appropriate place under 'cartopy/examples/` and the relevant subdirectory to categorise - I think 'miscellanea' is most suitable, but happy to move it if you wish.

Please let me know if you would like me to adjust it in any way if you would accept it for the gallery - either codewise or output-wise.

## Implications

A new example recipe will be added to the docs and maybe somebody will find it useful and/or amusing. 🙂 

## Notes

The output (with Python 3.12) is:

Original from when PR opened:

504936792-10f33258-e212-484b-9eb1-57b190c940e0

Updated output after feedback response and tweaks, at a8baf8af1b21190b7015c462ea386f1e0ac0331c:

new_simpsons_map

For reference, the original [looks like this](https://simpsons.fandom.com/wiki/Brockway?file=S04E12_Marge_vs._the_Monorail_-_Maps.jpg).

### DIFF
--- a/examples/miscellanea/simpsons-monorail-map.py
+++ b/examples/miscellanea/simpsons-monorail-map.py
@@ -112,7 +112,7 @@ def main():
     # Add the bottom left 'compass' legend in spirit of the original map.
     map_ax.text(
         0.14,
-        0.0,
+        0.10,
         leg_text,
         transform=map_ax.transAxes,
         fontsize=11,
--- a/examples/miscellanea/simpsons-monorail-map.py
+++ b/examples/miscellanea/simpsons-monorail-map.py
@@ -25,9 +25,11 @@ import cartopy.io.shapereader as shpreader
 
 # Define choices for projection and locations to plot
 GEOM_PROJ = ccrs.PlateCarree()
-# Not real places, so locations pulled from location on map in still image
-# First value 2-tuple is the dot placemap location, second is where to write
-# the text label relative to that dot, to best match the map from the show.
+# Not real places, so locations pulled from location on map in still image.
+# The items in order correspond to:
+# * an x-y 2-tuple for where to plot the location dot on the map
+# * an x-y 2-tuple for where to write the text label relative to the dot above
+# * an integer for how much to rotate the text to best match the map text.
 LOCATIONS_TO_PLOT = {
     "Ogdenville": [(-111.8, 35.5), (1.5, -2.2), -6],
     "North\nHaverbrook": [(-99.0, 43.5), (2.8, -0.5), -1],
--- a/examples/miscellanea/simpsons-monorail-map.py
+++ b/examples/miscellanea/simpsons-monorail-map.py
@@ -18,11 +18,9 @@ https://simpsons.fandom.com/wiki/Brockway.
 """
 
 import matplotlib.pyplot as plt
-from matplotlib.transforms import offset_copy
 
 import cartopy.crs as ccrs
 import cartopy.io.shapereader as shpreader
-from matplotlib.patches import Rectangle
 
 
 # Define choices for projection and locations to plot
--- a/examples/miscellanea/simpsons-monorail-map.py
+++ b/examples/miscellanea/simpsons-monorail-map.py
@@ -22,6 +22,7 @@ from matplotlib.transforms import offset_copy
 
 import cartopy.crs as ccrs
 import cartopy.io.shapereader as shpreader
+from matplotlib.patches import Rectangle
 
 
 # Define choices for projection and locations to plot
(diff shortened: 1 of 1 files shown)


examples/miscellanea/simpsons-monorail-map.py
```diff
@@ -0,0 +1,131 @@
...
+        fontsize=11,
+        horizontalalignment="center",
+        verticalalignment="center",
+        style="italic",
+        bbox=dict(facecolor="#A5B5CE"),
+    )
+
+    # Make border symmetrical since default 'rc' file has asymmetric side pad
+    fig.tight_layout()
+    fig.subplots_adjust(left=0.035, bottom=0.035, right=0.965, top=0.965)
```
greglucas: Do you want to call `add_axes()` above to make an explicit placement in figure space?

examples/miscellanea/simpsons-monorail-map.py
```diff
@@ -0,0 +1,131 @@
...
+        transform=ax.transAxes,
+        fontsize=11,
+        horizontalalignment="center",
+        verticalalignment="center",
+        style="italic",
+        bbox=dict(facecolor="#A5B5CE"),
+    )
+
+    # Make border symmetrical since default 'rc' file has asymmetric side pad
+    fig.tight_layout()
```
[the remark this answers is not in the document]
sadielbartholomew: Fair enough, thanks for pointing this out. I will remove it in a feedback-response commit which I will push shortly.

examples/miscellanea/simpsons-monorail-map.py
sadielbartholomew: I will try playing around with use of `add_axes`, though from your previous comment it seems we don't want multiple axes so I am a bit confused as to how that fits with said plan - unless I use that to create the one and only axis. Would be useful if you can clarify?

greglucas: Sorry for the confusion, I was making various comments depending on which direction you were going to go. I agree with only one `axes`, how you put it in depends on what you want to do with it. If you want to explicitly place it or for it to take up the entire figure then I would manually place it with `ax = fig.add_axes([0, 0, 1, 1], projection=GEOM_PROJ)` (or subtract the margins you have above as necessary)

examples/miscellanea/simpsons-monorail-map.py
```diff
@@ -0,0 +1,136 @@
...
+    ]
+    map_ax.add_geometries(
+        usa_border,
+        GEOM_PROJ,
+        facecolor="#C39B6A",
+        edgecolor="black",
+    )
+
+    # Now add the location labels. The general approach is that covered in
+    # the 'Map tile acquisition' Catopy example, including the comment below.
```
rcomer: I think [annotate](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.annotate.html#matplotlib.axes.Axes.annotate) is the right tool for this job these days, so I have opened #2596 to update the existing example.

Note, you could use 'offset points' or 'offset fontsize' if you prefer one of those to 'offset pixels'.

sadielbartholomew: OK, so what should I do for my PR code, if anything?

rcomer: I recommend using `annotate`, as it should simplify the code (and not make use of the private `_as_mpl_transform` method).

sadielbartholomew:
Have addressed new feedback from @rcomer (awaiting one clarification RE https://github.com/SciTools/cartopy/pull/2585#discussion_r2489279148) and force pushed after rebasing onto main.

examples/miscellanea/simpsons-monorail-map.py
```diff
@@ -0,0 +1,129 @@
...
+import cartopy.crs as ccrs
+import cartopy.io.shapereader as shpreader
+from matplotlib.patches import Rectangle
+
+
+# Define choices for projection and locations to plot
+GEOM_PROJ = ccrs.PlateCarree()
+# Not real places, so locations pulled from location on map in still image
+# First value 2-tuple is the dot placemap location, second is where to write
+# the text label relative to that dot, to best match the map from the show.
```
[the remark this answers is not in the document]
sadielbartholomew: Done in [6dac226](https://github.com/SciTools/cartopy/pull/2585/commits/6dac22686d1ecaee830599e75d9a88af02d1da4c) (and updated in [51631ed](https://github.com/SciTools/cartopy/pull/2585/commits/51631edc3ad19b386db2a27bb5bc57e3044ffd6d) after the effective unit change from further feedback). Thanks.

sadielbartholomew:
Sorry for the delay @rcomer and team - hard to find the time at the moment - but I believe I have addressed your very helpful next round of feedback, so again should be ready for re-review or merge. Thanks for your patience.

rcomer (approved):
Thanks for your work on this @sadielbartholomew - looks great!

PR #2602, opened 2025-11-15
FIX: Image tiles crossing the dateline need special handling

### LINKED ISSUE #2579: Coastlines shifted when using add_image with Orthographic projection centered on dateline
When using `ax.add_image(cimgt.GoogleTiles(), …)` in an Orthographic projection with central_longitude=180, the coastlines from the image tiles are noticeably shifted relative to the actual coastlines.

Environment: python=3.11.9, matplotlib=3.10.3, cartopy=0.25.0

```python
from matplotlib import pyplot as plt
import cartopy.crs as ccrs
import cartopy.io.img_tiles as cimgt

ax = plt.axes(projection=ccrs.Orthographic(central_longitude=180))
ax.add_image(cimgt.GoogleTiles(), 2)
ax.coastlines()
plt.show()
```

Image

When using lower zoom levels (e.g., `ax.add_image(cimgt.GoogleTiles(), 1)`) the coastlines seem to align, as expected.

```python
ax = plt.axes(projection=ccrs.Orthographic(central_longitude=180))
ax.add_image(cimgt.GoogleTiles(), 1)
ax.coastlines()
plt.show()
```

Image

Is this expected behaviour for Web Mercator image tiles? Is there a fix for this to prevent images tiles with zoom levels higher than 2 from becoming misaligned?

greglucas:
When image tile requests crossing a dateline happen, there is a multipolygon extent that gets passed to the find images function. We don't necessarily want just the tiles that are within those geometries, but also the ones that could be between them.

closes #2579

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -125,6 +125,13 @@ class GoogleWTS(metaclass=ABCMeta):
         # Recursively drill down to the images at the target zoom.
         x0, x1, y0, y1 = self._tileextent(start_tile)
         domain = sgeom.box(x0, y0, x1, y1)
+
+        if hasattr(target_domain, 'geoms') and len(target_domain.geoms) > 1:
+            # For MultiPolygon, create a unified coverage area that includes
+            # the space between disconnected parts. This ensures tiles spanning
+            # across dateline discontinuities are included.
+            target_domain = sgeom.box(*target_domain.bounds)
+
         if domain.intersects(target_domain):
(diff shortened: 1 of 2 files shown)


greglucas:
Thinking about this a bit more, this might expand the domain a little bit too much depending on the bounds of the polygons and turn two halves of an oval shape into a rectangle unnecessarily. I'll try and look into whether buffering the multipolygon is more appropriate here because I am now wondering if this has to do with the inexact boundary of Orthographic (or any curved boundaries).

Issue #2595, opened 2025-11-03
Some tile classes clash local cached tile file names


Guymer:
Currently, the `GoogleWTS` class does not use `style` when determining the local file path in the cache (e.g., `GoogleTiles` has `style="street"` by default but accepts "street", "satellite", "terrain" and "only_streets"). Therefore, if a tile already exists for the same `(x, y, z)` which the user requests then it will be blindly used even if the cached style is not what is currently being asked for. The local path of cached tiles should be changed to include all variables which are used to create the URL.

PR #2603, opened 2025-11-15
Bugfix/avoid cache clashes

Guymer:
## Rationale

This Pull Request solves #2595.

## Implications

This Pull Request:

* Added `resolution`, `layer` and `style` optional keyword arguments, and corresponding attributes, to `GoogleWTS` (this means that when other classes set the `resolution`, `layer` and `style` attributes then they are merely populating a pre-existing attribute rather than creating a new one).
* Moved some docstrings from the top of the class definition to be in the `__init__` method instead (which is now consistent across all the classes).
* Added a docstring to `GoogleWTS`, including the new `resolution`, `layer` and `style` optional keyword arguments.
* The `_cache_dir()` method now uses the (always present) `resolution`, `layer`, `style` attributes if they are set (this is the actual change which fixes the issue - the rest is tidying and/or consistency and/or neatness).
* Renamed "map_id" to "style" in `MapboxTiles` and `MapboxStyleTiles` (this was a positional argument, not a keyword argument, so this should not break any users' scripts).
* Renamed "layer_id" to "layer" in `LINZMapsTiles` (this was also a positional argument, not a keyword argument, so this should not break any users' scripts).

Whilst `layer` and `style` now do the same thing functionally, I decided to *not* rename all usage of "layer" to "style" as that might upset people and/or break scripts.

Here is an example using the OSM tile provider (showing no change):

norway-aug-2024

Here is an arbitrary example using a style of "outdoors" and a resolution of "@2x" using the Thunderforest tile provider:

norway-aug-2024_outdoors_@2x

Here is a list of the folders created during my own testing in my local cache:

```
redacted@redacted ~ % find ~/.local/share/cartopy_cache -type d | sort
/Users/redacted/.local/share/cartopy_cache
/Users/redacted/.local/share/cartopy_cache/OSM
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/atlas
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/atlas/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/cycle
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/cycle/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/landscape
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/landscape/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/mobile-atlas
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/mobile-atlas/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/neighbourhood
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/neighbourhood/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/outdoors
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/outdoors/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/pioneer
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/pioneer/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/spinal-map
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/spinal-map/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/transport
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/transport-dark
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/transport-dark/@2x
/Users/redacted/.local/share/cartopy_cache/ThunderforestTiles/transport/@2x
```

### DIFF
--- a/lib/cartopy/io/img_tiles.py
+++ b/lib/cartopy/io/img_tiles.py
@@ -186,7 +186,6 @@ class GoogleWTS(metaclass=ABCMeta):
 
     def tile_bbox(self, x, y, z, y0_at_north_pole=True):
         """The ``(x0, x1), (y0, y1)`` bounding box for the given tile position.
-        tile position.
 
         Parameters
         ----------


lib/cartopy/io/img_tiles.py
```diff
@@ -862,7 +851,7 @@ def _image_url(self, tile):
 
 class LINZMapsTiles(GoogleWTS):
 
-    def __init__(self, apikey, layer_id, api_version="v4",
+    def __init__(self, apikey, layer, api_version="v4",
```
greglucas: This one is called "layer_id" in their documentation. I'm not sure how I feel about changing this.
https://www.linz.govt.nz/guidance/data-service/linz-data-service-guide/changesets/changeset-url-syntax

What if you kept it all the same here and in the super call you call it with `layer=layer_id`

greglucas (commented):
Overall looks like a good contribution to me. Some minor notes on trying to maintain backwards compatibility if we can.

If we are passing everything up to the `super()` class, then I think you can also remove a lot of the `self.layer = layer` and let super take care of that for you.

lib/cartopy/io/img_tiles.py
```diff
@@ -32,29 +32,51 @@
...
 
 class GoogleWTS(metaclass=ABCMeta):
-    """
-    Implement web tile retrieval using the Google WTS coordinate system.
+    _MAX_THREADS = 24
 
-    A "tile" in this class refers to the coordinates (x, y, z).
+    def __init__(self, desired_tile_form='RGB',
+                 user_agent=f'CartoPy/{cartopy.__version__}',
+                 resolution="", layer=None, style=None, cache=False):
```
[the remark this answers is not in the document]
Guymer: An excellent point: I was assuming that users were only passing them as keyword arguments. I have moved the new keyword arguments to the end of the definition to maintain backwards compatibility as requested - I will push that commit shortly.

lib/cartopy/io/img_tiles.py
```diff
@@ -477,51 +495,44 @@ def _image_url(self, tile):
...
-    even if the resolution or style is not what was asked for. You may need to
-    manage multiple cache paths to work around this issue.
-    """
-
     def __init__(self,
-                    apikey,
-                    style="landscape",
-                    resolution="",
-                    cache=False):
-        super().__init__(cache=cache, desired_tile_form="RGBA")
```
Guymer: I checked: Thunderforest PNGs are only 3-channels.

lib/cartopy/io/img_tiles.py
```diff
@@ -532,72 +535,64 @@ def _image_url(self, tile):
...
 class MapboxTiles(GoogleWTS):
-    """
-    Implement web tile retrieval from Mapbox.
-
-    For terms of service, see https://www.mapbox.com/tos/.
-
-    """
-
-    def __init__(self, access_token, map_id, cache=False):
+    def __init__(self, access_token, style, cache=False):
```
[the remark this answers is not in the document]
Guymer: I have reverted the variable rename, left `map_id` in-place, and then just done `style=map_id` in the `.super()` call.

lib/cartopy/io/img_tiles.py
```diff
@@ -32,29 +32,55 @@
...
+    _MAX_THREADS = 24
 
-    A "tile" in this class refers to the coordinates (x, y, z).
+    def __init__(self,
+                 desired_tile_form='RGB',
+                 user_agent=f'CartoPy/{cartopy.__version__}',
+                 cache=False,
+                 resolution="",
+                 layer=None,
+                 style=None):
```
greglucas: In the cache directory implementation you use the ordering (layer, style, resolution), for consistency we should keep things here the same as well then. We might as well put these after a keyword only flag too.
```suggestion
                 *,
                 layer=None,
                 style=None
                 resolution="",):
```

lib/cartopy/io/img_tiles.py
```diff
@@ -32,29 +32,55 @@
...
+            in which case a change of ``user_agent`` may fix the issue.
+        resolution : str, optional
+            Some providers (like Stadia and Thunderforest) offer tiles at
+            different resolutions.
+        layer : str, optional
+            Some providers (like LINZ and Ordnance Survey) offer tiles of
+            different layers.
+        style : str, optional
+            Some providers (like Google, Mapbox, Stadia, Stamen and
+            Thunderforest) offer tiles in different styles.
```
greglucas: Move these after the `cache` keyword so that they follow the same ordering as the argument order call.

lib/cartopy/io/img_tiles.py
```diff
@@ -602,20 +619,26 @@ def __init__(self, access_token, username, map_id,
...
+            can create a local cache of previously fetched tiles. The default
+            path is ``cartopy.config["cache_dir"]``. If it is set to ``True``,
+            the default path is used. If it is set to a custom path, then this
+            path is used instead of the default one. If it is set to ``False``,
+            the tiles are downloaded each time.
         """
+
         self.access_token = access_token
         self.username = username
-        self.map_id = map_id
```
greglucas: This is still breaking because someone could have been setting `my_mapbox_provider.map_id = my_new_style`.

```python
@property
def map_id(self):
    return self.style

@property.setter
def map_id(self, val):
    self.style = val
```

lib/cartopy/io/img_tiles.py
```diff
@@ -881,16 +918,24 @@ def __init__(self, apikey, layer_id, api_version="v4",
...
+            path is ``cartopy.config["cache_dir"]``. If it is set to ``True``,
+            the default path is used. If it is set to a custom path, then this
+            path is used instead of the default one. If it is set to ``False``,
+            the tiles are downloaded each time.
         """
-        super().__init__(desired_tile_form=desired_tile_form, cache=cache)
+        super().__init__(layer=layer_id,
+                         desired_tile_form=desired_tile_form, cache=cache)
         self.apikey = apikey
-        self.layer_id = layer_id
```
greglucas: Same comment about backwards compatibility and mapping a `layer_id` property to `layer`.

lib/cartopy/io/img_tiles.py
Guymer: I have made this change and will push the commit shortly.

lib/cartopy/io/img_tiles.py
```diff
@@ -145,27 +185,24 @@ def subtiles(self, x_y_z):
     _subtiles = subtiles
 
     def tile_bbox(self, x, y, z, y0_at_north_pole=True):
-        """
-        Return the ``(x0, x1), (y0, y1)`` bounding box for the given x, y, z
+        """The ``(x0, x1), (y0, y1)`` bounding box for the given tile position.
         tile position.
```
greglucas:
```suggestion
```

greglucas (approved):
Looks good to me and thanks for correcting the typos on my suggestions, I hope you were able to learn a bit along the way because it looks like you figured it out nicely :) 

Thanks @Guymer!

PR #2619, opened 2025-12-18
return valid (empty) path if path contains no polygons

raphaelquast:
## Rationale

While playing around in [EOmaps](https://github.com/raphaelquast/EOmaps) with overlapping clipped axes that share limits, I noticed that in case one of the clip-paths is moved completely outside the field of view, the `_ensure_path_closed` method tries to initialize a path without vertices which raises the following error:
```
ValueError: 'vertices' must be 2D with shape (N, 2), but your input has shape (0,)
```

The fix ensures that a valid (empty) path is returned in case the path contains no polygons.

### DIFF
--- a/lib/cartopy/mpl/path.py
+++ b/lib/cartopy/mpl/path.py
@@ -43,6 +43,9 @@ def _ensure_path_closed(path):
     finally:
         path.should_simplify = should_simplify
 
+    if len(polygons) == 0:
+        return Path(np.empty((0,2)))
+
     codes, vertices = [], []
     for poly in polygons:
         vertices.extend([poly[0], *poly])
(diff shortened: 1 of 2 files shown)


raphaelquast:
Well, that depends on how "complete" the test should be... 

Would something like this be sufficient?
(I'd add it to `lib/cartopy/tests/mpl/test_path.py`)

```python
from cartopy.mpl.path import _ensure_path_closed
import numpy as np
import pytest

no_polygon_path = Path([[0,0], [1,1]], codes=[Path.MOVETO, Path.LINETO])
empty_path = Path(np.empty((0, 2)))

class Test_ensure_path_closed:
    @pytest.mark.parametrize('path', [no_polygon_path, empty_path])
    def test_non_polygon_path_closing(self, path):

        closed_path = _ensure_path_closed(path)
        assert isinstance(closed_path, Path)
        assert closed_path.vertices.size == 0
```

greglucas:
Sure, that seems fine to me. Indicating it is a smoke test that we don't error like you found before.

PR #2626, opened 2026-01-21
ci: Fix install of extra packages

QuLogic:
## Rationale

Doing this a second time causes the main package to be re-installed in a non-editable way. This causes coverage to bake in the install path instead of being relative to the package.

Also, add a comment why Cython is manually installed when it's already a requirement for our build.

## Implications

Coverage results are more consistent.


trexfeathers:
> Also, when did the CLA change? I'm sure I signed a v5 already...

https://github.com/SciTools/.github/issues/198#issuecomment-3328823625

Issue #2286, opened 2023-11-14
Expose PROJ parameter +gamma for Oblique Mercator


Jeitan:
### Description

I was/am super happy to see the Oblique Mercator projection in Cartopy as of 0.22, because I have had need of it for a while!  So I guess now the reward is I have an additional **feature request**.

I've actually wanted to have the ability to use it with **a still-rotated resulting projection** (such that it does not force north to be up).  Not being very familiar with PROJ, it took me longer than I care to admit to figure out that all I really needed to do was make a coordinate system with "+gamma" set to 0 (or, it turns out after more testing, 180).

I've written my own sub-class of Projection that looks just like the Oblique Mercator class in crs.py, except am able to pass in gamma (RotatedObliqueMercator), and it seems to work very nicely in my (albeit *very* limited) test cases.  However, it would be much easier to just have gamma exposed as an input parameter in the built-in class.  It could be defaulted to match alpha to maintain any backward compatibility (which is what I think PROJ does if gamma is not given, if I'm reading the documentation correctly).

My only hesitation is potentially anything weird coming from it pulling its limits from Mercator ... but so far that hasn't been an issue (again, limited test cases).

It does seem like this is something that at least a few people are interested in.
I've made an answer to this (admittedly old) question on GIS stack exchange ... not directly relevant to UTM but accomplishes the desired task:
https://gis.stackexchange.com/questions/429513/how-to-produce-a-cartopy-map-in-utm-with-north-east-up

It would also address this question on StackOverflow (the current answer does work but it involves figuring out the location of a new pole which is annoying)
https://stackoverflow.com/questions/71027066/with-cartopy-can-a-local-map-be-rotated-so-that-north-points-in-an-arbitrary-di

It seems like the ramifications would be pretty small, but key for people like me who want that rotation.

greglucas:
Seems reasonable to expose that keyword argument if you'd find it helpful.
https://proj.org/en/9.3/operations/projections/omerc.html

@trexfeathers added this in https://github.com/SciTools/cartopy/pull/2096 so they might have thoughts/opinions on it as well.

trexfeathers:
>@trexfeathers added this in https://github.com/SciTools/cartopy/pull/2096 so they might have thoughts/opinions on it as well.

[...]

PR #2627, opened 2026-02-03
Expose +gamma for Oblique Mercator

epn09:
## Rationale

* I think it was agreed in #2286 that it is fine to expose the gamma parameter of Oblique Mercator projection. However, I have not seen it implemented.

## Implications

* It will be possible to make a rotated map like the example below.

```python
import cartopy.crs as ccrs

japan_extent = [128, 146, 24, 46]
fig = plt.figure(figsize=(8, 8), layout="constrained")

ax1 = fig.add_subplot(
    1,
    2,
    1,
    projection=ccrs.ObliqueMercator(
        central_latitude=35, central_longitude=137, azimuth=40
    ),
)
ax1.coastlines()
ax1.gridlines(draw_labels=True)
ax1.set_extent(japan_extent, crs=ccrs.PlateCarree())

ax2 = fig.add_subplot(
    1,
    2,
    2,
    projection=ccrs.ObliqueMercator(
        central_latitude=35, central_longitude=137, azimuth=40, gamma=0
    ),
)
ax2.coastlines()
ax2.gridlines(draw_labels=True)
ax2.set_extent(japan_extent, crs=ccrs.PlateCarree())

plt.show()
```

image


lib/cartopy/crs.py
```diff
@@ -3178,6 +3178,9 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
         azimuth: optional
             Azimuth of centerline clockwise from north at the center point of
             the centre line. Defaults to 0.
+        gamma: optional
+            Azimuth of centerline clockwise from north of the rectified bearing
+            of centre line. If omitted, `alpha` determines the values of `gamma`.
```
QuLogic: It seems like maybe you copied this description from `proj`. There is no `alpha` here.

QuLogic (commented):
This would needs tests as well.

lib/cartopy/crs.py
epn09: @QuLogic Thank you for pointing out. `alpha` of `proj` is named `azimuth` here. So, I have updated the docstring.

epn09:
@QuLogic I have also added a test.

lib/cartopy/crs.py
```diff
@@ -3184,6 +3184,9 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
         azimuth: optional
             Azimuth of centerline clockwise from north at the center point of
             the centre line. Defaults to 0.
+        gamma: optional
+            Azimuth of centerline clockwise from north of the rectified bearing
+            of centre line. If omitted, `azimuth` determines the values of `gamma`.
```
QuLogic, line 3189: These should be in double backticks for code style, as there is nothing to cross-reference.

epn09: @QuLogic This has been fixed. Thank you.

QuLogic (approved):

QuLogic:
Thanks @epn09! Congratulations on your first PR to Cartopy :tada: We hope to hear from you again.

PR #2640, opened 2026-03-15
FIX: Handle empty points in geometry projection

### LINKED ISSUE #2635: Error drawing 0-area path/contour
In MetPy I have a test failing with Cartopy 0.25 (worked in 0.24), that does some plotting using `contour()`. The error is:

```pytb
src/metpy/plots/declarative.py:1098: in _build
    self.handle = self.parent.ax.contourf(x_like, y_like, imdata, self.contours, **kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/mpl/geoaxes.py:291: in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/mpl/geoaxes.py:335: in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/mpl/geoaxes.py:1644: in contourf
    datalim = result.get_datalim(self.transData)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/matplotlib/collections.py:273: in get_datalim
    paths = [transform.transform_path_non_affine(p) for p in paths]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/matplotlib/transforms.py:2427: in transform_path_non_affine
    return self._a.transform_path_non_affine(path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/mpl/geoaxes.py:174: in transform_path_non_affine
    transformed_geom = self.target_projection.project_geometry(
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/crs.py:833: in project_geometry
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/cartopy/crs.py:836: in _project_point
    return sgeom.Point(*self.transform_point(point.x, point.y, src_crs))
                                             ^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/shapely/geometry/point.py:91: in x
    return float(shapely.get_x(self))
                 ^^^^^^^^^^^^^^^^^^^
../../miniconda3/envs/py314/lib/python3.14/site-packages/shapely/decorators.py:88: in wrapped
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

point = <POINT EMPTY>, kwargs = {}

    @multithreading_enabled
    def get_x(point, **kwargs):
        """Return the x-coordinate of a point.
    
        Parameters
        ----------
        point : Geometry or array_like
            Non-point geometries will result in NaN being returned.
        **kwargs
            See :ref:`NumPy ufunc docs <ufuncs.kwargs>` for other keyword arguments.
    
        See Also
        --------
        get_y, get_z, get_m
    
        Examples
        --------
        >>> import shapely
        >>> from shapely import MultiPoint, Point
        >>> shapely.get_x(Point(1, 2))
        1.0
        >>> shapely.get_x(MultiPoint([(1, 1), (1, 2)]))
        nan
    
        """
>       return lib.get_x(point, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
E       shapely.errors.GEOSException: UnsupportedOperationException: getX called on empty Point

../../miniconda3/envs/py314/lib/python3.14/site-packages/shapely/_geometry.py:298: GEOSException
```

I traced it back to a 0-area filled contour being created. When passing through `path_to_shapely()`, this converts to an empty `Point`, which fails when going into `project_geometry()`. This can be recreated directly with the path:
```python
import cartopy.crs as ccrs
from matplotlib.collections import PathCollection
from matplotlib.path import Path
import numpy as np

p = Path(np.array([[214.,  54.], [215.,  54.], [215.,  54.], [215.,  54.],
                   [214.,  54.], [214.,  54.], [214.,  54.]]),
         np.array([ 1,  2,  2,  2,  2,  2, 79], dtype=np.uint8))
coll = PathCollection([p], transform=ccrs.PlateCarree())

fig = plt.figure()
ax = fig.add_subplot(projection=ccrs.LambertConformal())
ax.add_collection(coll)
plt.show()
```

This can be fixed in `transform_path_non_affine()` by changing the projection to handle the empty point:
```python
        geom = cpath.path_to_shapely(src_path)
        if geom.is_empty:
            result = cpath.shapely_to_path(geom)
        else:
          transformed_geom = self.target_projection.project_geometry(
            geom, self.source_projection)

        result = cpath.shapely_to_path(transformed_geom)
```

That said I'm not sure if that's the best way to fix it, or even if we want to make it pass. We do currently have a situation where `path_to_shapely()` can return an empty point, which will error in `project_geometry()`. Also, given that this worked on 0.24, it's unclear if something changed in the code around `contourf()`, since I don't see any relevant changes in the transform code.

greglucas:
It looks like we didn't have a check for an empty `shapely.Point()`, but did have an empty array check on the `transform_points()` method. The error is in the attribute look-up, which we can avoid by just returning the empty point directly since it can't be transformed anyways.

closes #2635

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -833,6 +833,8 @@ class Projection(CRS, metaclass=ABCMeta):
         return getattr(self, method_name)(geometry, src_crs)
 
     def _project_point(self, point, src_crs):
+        if point.is_empty:
+            return point
         return sgeom.Point(*self.transform_point(point.x, point.y, src_crs))
 
     def _project_line_string(self, geometry, src_crs):
--- a/lib/cartopy/tests/test_crs.py
+++ b/lib/cartopy/tests/test_crs.py
@@ -314,6 +314,11 @@ def test_transform_points_empty():
     assert_array_equal(result, np.array([], dtype=np.float64).reshape(0, 3))
 
 
+def test_project_geometry_empty_point():
+    """Test with an empty shapely point."""
+    assert ccrs.PlateCarree().project_geometry(sgeom.Point()).is_empty
+
+
 def test_transform_points_outside_domain():
     """Test CRS.transform_points with out of domain arrays."""
     # Length-1 arrays error out with a bad status code, while


rcomer (approved):
Seems right.  The doc build failure is troubling but also appearing in other PRs so nothing to do with this change.

PR #2641, opened 2026-03-15
DOC: Whats new entry for v0.26

greglucas:
Let's prepare for a v0.26 release since it has been about 6 months and people have been wanting 3.14 wheels for a while now.

I went through the commits and PRs as best I could since the last release and added items here. Please let me know if I've missed anything you want in the release notes!


dopplershift:
Since we don't release very often I think it's pretty important that we address the PROJ 9.8 issues in 0.26.

rcomer:
Do we think we will be able to fix with PROJ 9.8 soon?  If not, should we consider a release with the PROJ pin?  A lot of people are raising issues that were fixed by #2667, so it would be good to get that into a release (as well as the py314 wheels).

I appreciate there are risks with pinning PROJ and letting the work on that slide.  However, Matplotlib doesn't break us without a deprecation very often, so I would hope we wouldn't be in quite the same position for the next release and could more comfortably hold it up for the PROJ work.

In another few months people will likely be asking for py315 wheels....

dopplershift:
@rcomer The problem there is that CartoPy doesn't actually depend on PROJ, it depends on PyPROJ. (The pin we have is only in `environment.yml` for conda-based CI.) So then CartoPy is beholden to whenever PyPROJ releases wheels built against PROJ 9.8.

That said, that's also the status quo. There's literally nothing we can do in the PyPI metadata to rule out PROJ 9.8.

On the other hand, #2653 seems to have decent consensus, we just need to finalize one choice and have the collective will to land that PR.

greglucas:
I've just updated this with a bunch of PRs that have landed since I originally opened this.

With the Python 3.15 updates now also added, I think we should aim to release ASAP so we get both 3.14 and 3.15 wheels out to help the community.

rcomer:
Thanks for pulling all this together!  I found a couple of feature PRs that are missing: #2627, #2609.

Also maybe Clément should get some credit for his wheels work: #2581, #2556

There was also this new contributor PR #2613

Do we add the link from the [index page](https://cartopy--2641.org.readthedocs.build/2641/whatsnew/index.html) now, or does that come later?

greglucas:
Awesome, thank you for catching all of that @rcomer. Updated now.

QuLogic:
You have a lot of items as references (single backticks) but not actually linked. They should probably be code style (double backticks) or add an initial dot to fix the reference.

greglucas:
Thanks @QuLogic, just updated those with a mix of double backticks and actual references.

rcomer (approved):

greglucas:
Do we want to make the release today or have an expectation on when we want to do it? If so I can add a date to this PR as well and then click the make release button on GitHub unless someone else wants to take on cutting a release here.

QuLogic:
If you have a bit of time, I'm mostly through reviewing the open PRs. Maybe we can target tomorrow?

greglucas:
Sounds good. I added the date to the heading for September 16th.

docs/source/whatsnew/v0.26.rst
```diff
@@ -0,0 +1,174 @@
...
+
+* Guillermo N. Leiro Arroyo fixed gridline labels on curved projection boundaries (e.g.
+  Robinson) ignoring `.right_labels`/`.left_labels` visibility settings. (:pull:`2649`)
+
+* Guillermo N. Leiro Arroyo fixed shared `.GeoAxes` (via ``sharex``/``sharey``) not propagating
+  interactive zoom to sibling subplots after features were added. (:pull:`2652`)
+
+* Ryan May fixed compatibility with PROJ 9.8, which changed how the equidistant
+  cylindrical projection handles ellipsoids and broke the assumptions `.PlateCarree`
+  relied on; the temporary upper pin on PROJ has been removed. (:pull:`2642`, :pull:`2653`)
```
QuLogic: The temporary pin was for our testing environment; I don't think it needs to be mentioned to users, or it goes in Maintenance, at most.

rcomer:
We now have an entry for the last section #2744

docs/source/whatsnew/v0.26.rst
greglucas: Removed the temporary pin mention and PR number.

greglucas:
> We now have an entry for the last section #2744

Added this to the final section.

PR #2642, opened 2026-03-15
MNT: Temporarily add an upper pin to PROJ

greglucas:
This is a shot in the dark to see if pinning `PROJ<9.8` will fix the doc build issue (and also the conda test images too).


dopplershift (approved):

dopplershift:
Merging to clean CI, but definitely don't want this to linger. It's already the default on conda-forge.

PR #2649, opened 2026-03-24
Fix gridline labels ignoring side visibility on curved projections

### LINKED ISSUE #2607: Cannot disable the right latitude labels of the gridlines
I am trying to plt the global map with the coordinate of Robinson, whereas the `gl.right_labels = False` cannot disable the latitude labels on the right side of the gridlines. The labels `60˚N` and `60˚S` still exists in the right side.
The cartopy 0.25.0, numpy 2.3.5, matplotlib 3.10.8 and python 3.14.0 are used.

```python
import cartopy.crs as ccrs

fig = plt.figure(figsize=(10, 5), dpi=300)
lon_center = 0
map_proj = ccrs.Robinson(central_longitude=lon_center)
data_proj = ccrs.PlateCarree(central_longitude=lon_center)
ax = fig.add_subplot(1, 1, 1, projection=map_proj)
ax.coastlines(linewidth=0.3)

gl = ax.gridlines(draw_labels=True, linewidth=1, color='gray', alpha=0.5, linestyle='--')
gl.top_labels = False
gl.right_labels = False
```

Image

guillermodotn:
## Rationale

On non-rectangular projections (e.g. Robinson), gridline labels on the curved boundary get classified as `"geo"` instead of a named side, bypassing `right_labels=False` (or equivalent). 

For example, 60°N and 60°S still appear on the right even when `gl.right_labels = False`.

Closes #2607

**After fix sample:**
image

## Implications

- Single condition expanded in `_draw_gridliner`: when a label's `loc` is `"geo"`, also check the angle-derived side attribute. Uses `is False` to avoid changing default behavior for users who haven't explicitly disabled a side.
- No additional tests added.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -942,6 +942,15 @@ class Gridliner(matplotlib.artist.Artist):
                     # Is this kind label allowed to be drawn?
                     if not self._draw_this_label(xylabel, loc):
                         visible = False
+                    # For "geo" labels, also check against the
+                    # angle-derived side so that e.g.
+                    # right_labels=False is respected on curved
+                    # projections where labels cannot be classified
+                    # via spine geometry.
+                    elif loc == 'geo' and not getattr(
+                            self, self._get_loc_from_angle(
+                                segment_angle) + '_labels'):
+                        visible = False
 
                     elif x_inline or y_inline:
                         # Check that it does not overlap the map.
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -605,3 +605,20 @@ def test_gridliner_with_globe():
     fig.draw_without_rendering()
 
     assert gl in ax.artists
+
+
+def test_gridliner_geo_labels_respect_side_visibility():
+    fig = plt.figure()
+    ax = fig.add_subplot(1, 1, 1, projection=ccrs.Robinson())
+    ax.set_global()
+    gl = ax.gridlines(draw_labels=True)
+    gl.ylocator = mticker.FixedLocator([-60, -30, 0, 30, 60])
+
+    fig.draw_without_rendering()
+    labels = [a.get_text() for a in gl.geo_label_artists if a.get_visible()]
+    assert labels == ['60°S', '60°S', '60°N', '60°N']
+
+    gl.right_labels = False
+    fig.draw_without_rendering()
+    labels = [a.get_text() for a in gl.geo_label_artists if a.get_visible()]
+    assert labels == ['60°S', '60°N']


adrifoster:
Is there a way to fix this without having to update the version once this comes in? I'm trying to plot something and I am having trouble getting the 60N and 60S to disappear:

```
    fig, axes = plt.subplots(nrows, ncols, figsize=(width, height),
                            subplot_kw=dict(projection=ccrs.Robinson()),
                            layout="compressed")
                            
    for i, da in enumerate(das):

        map_function(axes[i], da, titles[i], cmap=cmap, vmin=-0.5, vmax=len(cols)-0.5)
        
        gls = axes[i].gridlines(draw_labels=True, dms=True, x_inline=False,
                          y_inline=False)
        gls.top_labels = False
        gls.right_labels = False
        
        if i%ncols > 0:
            gls.left_labels = False
        if i < (len(axes) - ncols):
            gls.bottom_labels = False
    
        gls.ylocator = FixedLocator([-60, -30, 0, 30, 60])
        gls.xlabel_style = {'size': 10}
        gls.ylabel_style = {'size': 10}
        axes[i].set_extent([-180, 180, -60, 89], crs=ccrs.PlateCarree())
        
        ```

**Comment by guillermodotn** [comment]
@adrifoster Yes, you could likely patch it manually, but I’d recommend installing directly from my fork or this PR branch instead:
```
pip install git+https://github.com/guillermodotn/cartopy.git@main
```

Once `cartopy` releases a new version with this fix just do:
```
pip install --upgrade cartopy
```
and no changes needed in your code

File: lib/cartopy/mpl/gridliner.py [hunk 1]
```diff
@@ -940,7 +940,15 @@ def update_artist(artist, renderer):
...
-                    if not self._draw_this_label(xylabel, loc):
+                    # For "geo" labels, also check against the
+                    # angle-derived side so that e.g.
+                    # right_labels=False is respected on curved
+                    # projections where labels cannot be classified
+                    # via spine geometry.
+                    if not self._draw_this_label(xylabel, loc) or (
+                            loc == 'geo'
+                            and getattr(self, self._get_loc_from_angle(
+                                segment_angle) + '_labels') is False):
```
greglucas: I feel like this reads like a separate check and should therefore be a separate elif block here with a different comment associated with that block. I don't feel strongly about that though and it could be left as an or too.

Rather than `is False` it should be `not getattr`


```suggestion
                    if not self._draw_this_label(xylabel, loc):
                        visible = False
                    elif loc == 'geo' and not getattr(self, self._get_loc_from_angle(
                                segment_angle) + '_labels')):
                        visible = False
```

greglucas (commented):
Looks right to me, just a minor nit in formatting.

lib/cartopy/mpl/gridliner.py
guillermodotn: Changes squashed into commit.

greglucas:
I forgot to mention that a test for this would be nice. It doesn't need to be an image test, I'm just thinking of something simple where you compare against the number of expected labels that appear after a draw, something like here but for your example case.

https://github.com/SciTools/cartopy/blob/05e5c2056a3a6dfa1edddd5b676a49d5380f95bb/lib/cartopy/tests/mpl/test_gridliner.py#L592-L597

guillermodotn:
Thanks for the snippet! Added a test following that pattern.

greglucas (approved):
Looks good. Thank you!

greglucas:
test failures are unrelated.

Issue #2650, opened 2026-03-25
_rings_to_multi_polygon produces inside-out polygons for features crossing the ObliqueMercator cut line


jackbenn:
# Description

When projecting Natural Earth land (or ocean) features into an ObliqueMercator projection whose cut line passes through large land masses, some polygons are returned as their own complement — a shape covering ~80–95% of the projection domain with a small hole where the land should be. The result is that ocean areas render as land color (tan).

# Reproducer

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import cartopy.feature as cfeature

# ObliqueMercator along the Seattle–Tokyo great-circle arc.
# The cut line (antipodal great circle) passes through Alaska and
# eastern Russia, so several large land polygons cross it.
projection = ccrs.ObliqueMercator(
    central_longitude=-161.07,
    central_latitude=54.55,
    azimuth=90.0,
    scale_factor=1,
)

fig, axes = plt.subplots(1, 2, subplot_kw={'projection': projection},
                         figsize=(14, 4))
fig.suptitle('Inside-out polygon bug — ObliqueMercator over North Pacific')

for ax, label, x_half, y_half in [
    (axes[0], 'larger extent (less affected)',    6_460_000, 1_960_000),
    (axes[1], 'smaller extent (ocean = tan bug)', 5_070_000, 1_170_000),
]:
    ax.set_facecolor(cfeature.COLORS['water'])
    ax.add_feature(cfeature.LAND)
    ax.add_feature(cfeature.COASTLINE)
    ax.set_extent([-x_half, x_half, -y_half, y_half], crs=projection)
    ax.set_title(label)

plt.tight_layout()
plt.show()
```

The right-hand axes shows the ocean rendered as tan (land color): <img width="1400" height="400" alt="Image" src="https://github.com/user-attachments/assets/d80a913c-74a5-4858-83d5-7c4c61182b49" />

The bug is triggered when AdaptiveScaler selects 50m data (which happens when the map extent is small enough — roughly < 50° in the shortest dimension). It does not affect 110m data.

# Root cause

[...]

PR #2651, opened 2026-03-25
Fix inside-out polygon artifact in ObliqueMercator projection

jackbenn:
## Rationale

When a polygon crosses the projection cut line (the antipodal great circle), _attach_lines_to_boundary can produce an exterior ring that traces the *complement* of the intended shape while still carrying the correct CCW/CW winding flag.  The ring is therefore classified as exterior by `_rings_to_multi_polygon`, and `sgeom.Polygon(exterior_ring)` produces a polygon covering ~80-95% of the projection domain — causing ocean areas to render as land colour.

The fix: before processing each exterior ring in the normal holes-slurping loop, check whether the ring's polygon area exceeds half the projection domain area.  If it does, the ring is inverted; replace it with `boundary_poly.difference(ring_polygon)` to recover the correct shape and continue to the next ring.

Regression test added in `TestMisc.test_oblique_mercator_no_inverted_land` using 50m Natural Earth land data (the bug only manifests at 50m or finer; `AdaptiveScaler` picks 50m when the map extent is small enough to require more detail).

Fixes rendering of land/ocean features in ObliqueMercator projections over the North Pacific (e.g. a Seattle–Tokyo strip map), where several large land masses (Alaska, Chukotka, Kamchatka) cross the cut line.

This is a fix for https://github.com/SciTools/cartopy/issues/2650

## Implications

This should have no effect beyond fixing the immediate bug.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1192,10 +1192,38 @@ class Projection(CRS, metaclass=ABCMeta):
 
         polygon_bits = []
 
+        # Precompute the projection boundary polygon for inversion detection.
+        boundary_poly = self.domain
+
         # Turn all the exterior rings into polygon definitions,
         # "slurping up" any interior rings they contain.
         for exterior_ring in exterior_rings:
             polygon = sgeom.Polygon(exterior_ring)
+
+            # When a polygon crosses the projection cut line,
+            # _attach_lines_to_boundary can produce a ring that traces the
+            # *complement* of the intended shape while still carrying the
+            # correct CCW/CW winding flag (so it is classified as an exterior
+            # ring rather than an interior one).  The tell-tale sign is an
+            # area that exceeds half the projection domain: a genuine land or
+            # ocean polygon can never be that large.  Invert such rings back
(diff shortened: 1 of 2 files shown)


lib/cartopy/crs.py
```diff
@@ -1185,27 +1185,45 @@ def _rings_to_multi_polygon(self, rings, is_ccw):
...
         for exterior_ring in exterior_rings:
             polygon = sgeom.Polygon(exterior_ring)
+
+            # When a polygon crosses the projection cut line,
+            # _attach_lines_to_boundary can produce a ring that traces the
+            # *complement* of the intended shape while still carrying the
+            # correct CCW/CW winding flag (so it is classified as an exterior
+            # ring rather than an interior one).  The tell-tale sign is an
+            # area that exceeds half the projection domain: a genuine land or
+            # ocean polygon can never be that large.  Invert such rings back
```
QuLogic, line 1209: As noted in #2650, I don't know why this assertion holds. Surely if you zoom in enough, _any_ land or ocean polygon can fill up to the _entire_ domain.

PR #2652, opened 2026-03-26
Fix shared GeoAxes zoom not propagating to sibling subplots

### LINKED ISSUE #2637: Shared axes not sharing zoom after features are added
When sharing axes - using `sharex` and `sharey` - zooming on any individual subplot does not reliably mirror the zoom on the other subplot(s), ONLY WHEN features such as coastlines or land have been added. Panning seems to work fine.

## Workaround

Calling `.set_xlim()` and `.set_ylim()` on the axes AFTER adding the feature restores the expected shared zooming behaviour.

## Demonstration

The GIF below demonstrates the following:

- Zooming on the 2nd subplot does nothing
- Pressing `back` shows the originally requested zoom on both subplots
- Zooming on the 1st subplot does nothing the 2nd subplot
- After the above sequence, zooming either subplot now correctly shares zoom with the other subplot

Image

### Code to replicate

```python
import cartopy.crs as ccrs

fig = plt.figure()
ax2 = fig.add_subplot(2, 1, 2, projection=ccrs.PlateCarree(), sharex=ax1, sharey=ax1)

ax1.add_feature(cfeature.LAND)
ax2.add_feature(cfeature.LAND)

plt.show()
```

## Idea for why

It seems that directly using Matplotlib's interface - either the API as described in `Workaround`, or the GUI buttons - can restore the intended behaviour. Perhaps whatever low-level behaviour is invoked when adding a Cartopy feature is breaking expected behaviour regarding plot limits, and further MPL calls re-establishes intended behaviour?

I was unable to reproduce using simple MPL line plots including `LineCollection`s.

## Low urgency

The above `Workaround` is simple to do. I just wanted to make sure this got reported.

guillermodotn:
## Rationale

When using `sharex`/`sharey` on cartopy GeoAxes, interactive zoom does not propagate to sibling subplots after adding features. `_draw_preprocess()` calls `autoscale_view()` on every draw, resetting all shared limits to full extent. `_get_extent_geom()` compounds this by calling `autoscale_view()` inside `hold_limits()`, which propagates resets to siblings but only restores the current axes.

issue_fix

Closes #2637

## Implications

- `_draw_preprocess()`: clear `ignore_existing_data_limits` after the first `autoscale_view()`, matching matplotlib's own pattern.
- `_get_extent_geom()`: read `viewLim` directly instead of calling `autoscale_view()`. A `dataLim` fallback handles the pre-draw case.
- No test added.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -459,6 +459,7 @@ class GeoAxes(matplotlib.axes.Axes):
         # then we should autoscale the view.
         if self.get_autoscale_on() and self.ignore_existing_data_limits:
             self.autoscale_view()
+            self.ignore_existing_data_limits = False
 
         # apply_aspect may change the x or y data limits, so must be called
         # before the patch is updated.
@@ -564,6 +565,8 @@ class GeoAxes(matplotlib.axes.Axes):
 
         self.dataLim.intervalx = self.projection.x_limits
         self.dataLim.intervaly = self.projection.y_limits
+        self.viewLim.intervalx = self.projection.x_limits
+        self.viewLim.intervaly = self.projection.y_limits
 
     def clear(self):
         """Clear the current Axes and add boundary lines."""
@@ -747,10 +750,7 @@ class GeoAxes(matplotlib.axes.Axes):
 
     def _get_extent_geom(self, crs=None):
         # Perform the calculations for get_extent(), which just repackages it.
-        with self.hold_limits():
-            if self.get_autoscale_on():
-                self.autoscale_view()
-            [x1, y1], [x2, y2] = self.viewLim.get_points()
+        [x1, y1], [x2, y2] = self.viewLim.get_points()
 
         domain_in_src_proj = sgeom.Polygon([[x1, y1], [x2, y1],
                                             [x2, y2], [x1, y2],
--- a/lib/cartopy/tests/mpl/test_axes.py
+++ b/lib/cartopy/tests/mpl/test_axes.py
@@ -154,3 +154,26 @@ def test_geoaxes_set_boundary_clipping():
                      transform=ax1.transAxes)
 
     return fig
+
+
+def test_shared_axes_zoom_propagation():
+    fig = plt.figure()
+    proj = ccrs.PlateCarree()
+    ax1 = fig.add_subplot(1, 2, 1, projection=proj)
+    ax2 = fig.add_subplot(1, 2, 2, projection=proj,
+                          sharex=ax1, sharey=ax1)
+
+    fig.draw_without_rendering()
+
+    # Simulate interactive zoom tool behavior:
+    # set_xbound/set_ybound use auto=None (no-op for autoscale)
+    # set_autoscalex/y_on(False) only affects the calling axes
+    ax1.set_xbound(-20, 20)
+    ax1.set_autoscalex_on(False)
+    ax1.set_ybound(-10, 10)
+    ax1.set_autoscaley_on(False)
+
+    fig.draw_without_rendering()
+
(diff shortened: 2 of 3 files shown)


lib/cartopy/mpl/geoaxes.py
```diff
@@ -747,10 +748,11 @@ def get_extent(self, crs=None):
...
         # Perform the calculations for get_extent(), which just repackages it.
-        with self.hold_limits():
-            if self.get_autoscale_on():
-                self.autoscale_view()
-            [x1, y1], [x2, y2] = self.viewLim.get_points()
+        [x1, y1], [x2, y2] = self.viewLim.get_points()
+        # viewLim defaults to [[0,0],[1,1]] before the first draw.
+        # Fall back to dataLim (set to projection bounds in __clear).
+        if x1 == 0 and y1 == 0 and x2 == 1 and y2 == 1:
+            [x1, y1], [x2, y2] = self.dataLim.get_points()
```
greglucas: I think your fix above is right with updating the ignore_existing_data_limits change. Can we leverage that information here?

```python
    if self.ignore_existing_data_limits:
        # viewLim not yet set (pre-draw), fall back to dataLim
        [x1, y1], [x2, y2] = self.dataLim.get_points()
    else:
        [x1, y1], [x2, y2] = self.viewLim.get_points()
```

or should we set the viewLim to a default projection value in `__clear()` like we do for the dataLim?

greglucas (commented):
Can you add a test for this?

I think you should be able to set_xlim/ylim on one axis and ensure that it updates the limits as you expect on the other one in your shared examples.

lib/cartopy/mpl/geoaxes.py
guillermodotn: Went with your second option, initialized `viewLim` to projection bounds in `__clear()`. The first option doesn't work because `ignore_existing_data_limits` stays `True` after `set_extent()`.

lib/cartopy/tests/mpl/test_set_extent.py
greglucas: This is an interesting case. I guess there is an expectation that it should be the default [[0, 0], [1, 1]] bbox per the comment... I'm not sure if we should be changing this or not. Are there any downsides to expanding the viewLim to the full projection bounds within __clear, like will it not shrink/expand in a certain case so it was better before?

The test then compares to the expanded viewLim after a draw (which would be better to trigger with a `fig.draw_without_rendering()` call rather than actually writing to a temp file while you're here).

lib/cartopy/tests/mpl/test_axes.py
```diff
@@ -154,3 +154,29 @@ def test_geoaxes_set_boundary_clipping():
                      transform=ax1.transAxes)
 
     return fig
+
+
+def test_shared_axes_zoom_propagation():
```
greglucas, line 159: This test doesn't fail on main for me. Is there a test that you can add which fails on main to demonstrate the zoom behavior that is buggy?

guillermodotn: To be honest I didn't check for that. I will look into it.

lib/cartopy/tests/mpl/test_set_extent.py
guillermodotn: I don't see any downsides. `viewLim` gets the same value that `autoscale_view()` would set on the first draw.

Replaced `plt.savefig` with `draw_without_rendering()`.

lib/cartopy/tests/mpl/test_axes.py
guillermodotn: Now fails on `main`. Dropped `coastlines()` to avoid network downloads, the bug reproduces without it.

greglucas (approved):

greglucas:
Thanks @guillermodotn!

PR #2653, opened 2026-03-27
Fixing with Proj 9.8

### LINKED ISSUE #2634: [TST] Upcoming dependency test failures
The build with nightly wheels from matplotlib, pyproj, scipy, shapely and
their dependencies has failed. Check the logs for any updates that need to
be made in cartopy.
https://github.com/SciTools/cartopy/actions/runs/22881338352

### LINKED ISSUE #2645: Inconsistent transformations between data and features with PROJ 9.8
Probably related to #2634.

I went on Google Maps and took the longitude and latitude [of a spot near the point of an island](https://www.google.com/maps/place/46%C2%B051'51.6%22N+71%C2%B005'58.3%22W/@46.8643316,-71.1021009,888m/data=!3m2!1e3!4b1!4m13!1m8!3m7!1s0x4cb8bf70c5961be7:0x8e43cee0f02d62b9!2sPit-chez-Gosselin,+QC+G0A+3Z0!3b1!8m2!3d46.860833!4d-71.0830559!16s%2Fg%2F1tknk14b!3m3!8m2!3d46.864328!4d-71.099526?entry=ttu&g_ep=EgoyMDI2MDMxNS4wIKXMDSoASAFQAw%3D%3D) near where I live. The exact spot isn't important , I just wanted to have something we can easily find on a map.

Then I plotted an area around that spot, adding the features for land and coastlines. The first plot is done in a projection of LambertConformal using a spherical globe and the second one is done with LambertConformal but with the default globe, which has a different radius than the previous one.

My issue is that the result is different.

```python3
import cartopy.crs as ccrs
import cartopy.feature as cfeat

sph = ccrs.Globe(ellipse='sphere', semimajor_axis=6370997)
# The exact same projection as the Land and Coastline data, this is taken from their `*.prj` file.
LL = ccrs.Projection('GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137.0,298.257223563]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]]')
# Spherical versions of the projections
LCs = ccrs.LambertConformal(globe=sph)
PCs = ccrs.PlateCarree(globe=sph)
# Ellipsoidal versions
PCe = ccrs.PlateCarree()
LCe = ccrs.LambertConformal(globe=PCe.globe)
```

Map 1  : Spherical version of Lambert Conformal
```python3
fig, ax = plt.subplots(subplot_kw={'projection': LCs})
ax.set_extent((1536000, 1973000, 856000, 1249000), crs=LCs)

# The actual point
ax.plot(-71.1021009, 46.8643316, 'ro', transform=LL)
# What if I cheat ?
ax.plot(-71.1021009, 46.8643316, 'bo', transform=PCe)

ax.add_feature(cfeat.LAND, color='gray')
ax.coastlines()
```

Image

Map 2 : Ellipsoidal version of Lambert Conformal
```python3
fig, ax = plt.subplots(subplot_kw={'projection': LCe})
ax.set_extent((1536000, 1973000, 856000, 1249000), crs=LCe)

# The actual point
ax.plot(-71.1021009, 46.8643316, 'ro', transform=LL)
# What if I cheat ?
ax.plot(-71.1021009, 46.8643316, 'bo', transform=PCe)

ax.add_feature(cfeat.LAND, color='gray')
ax.coastlines()
```

Image

I think map 1 is incorrect. Assuming Google Maps gives me a geographical coordinate and not a projected one, I would assume the transformation to work in the same way for my point than for the features. Here it appears to be done differently ?

The blue point is if I assume the input coordinates are actually in the EQC projection **with an ellipsoid**. This is the new behaviour that comes with PROJ 9.8. Previously, the default `PlateCarree` was an EQC projection but with a spherical earth. The blue point should be wrong in both maps, no ?

### LINKED ISSUE #2708: Coastlines drawn ~0.19° south when another GeoAxes with a different central_longitude is drawn first in the same process
## Description

When a figure (or process) draws coastlines on a `PlateCarree` GeoAxes with one
`central_longitude`, any coastlines subsequently drawn on a GeoAxes with a
**different** `central_longitude` are displaced in **latitude** by up to
~0.19°. The displacement matches the geodetic-vs-geocentric latitude
difference (≈ 0.19°·sin(2φ), largest in mid-latitudes), suggesting a wrong
transformation pipeline is being selected.

Only the feature/geometry pipeline (`FeatureArtist` → `project_geometry`) is
affected: gridlines and artists drawn with `transform=ccrs.PlateCarree(...)`
(scatter, plot) remain correct, so the resulting maps show coastlines that are
inconsistent with the graticule and with plotted data points.

Once it happens, the wrong paths are frozen in
`FeatureArtist._geom_key_to_path_cache` (keyed by projection *value*), so every
later figure in the same process that uses an equal projection silently reuses
the displaced coastlines. It affects figure that contains two map panels with different central longitudes.

## Code to reproduce

```python
import matplotlib
matplotlib.use("Agg")
import numpy as np
import cartopy.crs as ccrs
from cartopy.mpl.feature_artist import FeatureArtist

POISON = True  # set False and rerun: the printed mean changes by ~0.19 deg

fig = plt.figure(figsize=(11, 4.5))
if POISON:
    ax1 = fig.add_subplot(121, projection=ccrs.PlateCarree(central_longitude=180))
    ax1.coastlines()
ax2 = fig.add_subplot(122, projection=ccrs.PlateCarree(central_longitude=130.9))
ax2.set_extent([-5.9, 4.1, 29.0, 39.0],
               crs=ccrs.PlateCarree(central_longitude=130.9))
ax2.coastlines()
gl = ax2.gridlines(draw_labels=True, linestyle=":")
fig.canvas.draw()

# mean latitude of the coastline vertices that were actually drawn on ax2
# (read back from the private path cache, for diagnosis only)
ys = [p.vertices[:, 1]
      for mapping in FeatureArtist._geom_key_to_path_cache.values()
      for proj, p in mapping.items()
      if proj is ax2.projection and len(p.vertices)]
print(f"POISON={POISON}: mean latitude of drawn coastline vertices on ax2 = "
      f"{np.concatenate(ys).mean():.5f}")
fig.savefig(f"repro_poison_{POISON}.png", dpi=110)
```

## Actual output

```
POISON=True : mean latitude of drawn coastline vertices on ax2 = 34.44199
POISON=False: mean latitude of drawn coastline vertices on ax2 = 34.63263
```

With `POISON = True` every coastline vertex on `ax2` is ~0.19° south of its
true position (compare e.g. the position of the Korean coast or the Kanmon
strait against the gridline labels in the saved PNG). With `POISON = False`
(remove the first axes, or just its `coastlines()` call) the coastlines are
correct. `ax1` itself is always drawn correctly — only the *second* central
longitude used in the process is affected.

## Visual demonstration

The attached figure `repro_overlay.png` shows the displacement directly: on
top of the poisoned `ax2` (black coastlines from `ax2.coastlines()`), the same
Natural Earth coastline is re-drawn in red via the ordinary Matplotlib
transform path, which is **not** affected by the bug:

```python
for geom in cfeature.COASTLINE.with_scale("10m").geometries():
    # (clipped to the map window)
    xy = np.asarray(line.coords)
    ax2.plot(xy[:, 0], xy[:, 1], transform=ccrs.PlateCarree(),
             color="red", lw=0.6)
```

The black (FeatureArtist) coastline runs parallel to the red (true) one,
uniformly ~0.19° to the south across the whole map. The same red-vs-black
overlay drawn without the first axes coincides exactly.

## Expected output

Identical coastline geometry on `ax2` regardless of what was drawn before it;
drawing order within a process should not change where coastlines land.

## Additional diagnostics

- The wrong result is produced only **while the Matplotlib draw is running**.
  Calling `ax2.projection.project_geometry(geom, feature_crs)` manually right
  after `fig.canvas.draw()` (same projection instance, same geometry) usually
  returns the correct coordinates. The wrong vertices are the ones computed
  inside `FeatureArtist.draw` and then cached.
- Instrumenting `cartopy.trace.project_linear` shows that during a poisoned
  draw the transformer used for `(PlateCarree(0) → PlateCarree(130.9))` is

  ```
  desc: Ballpark geographic offset from WGS 84 to unknown + unknown
  def : proj=pipeline step proj=unitconvert xy_in=deg z_in=m xy_out=rad ...
        step proj=eqc lat_ts=0 lat_0=0 lon_0=130.9 x_0=0 ...
  ```

  i.e. the *source* `PlateCarree(0)` coordinates are being treated as WGS84
  geographic degrees, whereas outside the draw (and in a clean process) the
  correct pipeline is selected:

  ```
  def : proj=pipeline step proj=unitconvert xy_in=111319.490793274 ...
        step inv proj=eqc lat_ts=0 lat_0=0 lon_0=0 ...
        step proj=eqc ... lon_0=130.9 ...
  ```

- The two projections compare/hash as different
  (`PlateCarree(130.9) == PlateCarree(180)` is `False`), so this does not look
  like a simple cache-key collision at the Python level.
- Clearing `FeatureArtist._geom_key_to_path_cache`,
  `cartopy.trace._interpolator` (lru) and
  `cartopy.crs._get_transformer_from_crs` (lru) between renders does **not**
  help — the next draw recomputes the same wrong result.
- The Natural Earth source geometries are not mutated (verified by
  snapshotting coordinates before/after).
- We could not reproduce it without the Matplotlib draw stack: calling
  `project_geometry` over the full coastline set for central_longitude A and
  then B (including reverse `transform_points` calls mimicking `get_extent`)
  in a plain script always yields correct results.
- Also reproducible with two sequential single-axes figures in one process
  (draw figure at central_longitude A, close it, draw figure at B), which is
  how we originally encountered it.

## Operating system

macOS 15.7.5 (arm64)

## Version info

```
python     : 3.12.13
cartopy    : 0.25.0
matplotlib : 3.10.8
pyproj     : 3.7.2 (PROJ 9.8.0)
shapely    : 2.1.2
numpy      : (conda-forge / miniforge environment)
```

Image

Image

dopplershift:
As noted in #2634, Proj 9.8 added support for properly handling ellipsoids to the equidistant cylindrical projection (OSGeo/PROJ#4656). This breaks assumptions we make in PlateCarree, where we try to essentially make it treat lat/lon as unchanged Cartesian coordinates; this is accomplished by undoing the scaling by the Earth's radius (and converting radians back to degrees).

This PR attempts to update PlateCarree to correctly function with Proj 9.8 by avoiding scaling hacks altogether and just use the `latlong` projection to accomplish what we intend with PlateCarree. In order to handle the offsets that were previously supported, we use the `pm` (Prime Meridian) PROJ parameter. (Using `lon_0` in this case created problems with finding transformations in some cases because this ends up encoded as a remark.) `pm` causes some challenges in that it ends up attached to the datum, which violates some assumptions we were previously making.

This breaks some tests, but nothing that seems to be critical. One oddity is that some clipping around the edges of plots seem to be working less well for reasons I haven't identified.


lib/cartopy/crs.py
```diff
@@ -1414,7 +1411,7 @@ def quick_vertices_transform(self, vertices, src_crs):
         if return_value is None and isinstance(src_crs, PlateCarree):
             self_params = self.proj4_params.copy()
             src_params = src_crs.proj4_params.copy()
-            self_params.pop('lon_0'), src_params.pop('lon_0')
+            self_params.pop('pm'), src_params.pop('pm')
```
greglucas: this confused me at first thinking there we missed the equals sign, maybe make it two lines?

lib/cartopy/crs.py
```diff
@@ -1440,6 +1437,12 @@ def quick_vertices_transform(self, vertices, src_crs):
 
         return return_value
 
+    def as_geodetic(self):
+        from pyproj import CRS as P_CRS
```
greglucas, line 1442: Can you use `_CRS` which is already imported at the top of this file?

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -11,7 +11,7 @@
 import matplotlib.ticker as mticker
 import numpy as np
 import pytest
-from shapely.geos import geos_version
+from shapely import geos_version
```
greglucas, line 14: Thank you! I'd been meaning to remove this due to the spurious warnings we've gotten.

greglucas (commented):
I think this makes sense for Natural Earth at a minimum
https://www.naturalearthdata.com/features/

```
Projection: Natural Earth’s WGS84 projection is specified in shapefile’s PRJ file. It can also be described as a PROJ4 code [EPSG:4326](http://spatialreference.org/ref/epsg/wgs-84/) string “+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs”.
```

The thing that I've always wondered though is if we shouldn't just make a `class LatLong()` instead (or in addition to as an alternative spelling)?

Regarding your question about clipping being worse, do you know if this takes the geodesic interpolator path because pyproj thinks even the native "projection" definition is geodetic?

lib/cartopy/crs.py
dopplershift: Unfortunately, we also do `from pyproj.crs import CustomConstructorCRS as _CRS` which yields some `NotImplementedError`s when I try to use it here.

lib/cartopy/crs.py
dopplershift: Done.

dopplershift:
I agree we should also have a `LatLong` class to do this. If we do so, I would consider not having that class support `lon_0`, because that offset has created a lot of headaches trying to move from our old implementation to just using `latlon`. (See the dropping `pm` hackery I had to do here.) I would want to see what other ways we can use to implement the shifting people use `lon_0` for with `PlateCarree`. In that event I would (probably with a quiet warning) deprecate `PlateCarree` and add a separate `EquidistantCylindrical` class. IMO, we've just been conflating too many things in `PlateCarree` for that to be salvageable API-wise.

No, it doesn't trigger Cartopy's geodesic interpolation (`SphericalInterpolator`), because while PROJ returns `True` for `is_geographic`, being a subclass of `Projection` means `is_geodetic()` for `PlateCarree` will always return `False`.

Test failure notes:
* I need to dig into the tests failures more, especially for `pcolormesh`. Since we allow a sizable tolerance, I'm not sure what changes represent only a change against the (long-term) baseline versus what this PR changes against `main`.
* A lot of the changes to the gridlining/labelling look like improvements.
* `test_gridliner_title_adjust` shows some shifts of gridlines relative to the coastlines for the Mercator panel. I want to understand what's causing that better.
* Not sure why `test_cursor_values` is getting different values

QuLogic:
> * A lot of the changes to the gridlining/labelling look like improvements.
> 
> * `test_gridliner_title_adjust` shows some shifts of gridlines relative to the coastlines for the Mercator panel. I want to understand what's causing that better.

See also my comments on #2685; grid labelling seems a bit fragile in some ways, but that's a bigger problem to fix.

QuLogic:
> Not sure why `test_cursor_values` is getting different values

This test is a bit weird; it uses default `PlateCarree` and tries to format [-181.5, 50]. That x coordinate is outside of the normal boundaries of longitudes. So for `latlong` over `eqc`, it seems to wrap that to the more "correct" longitude of +178.5 when we transform it from geodetic.

QuLogic:
> I need to dig into the tests failures more, especially for `pcolormesh`. Since we allow a sizable tolerance, I'm not sure what changes represent only a change against the (long-term) baseline versus what this PR changes against `main`.

A lot of these seem to have been increased to allow newer and newer PROJ. Since we have moved to pyproj and dropped support for several older versions, it may be time to refresh these and reset the tolerances.

QuLogic:
> > I need to dig into the tests failures more, especially for `pcolormesh`. Since we allow a sizable tolerance, I'm not sure what changes represent only a change against the (long-term) baseline versus what this PR changes against `main`.
> 
> A lot of these seem to have been increased to allow newer and newer PROJ. Since we have moved to pyproj and dropped support for several older versions, it may be time to refresh these and reset the tolerances.

The easiest way to see what has happened with the `pcolormesh` tests is from `test_grid_labels_inline[RotatedPole]`:

which has a diff of:
image

Namely, the gridline along the 180° line is halved (which is more in line with the 0° line.) I guess there is either some trimming or similar wrap-around effect as with `test_cursor_values`.

The _same_ thing happens with the `pcolormesh` tests:
image
The lines on the left/top of the Axes are known changes from before and the reason for the existing higher tolerances (and I think they are possibly wrong as they are on the outside of the Axes boundary line, but that's something to figure out elsewhere.) The line on the right side is at 0° but the underlying map has `central_longitude=180` so it's the same antimeridian. I do think that the result is better now as the squares appear to be of closer width (12px vs 13px for before).

QuLogic:
> A lot of the changes to the gridlining/labelling look like improvements.

So in the case of `test_grid_labels_inline[Mercator]`, `test_grid_labels_inline[Miller]`, `test_grid_labels_inline[AzimuthalEquidistant]`, and `test_grid_labels_inline[AlbersEqualArea]`, as well as the Mercator subplots of `test_grid_labels`, `test_grid_labels_tight`, and `test_gridliner_title_adjust`, we've gained a 180° label. Looking at the gridliner code, it appears that all longitude lines after CRS transform are slightly different and the 180° line now intersects the map boundary, so we get a new potential label site, which seems to fit without overlap, and gets added to the plot. My personal opinion is also that these are improvements.

> `test_gridliner_title_adjust` shows some shifts of gridlines relative to the coastlines for the Mercator panel. I want to understand what's causing that better.

This may be because tight layout changed the positioning slightly due to the top-right map gaining its 180° labels. The Mercator subplot in the new image is 1 pixel taller than before. I feel like if you try to overlay the other maps, you see a similar issue because they're all a pixel or two different in shape.

QuLogic:
Should I just push all the updated images here so everyone can review?

QuLogic:
I've rebased and added the changed images so that we can all have a look at them. Also posted the comparisons at https://qulogic.github.io/cartopy-proj98/fig_comparison.html

lib/cartopy/tests/mpl/baseline_images/mpl/test_examples/test_contour_label.png
rcomer, line 1: This seems better because in mpl you would expect one label per contour.

lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/test_grid_labels_inline_AlbersEqualArea.png
rcomer, line 1: More symmetrical label choices seem like a good thing to me.

lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/test_grid_labels_inline_AzimuthalEquidistant.png
rcomer, line 1: This one seems like a bugfix.

lib/cartopy/tests/mpl/baseline_images/mpl/test_web_services/test_wmts.png
rcomer, line 1: This test is xfailed, so I do not think we should update this one.  I suspect the diff is showing the failure to download the wmts data rather than the proj change.

rcomer (commented):
Apart from the wmts one, all the images look good to me.  Some of them look like improvements.

lib/cartopy/tests/mpl/baseline_images/mpl/test_web_services/test_wmts.png
QuLogic: Oh, I didn't notice that it was xfail. I don't think the diff shows a download failure, but upstream seems to have changed the images slightly. But we don't need to update that here.

dopplershift:
I'm happy to mark this ready if this is a direction we want to go. Not sure what's up with WMTS.

greglucas (commented):
Rather than messing with the prime meridian can we do something similar to the RotatedPole projection? That uses ob_tran to shift things and I think that is essentially what we want to do here, removing the latitude flexibility. should we actually be setting a WGS84 datum in the default case?

QuLogic:
Is that really any much better? Can we just merge this and look into that later?

greglucas:
> Is that really any much better?

That comment was in reference to this description from the inital PR:

>  In order to handle the offsets that were previously supported, we use the pm (Prime Meridian) PROJ parameter. (Using lon_0 in this case created problems with finding transformations in some cases because this ends up encoded as a remark.) pm causes some challenges in that it ends up attached to the datum, which violates some assumptions we were previously making.

I don't know enough about this area and transformation laws/rules in general. I also don't know how easy it will be to change later so figured I would bring it up. Maybe it is not a big deal because it is our own projection and not directly the CRS itself :shrug:

> Can we just merge this and look into that later?

Absolutely! I am not blocking anything and would prefer if we start merging things quicker.

dopplershift:
I went ahead and marked this ready for review since I think we're close. One thing that I'm curious about is why coverage decreased, given that this doesn't add any new blocks that might be unexecuted. That makes me think we no longer hit some special cases elsewhere?

rcomer:
I think I've seen a couple of PRs here lately where coverage failed and then passed on a subsequent push, with no obviously relevant change. 🤷‍♀️

rcomer (commented):
I approve all the changes under `tests`, but I do not know nearly enough about PROJ to have an opinion on the implementation.

QuLogic:
> Rather than messing with the prime meridian can we do something similar to the RotatedPole projection? That uses ob_tran to shift things and I think that is essentially what we want to do here, removing the latitude flexibility. should we actually be setting a WGS84 datum in the default case?

I tried this, but don't really understand `ob_tran` that well, and it seems to break the `pcolormesh` wrapping. It's probably easiest to just stay with what we have here.

I'm going to merge this and if we need to, can try other ideas later.

trexfeathers:
Thanks for your work on this everyone 💐

PR #2654, opened 2026-03-27
Alternative fix for Proj 9.8

dopplershift:
Alternative to #2653. Here I've forced the use of a spherical datum with the equidistant cyindrical projection, which causes our scaling to work properly again.

This does result in one test failure, caused by the fact that the forced R parameter means the datum for the projection is always spherical, which flows downstream into all calls of `PlateCarree.to_geodetic()`.

While this is a smaller change, my personal opinion is in favor of #2653 because we finally get rid of all the hacks we have been doing to try to make PlateCarree work using the eqc projection.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,10 +13,6 @@ dependencies:
   - shapely>=2.0
   - pyshp>=2.3.1
   - pyproj>=3.6.0
-  # PROJ 9.8 changes the way equicylindrical projections (PlateCarree)
-  # are transformed, which causes some issues. Temporary pin PROJ < 9.8
-  # to avoid these for now.
-  - proj<9.8
   # The testing label has the proper version of freetype included
   - conda-forge/label/testing::matplotlib-base>=3.6
 
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1353,9 +1353,8 @@ class PlateCarree(_CylindricalProjection):
     def __init__(self, central_longitude=0.0, globe=None):
         globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
         proj4_params = [('proj', 'eqc'), ('lon_0', central_longitude),
-                        ('to_meter', math.radians(1) * (
-                            globe.semimajor_axis or WGS84_SEMIMAJOR_AXIS)),
-                        ('vto_meter', 1)]
+                        ('to_meter', math.radians(1) * WGS84_SEMIMAJOR_AXIS),
+                        ('vto_meter', 1), ('R', WGS84_SEMIMAJOR_AXIS)]
         x_max = 180
         y_max = 90
         # Set the threshold around 0.5 if the x max is 180.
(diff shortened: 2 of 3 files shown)


QuLogic:
> This does result in one test failure, caused by the fact that the forced R parameter means the datum for the projection is always spherical, which flows downstream into all calls of `PlateCarree.to_geodetic()`.

We could override `to_geodetic` to return the ellipsoidal datum when no globe is specified. This is just more of a hack, but would be a smaller change.

Something hard-coded like this works, but I did not test all `Globe`s outside of what tests normally do:
```diff
diff --git a/lib/cartopy/crs.py b/lib/cartopy/crs.py
index 05ea62b1..cfae83b2 100644
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1361,6 +1361,7 @@ def _ellipse_boundary(semimajor=2, semiminor=1, easting=0, northing=0, n=201):
 
 class PlateCarree(_CylindricalProjection):
     def __init__(self, central_longitude=0.0, globe=None):
+        self._default_globe = globe is None
         globe = globe or Globe(semimajor_axis=WGS84_SEMIMAJOR_AXIS)
         proj4_params = [('proj', 'eqc'), ('lon_0', central_longitude),
                         ('to_meter', math.radians(1) * WGS84_SEMIMAJOR_AXIS),
@@ -1371,6 +1372,35 @@ class PlateCarree(_CylindricalProjection):
         self.threshold = x_max / 360
         super().__init__(proj4_params, x_max, y_max, globe=globe)
 
+    _default_globe_srs = (
+        'GEOGCRS["unknown",'
+        'DATUM["Unknown based on WGS 84 ellipsoid",'
+        'ELLIPSOID["WGS 84",{major},{inverse_flattening},'
+        'LENGTHUNIT["metre",1,ID["EPSG",9001]]]],'
+        'PRIMEM["Greenwich",0,ANGLEUNIT["degree",{degree}],ID["EPSG",8901]],'
+        'CS[ellipsoidal,3],'
+        'AXIS["longitude",east,ORDER[1],ANGLEUNIT["degree",{degree},ID["EPSG",9122]]],'
+        'AXIS["latitude",north,ORDER[2],ANGLEUNIT["degree",{degree},ID["EPSG",9122]]],'
+        'AXIS["ellipsoidal height (h)",up,ORDER[3],'
+        'LENGTHUNIT["metre",1,ID["EPSG",9001]]]]'
+    )
+
+    def as_geodetic(self):
+        """
+        Return a new Geodetic CRS with the same ellipse/datum as this
+        CRS.
+
+        """
+        return CRS(
+            self._default_globe_srs.format(
+                degree=math.radians(1),
+                major=WGS84_SEMIMAJOR_AXIS,
+                inverse_flattening=(WGS84_SEMIMAJOR_AXIS /
+                                    (WGS84_SEMIMAJOR_AXIS - WGS84_SEMIMINOR_AXIS)),
+            )
+            if self._default_globe else
+            self.geodetic_crs.srs)
+
     def _bbox_and_offset(self, other_plate_carree):
         """
         Return a pair of (xmin, xmax) pairs and an offset which can be used
```

QuLogic:
We went with #2653 for now.

Issue #2584, opened 2025-10-23
Ocean feature with scale 50m produces invalid results with EuroPP projection


mosc9575:
The ocean feature with a scale of "50m" produces an invalid result when using the `EuroPP` projection.

```python
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import cartopy.feature as cfeature

# Create a figure with the EuroPP projection
fig = plt.figure(figsize=(10, 8))
ax = plt.axes(projection=ccrs.EuroPP())

# Add ocean feature at 50m resolution
ax.add_feature(cfeature.OCEAN.with_scale('50m'))

# Optionally add coastlines for context
ax.coastlines(resolution='50m')
plt.show()
```

<img width="577" height="632" alt="Image" src="https://github.com/user-attachments/assets/cf7d5197-7df1-4c0b-9661-cba88431943e" />

The results are correct using the default scale with "110m".

This issue was first reported in https://github.com/holoviz/geoviews/issues/812 and the example was created by @hoxbro.

PR #2658, opened 2026-04-03
FIX: boundary attachment arc direction

### LINKED ISSUE #2137: Mollweide projection - completely corrupted contourf when there are gridpoints at the pole

### LINKED ISSUE #2650: _rings_to_multi_polygon produces inside-out polygons for features crossing the ObliqueMercator cut line

greglucas:
Previously `_attach_lines_to_boundary` always connected projected LineString endpoints by walking forward around the boundary, even when the shorter path went backward. For cut lines that split large polygons near the middle of a boundary edge (e.g. ObliqueMercator over Alaska/Russia), this caused the ring to trace ~98% of the perimeter instead of ~2%, producing a polygon that is the complement of the intended shape — an inside-out land or ocean feature.

Additionally, it had the possibility of adding a midpoint on the far edge boundary while doing that tracing, even though the two endpoints were not actually on the boundary. So the linestring should really be closed with itself and not attached to the boundary. So I've added in an extra check to make sure that we are only adding boundary points/attaching when we are within the threshold of the boundary, not in the middle of the domain.

This is a pretty major overhaul of the `_attach_line_to_boundary` function to try and make the intent a little more clear and improve the performance some as well. It is largely the same logic, just refactored with one fix to choose the shortest walk direction. This may fix more cases because it is a pretty low-level thing that is done a lot, but this specifically addresses #2650.

Fixes #2650

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -10,6 +10,7 @@ between them.
 """
 
 from abc import ABCMeta
+import bisect
 from collections import OrderedDict
 from functools import lru_cache
 import io
@@ -38,6 +39,9 @@ __document_these__ = ['CRS', 'Geocentric', 'Geodetic', 'Globe']
 
 WGS84_SEMIMAJOR_AXIS = 6378137.0
 WGS84_SEMIMINOR_AXIS = 6356752.3142
+
+# cache to avoid recreating in every ring call
+_EMPTY_MLS = sgeom.MultiLineString([])
 
 
 # Cache the transformer creation method
@@ -757,6 +761,33 @@ class Projection(CRS, metaclass=ABCMeta):
             domain = self._domain = sgeom.Polygon(self.boundary)
         return domain
 
+    @property
+    def _ring_threshold(self):
+        """Coordinate-match tolerance for ring stitching (1e-5 of domain scale).
+
+        Cached on first access; recomputed only if the projection bounds change.
+        """
+        try:
+            return self.__ring_threshold_cache
+        except AttributeError:
+            val = max(abs(v) for v in self.x_limits + self.y_limits) * 1e-5
+            self.__ring_threshold_cache = val
+            return val
+
+    @property
+    def _prepared_domain(self):
+        """Prepared (indexed) version of :attr:`domain`, cached per projection instance.
+
+        Used by :func:`cartopy.trace.project_linear` for fast point-in-polygon
+        tests.  The domain polygon is immutable, so a single prepared instance
+        is safe to reuse for the projection's lifetime.
+        """
+        try:
+            return self.__prepared_domain_cache
+        except AttributeError:
+            self.__prepared_domain_cache = prep(self.domain)
+            return self.__prepared_domain_cache
+
     def is_geodetic(self):
         return False
 
@@ -838,93 +869,96 @@ class Projection(CRS, metaclass=ABCMeta):
         return sgeom.Point(*self.transform_point(point.x, point.y, src_crs))
 
     def _project_line_string(self, geometry, src_crs):
-        return cartopy.trace.project_linear(geometry, src_crs, self)
+        return cartopy.trace.project_linear(geometry, src_crs, self, is_ring=False)
 
-    def _project_linear_ring(self, linear_ring, src_crs):
+    def _project_linear_ring(self, linear_ring, src_crs, is_ccw=None):
         """
-        Project the given LinearRing from the src_crs into this CRS and
-        returns a GeometryCollection containing zero or more LinearRings and
-        a single MultiLineString.
+        Project the given LinearRing from src_crs into this CRS.
 
+        When called from :meth:`_project_polygon`, ``is_ccw`` is the winding
+        of the source polygon exterior and the method returns a
+        ``(list[LinearRing], MultiLineString)`` tuple.
+
+        When called via ``project_geometry(linear_ring)`` (i.e. without
+        ``is_ccw``), the winding is inferred from the ring itself and the
+        method returns a ``GeometryCollection([*rings, MultiLineString])``.
         """
-        debug = False
-        # 1) Resolve the initial lines into projected segments
-        # 1abc
-        # def23ghi
-        # jkl41
-        multi_line_string = cartopy.trace.project_linear(linear_ring,
-                                                         src_crs, self)
+        if is_ccw is None:
+            # infer winding and produce GeometryCollection.
+            is_ccw = True if src_crs.is_geodetic() else linear_ring.is_ccw
+            rings, mls = self._project_linear_ring(linear_ring, src_crs, is_ccw)
+            return sgeom.GeometryCollection([*rings, mls])
 
-        # Threshold for whether a point is close enough to be the same
-        # point as another.
-        threshold = max(np.abs(self.x_limits + self.y_limits)) * 1e-5
+        # project_linear returns an ordered list of LineString fragments when
+        # given a LinearRing. Fragment i exits the domain boundary and
+        # fragment (i+1) % N enters it.
+        fragments = cartopy.trace.project_linear(
+            linear_ring, src_crs, self, is_ring=True
+        )
 
-        # 2) Simplify the segments where appropriate.
-        if len(multi_line_string.geoms) > 1:
-            # Stitch together segments which are close to continuous.
-            # This is important when:
-            # 1) The first source point projects into the map and the
-            # ring has been cut by the boundary.
-            # Continuing the example from above this gives:
-            #   def23ghi
-            #   jkl41abc
-            # 2) The cut ends of segments are too close to reliably
-            # place into an order along the boundary.
+        if not fragments:
+            return [], _EMPTY_MLS
 
-            line_strings = list(multi_line_string.geoms)
-            any_modified = False
-            i = 0
-            if debug:
-                first_coord = np.array([ls.coords[0] for ls in line_strings])
-                last_coord = np.array([ls.coords[-1] for ls in line_strings])
-                print('Distance matrix:')
-                np.set_printoptions(precision=2)
-                x = first_coord[:, np.newaxis, :]
-                y = last_coord[np.newaxis, :, :]
-                print(np.abs(x - y).max(axis=-1))
+        threshold = self._ring_threshold
 
-            while i < len(line_strings):
-                modified = False
-                j = 0
-                while j < len(line_strings):
-                    if i != j and np.allclose(line_strings[i].coords[0],
-                                              line_strings[j].coords[-1],
-                                              atol=threshold):
-                        if debug:
-                            print(f'Joining together {i} and {j}.')
-                        last_coords = list(line_strings[j].coords)
-                        first_coords = list(line_strings[i].coords)[1:]
-                        combo = sgeom.LineString(last_coords + first_coords)
-                        if j < i:
-                            i, j = j, i
-                        del line_strings[j], line_strings[i]
-                        line_strings.append(combo)
-                        modified = True
-                        any_modified = True
-                        break
-                    else:
-                        j += 1
-                if not modified:
-                    i += 1
-            if any_modified:
-                multi_line_string = sgeom.MultiLineString(line_strings)
+        if len(fragments) == 1:
+            ls = fragments[0]
+            if len(ls.coords) < 2:
+                return [], _EMPTY_MLS
+            c0, c1 = ls.coords[0], ls.coords[-1]
+            if abs(c0[0] - c1[0]) <= threshold and abs(c0[1] - c1[1]) <= threshold:
+                # Ring projected cleanly inside domain; no boundary stitching.
+                return [sgeom.LinearRing(ls.coords[:-1])], _EMPTY_MLS
+            # Single open fragment; needs boundary stitching.
+            return [], sgeom.MultiLineString([ls])
 
-        # 3) Check for rings that have been created by the projection stage.
-        rings = []
-        line_strings = []
-        for line in multi_line_string.geoms:
-            if len(line.coords) > 3 and np.allclose(line.coords[0],
-                                                    line.coords[-1],
-                                                    atol=threshold):
-                result_geometry = sgeom.LinearRing(line.coords[:-1])
-                rings.append(result_geometry)
+        # When the ring's first source point was inside the domain,
+        # fragments[0].start == fragments[-1].end (both map to the same interior
+        # point). Merge last + first[1:] so all fragments have both endpoints
+        # on the boundary.
+        frags = list(fragments)
+        c0, cm1 = frags[0].coords[0], frags[-1].coords[-1]
+        if abs(c0[0] - cm1[0]) <= threshold and abs(c0[1] - cm1[1]) <= threshold:
+            merged = sgeom.LineString(
+                list(frags[-1].coords) + list(frags[0].coords)[1:]
+            )
+            frags = [merged] + frags[1:-1]
+
+        # Merge consecutive fragments that share an endpoint. This happens
+        # when the ring crosses the boundary twice at nearly the same location.
+        # Adjacent matches only: the ordered fragment list guarantees that
+        # frag[i].end == frag[i+1].start is the only possible adjacent join.
+        i = 0
+        while i < len(frags) - 1:
+            ce, cs = frags[i].coords[-1], frags[i + 1].coords[0]
+            if abs(ce[0] - cs[0]) <= threshold and abs(ce[1] - cs[1]) <= threshold:
+                stitched = sgeom.LineString(
+                    list(frags[i].coords) + list(frags[i + 1].coords)[1:]
+                )
+                frags = frags[:i] + [stitched] + frags[i + 2:]
             else:
-                line_strings.append(line)
-        # If we found any rings, then we should re-create the multi-line str.
-        if rings:
-            multi_line_string = sgeom.MultiLineString(line_strings)
+                i += 1
 
-        return sgeom.GeometryCollection([*rings, multi_line_string])
+        # A fragment whose endpoints are nearly coincident is a complete ring
+        # that projected entirely within the domain — it doesn't cross the
+        # boundary and should not go to _attach_lines_to_boundary.
+        rings = []
+        open_frags = []
+        for frag in frags:
+            coords = frag.coords
+            ca, cb = coords[0], coords[-1]
+            if (len(coords) > 3
+                    and abs(ca[0] - cb[0]) <= threshold
+                    and abs(ca[1] - cb[1]) <= threshold):
+                ring = sgeom.LinearRing(coords[:-1])
+                # Discard near-zero area rings; they are boundary artefacts
+                # produced when a source vertex lies just outside the domain.
+                if sgeom.Polygon(ring).area >= threshold ** 2:
+                    rings.append(ring)
+            else:
+                open_frags.append(frag)
+
+        return rings, sgeom.MultiLineString(open_frags)
 
     def _project_multipoint(self, geometry, src_crs):
         geoms = []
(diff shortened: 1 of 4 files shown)


rcomer:
Cycling to pick up #2659

lib/cartopy/crs.py
```diff
@@ -988,198 +989,186 @@ def _project_polygon(self, polygon, src_crs):
...
+        Parameters
+        ----------
+        multi_line_strings : list of MultiLineString
+            One MultiLineString per source ring (exterior or interior).
+        is_ccw : bool
+            Expected winding of the source polygon exterior ring.
+        """
+        boundary = self.ccw_boundary if is_ccw else self.cw_boundary
+        perimeter = boundary.length
+        threshold = self.threshold
```
rcomer, line 1025: Do we have it documented anywhere what `self.threshold` is?  I have not previously known about it.

lib/cartopy/crs.py
```diff
@@ -988,198 +989,186 @@ def _project_polygon(self, polygon, src_crs):
...
+                        # on the boundary (needed for correct winding). Skip
+                        # when the arc has zero length (start == end).
+                        d_fwd = (d_ring_start - d_end) % perimeter
+                        if d_fwd > 0:
+                            if d_fwd <= perimeter / 2:
+                                d_mid = (d_end + d_fwd / 2) % perimeter
+                            else:
+                                d_mid = (d_end - (perimeter - d_fwd) / 2) % perimeter
+                            mid_pt = boundary.interpolate(d_mid)
+                            ring_coords.append((mid_pt.x, mid_pt.y))
```
rcomer, line 1122: Could this be done within `arc_corners`?  Since it already has the check on which route is shortest there wouldn't be too much to add in there.  Also I got myself a bit confused trying to map "start" <==> "to", "end" <==> "from" in my head, so having all that similar logic together might make it easier to follow.

lib/cartopy/crs.py
```diff
@@ -988,198 +989,186 @@ def _project_polygon(self, polygon, src_crs):
...
+                    if not corners and d_fwd_close > 0:
+                        # No boundary corners on the closing arc. Insert a
+                        # midpoint on the shorter arc so the ring has a point
+                        # on the boundary (needed for correct winding).
+                        if d_fwd_close <= perimeter / 2:
+                            mid_d = (d_end + d_fwd_close / 2) % perimeter
+                        else:
+                            mid_d = (d_end - (perimeter - d_fwd_close) / 2) % perimeter
+                        mid_pt = boundary.interpolate(mid_d)
+                        ring_coords.append((mid_pt.x, mid_pt.y))
```
rcomer, line 1152: Should this be skipped if the start and end are in the same place, similar to the case where there are no remaining lines?

lib/cartopy/crs.py
```diff
@@ -988,198 +989,186 @@ def _project_polygon(self, polygon, src_crs):
...
+        # have endpoints far from the boundary. Close them directly as rings
+        # rather than running them through the topology loop, which could
+        # incorrectly add a midpoint along the boundary and change the shape.
+        for k, ls in enumerate(line_strings):
+            d_start, start_on = endpoint_info(ls.coords[0])
+            d_end,   end_on   = endpoint_info(ls.coords[-1])
+            if start_on and end_on:
+                remaining_segs.append((d_start, d_end, k))
+            else:
+                completed.append(list(ls.coords))
```
rcomer, line 1087: Just to check my understanding...

I was wondering if there are cases where the line doesn't meet the boundary at both ends but also its start and end points are different, so at least one other line would need to be appended to it in order to close it.  Perhaps there are but it doesn't matter because ultimately polygons created from those lines will be sorted out as exteriors/interiors of each other? 🤔

rcomer (commented):
This certainly makes the code easier to read!  I also learned about a new module as I wasn't familiar with `bisect`.

I think I'm confused about the need to take the shortest route around the edge.

For the basic case with one linestring: if you go forward, you always get an exterior polygon, which may fill up most of the map.  If you go backwards, you get the inverse polygon, but it's identified as an interior.  So it will get [inverted](https://github.com/SciTools/cartopy/blob/ff1210f91072a6c16affaece4a36cf8527baa62b/lib/cartopy/crs.py#L1213) later on, and you end up with the same polygon as if you went forward.

So was the problem case one that should have been an interior contained within a polygon formed from a different linestring?  If so, shouldn't the "find the next edge thing" logic have found the start of that outer linestring before going all the way round?

rcomer:
This also gives a big improvement on the left panel from https://github.com/SciTools/cartopy/issues/2584#issuecomment-3510891379, though it's still not quite there

image

lib/cartopy/crs.py
greglucas: Yes it seems like there is! See this comment where I tried to allude to the ends being interior to the boundary and then our routines deciding to add a "midpoint" on the boundary on the far right edge. I think that is a case where the boundary is different than what the projection boundary is and where things get clipped within the PROJ resampling routines (but I have not dug into that case). In a few of the boundaries, we do have slight adjustments/shifts to avoid peculiarities. I am wondering if this is something related where we've buffered or shifted a boundary polygon too much.

https://github.com/SciTools/cartopy/issues/2650#issuecomment-4187096297

lib/cartopy/crs.py
greglucas: :eyes: apparently not! I thought we did, but I don't see a docstring on it immediately. It is a useful tool though when trying to make lines look smoother. I did add an example using it in my other adaptive-resampling PR because I think that illustrates the point quite well when it is symmetric and shows you what different values will do.

https://github.com/SciTools/cartopy/pull/2647/changes#diff-44b7ad8f2f0073da2576021a84484f3e4dcf524edf5ddaad9d12d0225451577b

lib/cartopy/crs.py
greglucas: Yes, let me think about this some more and how to consolidate it.

greglucas:
Thanks for adding that other example here too. I can try and investigate that specific case some more and see if there is another condition we can add to capture that as well. It again looks like something where we are closing/adding points in wrong locations.

Your comments actually got me to thinking, I wonder if we could pass some of this information into `project_linear()` so that we could tell project_linear itself that this was a ring coming in, so all outputs need to be closed by either being attached to a boundary or upon themselves. That would likely mean moving (at least some of) the attach_to_boundary logic into trace.pyx, but I'm wondering if it could simplify some of the dealing with linestrings/rings on the Python caller side because then we could say "project all polygon exteriors, then project all holes and do a difference on the results". We wouldn't have to keep track of rings/linestrings together and then determine if a string should have been a ring etc. That would be encoded in the calling logic.

I'll have a think about some of this and experiment around to see if there is a potential for any simplification, or if it is really just pushing the same complicated logic around to a different place.

lib/cartopy/crs.py
rcomer: Oh, I also just realized that the case of needing to append two lines that end away from the boundary is already handled in `_project_linear_ring`.

https://github.com/SciTools/cartopy/blob/1f15ef0836f18d005fe11cfbc3a184545b33798b/lib/cartopy/crs.py#L890-L897

rcomer:
I tried rebasing this against main, and found that it additionally fixes #2137, so I guess that needs a combination of this and #2665

image

rcomer:
I also found that it solves the error from the second case at https://github.com/SciTools/cartopy/issues/2176#issuecomment-2772601094, but the plot is probably not what the user wants
image

lib/cartopy/crs.py
```diff
@@ -757,6 +761,33 @@ def domain(self):
...
+        try:
+            return self.__ring_threshold_cache
+        except AttributeError:
+            val = max(abs(v) for v in self.x_limits + self.y_limits) * 1e-5
+            self.__ring_threshold_cache = val
+            return val
+
+    @property
+    def _prepared_domain(self):
+        """Prepared (indexed) version of :attr:`domain`, cached per projection instance.
```
QuLogic, line 779: See #2698; you should just use `shapely.prepare` and not bother with caching it yourself.

lib/cartopy/crs.py
```diff
@@ -757,6 +761,33 @@ def domain(self):
...
+    @property
+    def _ring_threshold(self):
+        """Coordinate-match tolerance for ring stitching (1e-5 of domain scale).
+
+        Cached on first access; recomputed only if the projection bounds change.
+        """
+        try:
+            return self.__ring_threshold_cache
+        except AttributeError:
+            val = max(abs(v) for v in self.x_limits + self.y_limits) * 1e-5
```
QuLogic, line 773: Do we not want to be looking at the difference in limits? If the limits were (1e7, 1e7+100), then this threshold would be bigger than the whole map.

lib/cartopy/trace.pyx
```diff
@@ -541,30 +579,43 @@ def project_linear(geometry not None, src_crs not None,
...
+        `~shapely.geometry.LineString` fragments (ring-traversal order);
+        ``False`` returns a `~shapely.geometry.MultiLineString`.
+        Defaults to ``False``.
 
     Returns
     -------
-    `shapely.geometry.MultiLineString`
-        The result of projecting the given geometry from the source projection
-        into the destination projection.
+    `shapely.geometry.MultiLineString` or list of `shapely.geometry.LineString`
```
QuLogic, line 591:
```suggestion
    `shapely.MultiLineString` or list of `shapely.LineString`
```

lib/cartopy/trace.pyx
```diff
@@ -541,30 +579,43 @@ def project_linear(geometry not None, src_crs not None,
         The coordinate system of the line to be projected.
     dest_projection : cartopy.crs.Projection
         The projection for the resulting projected line.
+    is_ring : bool, optional
+        Set to ``True`` when *geometry* is a closed ring.  Controls the
+        return type: ``True`` returns an ordered list of
+        `~shapely.geometry.LineString` fragments (ring-traversal order);
+        ``False`` returns a `~shapely.geometry.MultiLineString`.
+        Defaults to ``False``.
```
QuLogic, line 587:
```suggestion
        Set to ``True`` when *geometry* is a closed ring. Controls the return type:

        - ``True`` returns an ordered list of `~shapely.LineString` fragments
          (ring-traversal order);
        - ``False`` returns a `~shapely.MultiLineString`.

        Defaults to ``False``.
```

lib/cartopy/trace.pyx
```diff
@@ -541,30 +579,43 @@ def project_linear(geometry not None, src_crs not None,
...
-    g_domain = dest_projection.domain
-
     interpolator = _interpolator(src_crs, dest_projection)
 
     src_coords = np.asarray(geometry.coords)
     dest_coords = interpolator.project_points(src_coords)
-    gp_domain = sprep.prep(g_domain)
+    # Use the cached prepared domain from the projection – avoids rebuilding the
+    # prepared geometry (sprep.prep()) on every ring projection call.
+    gp_domain = dest_projection._prepared_domain
```
QuLogic, line 618: See above, re. prepared geometries.

lib/cartopy/crs.py
```diff
@@ -838,93 +869,96 @@ def _project_point(self, point, src_crs):
...
+    def _project_linear_ring(self, linear_ring, src_crs, is_ccw=None):
+        """
+        Project the given LinearRing from src_crs into this CRS.
+
+        When called from :meth:`_project_polygon`, ``is_ccw`` is the winding
+        of the source polygon exterior and the method returns a
+        ``(list[LinearRing], MultiLineString)`` tuple.
+
+        When called via ``project_geometry(linear_ring)`` (i.e. without
+        ``is_ccw``), the winding is inferred from the ring itself and the
```
QuLogic, line 883: It seems to depend on `src_crs` as well?

lib/cartopy/crs.py
```diff
@@ -838,93 +869,96 @@ def _project_point(self, point, src_crs):
...
+        # When the ring's first source point was inside the domain,
+        # fragments[0].start == fragments[-1].end (both map to the same interior
+        # point). Merge last + first[1:] so all fragments have both endpoints
+        # on the boundary.
+        frags = list(fragments)
+        c0, cm1 = frags[0].coords[0], frags[-1].coords[-1]
+        if abs(c0[0] - cm1[0]) <= threshold and abs(c0[1] - cm1[1]) <= threshold:
+            merged = sgeom.LineString(
+                list(frags[-1].coords) + list(frags[0].coords)[1:]
+            )
```
QuLogic, line 924:
```suggestion
            merged = sgeom.LineString([*frags[-1].coords, *frags[0].coords[1:]])
```

lib/cartopy/crs.py
```diff
@@ -838,93 +869,96 @@ def _project_point(self, point, src_crs):
...
+        # Adjacent matches only: the ordered fragment list guarantees that
+        # frag[i].end == frag[i+1].start is the only possible adjacent join.
+        i = 0
+        while i < len(frags) - 1:
+            ce, cs = frags[i].coords[-1], frags[i + 1].coords[0]
+            if abs(ce[0] - cs[0]) <= threshold and abs(ce[1] - cs[1]) <= threshold:
+                stitched = sgeom.LineString(
+                    list(frags[i].coords) + list(frags[i + 1].coords)[1:]
+                )
+                frags = frags[:i] + [stitched] + frags[i + 2:]
```
QuLogic, line 938:
```suggestion
                merged = sgeom.LineString([*frags[i].coords, *frags[i + 1].coords[1:]])
                frags[i:i + 2] = [merged]
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+        perimeter = boundary.length
+        threshold = self.threshold
+
+        # Pre-compute boundary corner (distance, coord) pairs once, sorted
+        # for O(log N) bisect queries inside fwd_arc_corners().
+        corner_pairs  = sorted(
+            (boundary.project(sgeom.Point(c)), (c[0], c[1]))
+            for c in boundary.coords[:-1]
+        )
+        corner_dists  = [d for d, _ in corner_pairs]
```
QuLogic, line 1042:
```suggestion
        corner_pairs = sorted(
            (boundary.project(sgeom.Point(c)), (c[0], c[1]))
            for c in boundary.coords[:-1]
        )
        corner_dists = [d for d, _ in corner_pairs]
```

QuLogic: Also, a boundary isn't necessarily a box or polygon, so I don't know why these variables are called `corner_*`. AFAICT, `corner_pairs` is basically the path length for each point on the geometry. Since that's calculated in order, why would it need sorting?

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
-                    # first boundary thing.
-                    if debug:
-                        print('   close loop')
-                    processed_ls.append(current_ls)
-                    if debug_plot_edges:
-                        coords = np.array(current_ls.coords)
-                        ax.plot(coords[:, 0], coords[:, 1], color='black',
-                                linestyle='--')
-                    break
+                completed.append(list(ls.coords))
```
QuLogic, line 1097: Minor clarification.
```suggestion
            d_start, start_on_boundary = endpoint_info(ls.coords[0])
            d_end, end_on_boundary = endpoint_info(ls.coords[-1])
            if start_on_boundary and end_on_boundary:
                remaining_segs.append((d_start, d_end, k, mls_idx_of_seg[k]))
            else:
                completed.append([*ls.coords])
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+        # and whose closed form has the wrong winding is a PROJ artefact from
+        # a source ring that lies entirely outside our domain. Keeping it
+        # would cause _rings_to_multi_polygon to invert it into a flood polygon.
+        #
+        # Build the set of mls_i values that appear exactly once (singletons).
+        # A source ring that produced only one boundary segment is the typical
+        # signature of a PROJ coordinate-wrapping artefact.
+        _seen, _multi = set(), set()
+        for _, _, _, mls_i in remaining_segs:
+            (_multi if mls_i in _seen else _seen).add(mls_i)
```
QuLogic, line 1116: No need to be obscure here.
```suggestion
        _seen = set()
        _multi = set()
        for _, _, _, mls_i in remaining_segs:
            if mls_i in _seen:
                _multi.add(mls_i)
            else:
                _seen.add(mls_i)
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+        singleton_mls = _seen - _multi
+        remaining_segs = [
+            (d_s, d_e, k, mls_i) for d_s, d_e, k, mls_i in remaining_segs
+            if not (mls_i in singleton_mls and d_e < d_s
+                    and _ring_is_ccw(
+                        sgeom.LinearRing(
+                            list(line_strings[k].coords) + arc_corners(d_e, d_s)
```
QuLogic, line 1123:
```suggestion
                        shapely.LinearRing(
                            [*line_strings[k].coords, *arc_corners(d_e, d_s)]
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+                        )
+                    ) != is_ccw)
+        ]
+        remaining_segs_d = [t[0] for t in remaining_segs]
+
+        def finish_ring(d_end_local, ring_coords_local, d_fwd_local,
+                        d_ring_start_local):
+            """Close ring_coords with the shorter arc back to d_ring_start_local."""
```
QuLogic, line 1131:
```suggestion
            """
            Close ring_coords_local with the shorter arc back to d_ring_start_local.
            """
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+                    # Connect to segment j via forward boundary arc, then
+                    # continue building the ring.  Using fwd_arc_corners
+                    # (not arc_corners) guarantees that corners are inserted
+                    # in the forward direction, preserving the ring's winding.
+                    ring_coords += fwd_arc_corners(d_end, d_fwd_next)
+                    ring_coords += list(line_strings[j].coords)
+                    remaining_segs.pop(pos)
+                    remaining_segs_d.pop(pos)
+                    d_end = d_next_end
+                    current_mls = j_mls  # track mls of most recently added seg
```
QuLogic, line 1203: Since the previous clause broke out of the loop, can dedent this half to reduce the wrapping.
```suggestion

                # Segment j comes before our ring start. If j is from a different source
                # ring and the gap to it is large, only connect if the ring can't
                # otherwise close without a very long boundary arc (> 1/5 perimeter).
                # This prevents an artifact segment from a source ring outside the
                # domain from stealing an adjacent exterior segment.
                if j_mls != current_mls:
                    next_start_xy = line_strings[j].coords[0]
                    gap = math.hypot(
                        ring_coords[-1][0] - next_start_xy[0],
                        ring_coords[-1][1] - next_start_xy[1],
                    )
                    if gap > threshold and d_fwd_close < perimeter / 5:
                        break  # close cannot be long; discard, leave j
                # Connect to segment j via forward boundary arc, then continue building
                # the ring. Using fwd_arc_corners (not arc_corners) guarantees that
                # corners are inserted in the forward direction, preserving the ring's
                # winding.
                ring_coords += [
                    *fwd_arc_corners(d_end, d_fwd_next),
                    *line_strings[j].coords,
                ]
                remaining_segs.pop(pos)
                remaining_segs_d.pop(pos)
                d_end = d_next_end
                current_mls = j_mls  # track mls of most recently added seg
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+        # as interiors so the box-difference path below inverts them correctly.
+        if interior_rings and exterior_rings:
+            interior_areas = [sgeom.Polygon(r).area for r in interior_rings]
+            max_interior_area = max(interior_areas)
+            reclassified = [r for r in exterior_rings
+                            if sgeom.Polygon(r).area < max_interior_area]
+            if len(reclassified) == len(exterior_rings):
+                for r in reclassified:
+                    exterior_rings.remove(r)
+                    interior_rings.append(r)
```
QuLogic, line 1236: Did I misread or isn't `reclassified == exterior_rings`? So that this is just:
```suggestion
                interior_rings += exterior_rings
                exterior_rings = []
```

lib/cartopy/crs.py
```diff
@@ -952,254 +986,255 @@ def _project_geometry_collection(self, geometry, src_crs):
...
+
+                if d_fwd_close <= d_fwd_next:
+                    # Ring start arrives first: close the ring.
+                    finish_ring(d_end, ring_coords, d_fwd_close, d_ring_start)
+                    break
+                else:
+                    # Segment j comes before our ring start.  If j is from a
+                    # different source ring and the gap to it is large, only
+                    # connect if the ring can't otherwise close without a very
+                    # long boundary arc (> 1/5 perimeter).  This prevents an
```
QuLogic, line 1183: Not sure I understand the 1/5 choice here.

lib/cartopy/crs.py
QuLogic: I don't know why GitHub made this suggestion look so weird. It's mostly just removing extra spaces and the `list`.

QuLogic:
Since #2698 is in now, you should also change `sgeom` to `shapely`.

greglucas:
Thanks for the review! I will be reworking this a little bit to try and take more of the d3-style approach in a similar vein to the adaptive-resampling work. I'm going to try and keep things similar here and only fix the boundary arc, but fix it with a more consistent way rather than the 1/5 heuristics you called out and add a few more individual tests for those periodic boundaries. It might take me a little bit more time but just wanted to leave an update.

PR #2660, opened 2026-04-04
TST: filter warnings from tests

greglucas:
We want to be notified of actual warnings/notes from upstream libraries, so this makes a few fixes that were being warned about while hiding other warnings that are out of our control.

With this I was able to get to no warnings locally.

Shapely:
* `resolution` was deprecated in favor of  `quad_segs` in buffer calls now.
* move geos_version import up a level to avoid warnings
* some of our nan tests emit shapely warnings which are expected

owslib:
* Lots of external warnings for lxml parsing which we have no control over, so just ignore those throughout the test runs.

Cartopy:
* we are warning about Gouraud shading, so we can assert against that warning
* remove OSGB precision warning. I personally don't think we need to warn what test precision we are working with and it clutters the log because I don't have that grid locally. I decided to remove the warning message, but can add it back if people feel strongly the other way. Note that the tests are seemingly more testing pyproj at this point since we are using pyproj for the transforms rather than our direct use of the PROJ library, so we have kind of punted those tests to pyproj IMO.

### DIFF
--- a/lib/cartopy/io/ogc_clients.py
+++ b/lib/cartopy/io/ogc_clients.py
@@ -155,7 +155,7 @@ def _target_extents(extent, requested_projection, available_projection):
     # the entire output requested_projection domain, then we erode the request
     # area to avoid re-projection instabilities near the projection boundary.
     buffered_target_box = target_box.buffer(requested_projection.threshold,
-                                            resolution=1)
+                                            quad_segs=1)
     fudge_mode = buffered_target_box.contains(requested_projection.domain)
     if fudge_mode:
         target_box = requested_projection.domain.buffer(
@@ -174,7 +174,7 @@ def _target_extents(extent, requested_projection, available_projection):
             # need to re-inflate.
             radius = min(max_x - min_x, max_y - min_y) / 5.0
             radius = min(radius, available_projection.threshold * 15)
-            poly = poly.buffer(radius, resolution=1)
+            poly = poly.buffer(radius, quad_segs=1)
             # Prevent the expanded request going beyond the
             # limits of the requested_projection.
             poly = available_projection.domain.intersection(poly)
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -11,7 +11,7 @@ import matplotlib.pyplot as plt
 import matplotlib.ticker as mticker
 import numpy as np
 import pytest
-from shapely.geos import geos_version
+from shapely import geos_version
 
 import cartopy.crs as ccrs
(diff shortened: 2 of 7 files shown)


rcomer:
Cycling to pick up #2659

rcomer (approved):
There are some other Cartopy deprecation warnings and some download related warnings showing in CI.  There's no reason not to merge this as-is though so I leave it up to you whether you want to look at those or self-merge this.

greglucas:
I could go either way on the download/regenerate warnings, those seem like they could be somewhat useful actually if you are wondering if you cleared your cache and wanted to know the files were redownloaded. (note those all weren't appearing for me locally because I already had the cached files) Do you have a preference?

```
lib/cartopy/tests/mpl/test_gridliner.py::test_grid_labels_inline[LambertConformal]
  /home/runner/work/cartopy/cartopy/lib/cartopy/io/__init__.py:242: DownloadWarning: Downloading: https://naturalearth.s3.amazonaws.com/110m_physical/ne_110m_coastline.zip
    warnings.warn(f'Downloading: {url}', DownloadWarning)

lib/cartopy/tests/test_img_nest.py::test_nest
  /home/runner/work/cartopy/cartopy/lib/cartopy/tests/test_img_nest.py:296: UserWarning: WMTS test data is out of date, regenerating at /home/runner/.local/share/cartopy/wmts/aerial.
    warnings.warn('WMTS test data is out of date, regenerating at '

lib/cartopy/tests/mpl/test_features.py::test_gshhs
  /home/runner/work/cartopy/cartopy/lib/cartopy/io/__init__.py:242: DownloadWarning: Downloading: https://www.soest.hawaii.edu/pwessel/gshhg/gshhg-shp-2.3.7.zip
    warnings.warn(f'Downloading: {url}', DownloadWarning)
```

Which runner do you see the deprecation warnings on? Those were the only ones I saw clicking into a few jobs.

rcomer:
Oh, I must have got unlucky with the one I clicked on!  It's in [Ubuntu py313](https://github.com/SciTools/cartopy/actions/runs/23998840944/job/69991470593?pr=2660#step:8:108).  From a quick search, we do not appear to be using those modules anywhere.  Maybe it's something to do with coverage?

rcomer:
> Do you have a preference?

I was thinking that if we get rid of all the warnings, we can turn on "raise warnings as errors" to get alerted to the real problems.  On second thoughts that might be annoying to have that on in CI given how often we have to work around other failures that the given PR didn't cause.

greglucas:
Sounds good on all fronts, I agree the raise on warnings was my initial goal but it does seem a bit ambitious. I am going to merge this now and we can do follow-ups later if desired.

PR #2664, opened 2026-04-11
MNT: use github mirror for gshhg data download

### LINKED ISSUE #2657: ngdc.noaa.gov gshhg url no longer valid
Seems this endpoint no longer exists. As a result cartopy can't download the coastline files.
https://www.ngdc.noaa.gov/mgg/shorelines/data/gshhg/latest

greglucas:
Switch to the GitHub url for downloading gshhg data. I've left the version option in even though it isn't used now. It may help in the future or if we want to revert to a different url. We could also just hardcode this since the dataset is unlikely to change if that is preferable, I don't have a major preference either way.

closes #2657


rcomer (approved):
Seems good to me, but I'll hold off from merging for a couple more days in case anyone else has an opinion.

Issue #2666, opened 2026-04-17
Nightlies failures: text issues


rcomer:
I just ran the tests against nightlies and we now have a lot more failures.
https://github.com/SciTools/cartopy/actions/runs/24556381161

So far I have only looked at a few images[^1], but they consistently have the text all bunched up.  E.g. for `test_ticks.py::test_set_yticks_cylindrical` we have

Baseline

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/46c37465-b569-44d0-91a6-f4820abcb3ea" />

Result

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/e13e781a-943c-489c-8250-784effb6424f" />

@QuLogic any ideas?

[^1]: Does `pytest-mpl` have any equivalent of Matplotlib's `triage_tests.py`?  I'm not sure how to look through the images efficiently.

rcomer:
Now I've looked through all the image failures.  Most of them just seem to be a tiny offset in the text that you would expect when changing the Freetype version.  But the bunching comes up in tick labels, some but not all of the `GridLiner` labels, annotate, contour labels, and weirdly even colorbar tick labels are affected - I don't get how Cartopy transforms could affect that.

`test_examples.py::test_contour_label`

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/99153648-285b-4b00-ab58-41a262f1647e" />

rcomer:
We also completely lost the titles in `test_gridliner.py::test_gridliner_title_adjust`

Baseline

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/1d9c75d8-74e7-43f5-9104-d7a201182416" />

Result

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/60d5a14d-0a14-4f70-8e01-68fa24e68d06" />

QuLogic:
> 1. Does `pytest-mpl` have any equivalent of Matplotlib's `triage_tests.py`?  I'm not sure how to look through the images efficiently. [↩](#user-content-fnref-1-9a6ffa25a3742a344e5cdcb9ddd5334f)

If you're running locally, then try [the `--mpl-generate-summary` argument](https://pytest-mpl.readthedocs.io/en/stable/usage.html#html-summary-reports). It's not quite as quick as `triage_tests.py` but should be close.

[...]

PR #2667, opened 2026-04-19
FIX: ignore invisible gridliner labels in title adjustment

rcomer:
## Rationale

Since https://github.com/matplotlib/matplotlib/pull/31285, `Text` objects return a null bbox if they are invisible or have an empty string.  This broke our title positioning logic, because we specifically look at the `ymax` of the Bbox rather than taking box unions.

The invalid title position gets incorporated into the axes tight bbox, which also becomes invalid and therefore messes up constrained layout and we get a plot like https://github.com/SciTools/cartopy/issues/2666#issuecomment-4273382421 where the titles have disappeared and the gridline labels overlap each other or go off the side of the plot.

This can all be avoided if we ignore the invisible or empty labels.  Running the relevant test locally with this branch and Elliot's branch from https://github.com/matplotlib/matplotlib/pull/31521, I got a sensible image

result

## Implications

This clearly won't help anyone who uses existing Cartopy releases with the new Matplotlib release.

### DIFF
--- a/lib/cartopy/mpl/geoaxes.py
+++ b/lib/cartopy/mpl/geoaxes.py
@@ -528,6 +528,8 @@ class GeoAxes(matplotlib.axes.Axes):
                 gl._draw_gridliner(renderer=renderer)
                 for label in (gl.top_label_artists +
                               gl.geo_label_artists):
+                    if not label.get_visible() or label.get_text() == "":
+                        continue
                     bb = label.get_tightbbox(renderer)
                     top = max(top, bb.ymax)
         if top < 0:


greglucas (approved):
This seems like a good thing to do anyways, but I'm wondering if we want to do the bbox union instead of max like you mentioned just below this to be more robust there too? That can be a future PR too, feel free to merge as is.

rcomer:
Looks like it is faster to only consider the ymax, although I only tried with one example.  Also most of the slowdown came from having the null bboxes in the union.

```
In [22]: def get_top():
    ...:     top = -1
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         if not label.get_visible() or label.get_text() == "":
    ...:             continue
    ...:         bb = label.get_tightbbox(renderer)
    ...:         top = max(top, bb.ymax)
    ...: 

In [23]: def get_top2():
    ...:     bbs = []
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         bbs.append(label.get_tightbbox(renderer))
    ...:         bb = mtransforms.Bbox.union(bbs)
    ...:         top = bb.ymax
    ...: 

In [24]: def get_top3():
    ...:     bbs = []
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         if not label.get_visible() or label.get_text() == "":
    ...:             continue
    ...:         bbs.append(label.get_tightbbox(renderer))
    ...:         bb = mtransforms.Bbox.union(bbs)
    ...:         top = bb.ymax
    ...: 

In [25]: %timeit get_top()
54.3 μs ± 113 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [26]: %timeit get_top2()
228 μs ± 667 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [27]: %timeit get_top3()
74.2 μs ± 165 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

QuLogic:
> ...:         bbs.append(label.get_tightbbox(renderer))
>     ...:         bb = mtransforms.Bbox.union(bbs)

Should you not be doing the union once at the end instead of each time in the loop?

rcomer:
> > ...:         bbs.append(label.get_tightbbox(renderer))
> > ...:         bb = mtransforms.Bbox.union(bbs)
> 
> Should you not be doing the union once at the end instead of each time in the loop?

oh yes 🤦‍♀️

rcomer:
I think I did it right this time, and only considering the ymax still comes out fastest, though these numbers are not large in the scheme of things: drawing the whole figure from the test with `draw_without_rendering`  takes nearly 2 seconds on my laptop!

```
In [23]: def get_top():
    ...:     top = -1
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         if not label.get_visible() or label.get_text() == "":
    ...:             continue
    ...:         bb = label.get_tightbbox(renderer)
    ...:         top = max(top, bb.ymax)
    ...: 

In [24]: def get_top2():
    ...:     bbs = []
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         bbs.append(label.get_tightbbox(renderer))
    ...:     bb = mtransforms.Bbox.union(bbs)
    ...:     top = bb.ymax
    ...: 

In [25]: def get_top3():
    ...:     bbs = []
    ...:     for label in (gl.top_label_artists +
    ...:                   gl.geo_label_artists):
    ...:         if not label.get_visible() or label.get_text() == "":
    ...:             continue
    ...:         bbs.append(label.get_tightbbox(renderer))
    ...:     bb = mtransforms.Bbox.union(bbs)
    ...:     top = bb.ymax
    ...: 

In [26]: %timeit get_top()
79.4 μs ± 699 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [27]: %timeit get_top2()
165 μs ± 524 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [28]: %timeit get_top3()
110 μs ± 565 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [29]: %timeit fig.draw_without_rendering()
1.92 s ± 21.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

PR #2668, opened 2026-04-20
fix: grid labels hidden on false overlap at rotation=90

### LINKED ISSUE #2394: Missing y-axis gridline labels when rotated to 90 degrees
### Description

Some y-axis gridline labels are missing when rotated to 90 degrees. This doesn't seem to happen when the rotation is 89.9 degrees or less. I only noticed this happening after upgrading Cartopy to 0.23.0.

#### Code to reproduce

```py
import cartopy.crs as ccrs


def plot_example_map(ylabel_rotation):
    fig = plt.figure()
    ax = fig.add_subplot(1, 1, 1, projection=ccrs.PlateCarree())
    ax.set_extent([80, 170, -45, 30], crs=ccrs.PlateCarree())

    ax.stock_img()

    ax.gridlines(
        draw_labels={"bottom": "x", "left": "y"},
        ylabel_style={"rotation": ylabel_rotation},
    )

    ax.coastlines()

    plt.tight_layout()
    plt.show()
```

#### Results

```py
# rotation = 90 degrees
plot_example_map(ylabel_rotation=90)
```

rotation = 90 degrees

```py
# rotation = 89.9 degrees
plot_example_map(ylabel_rotation=89.9)
```

image

<details>

### Operating system

Linux (Pop!_OS 22.04 LTS x86_64)

### Cartopy version

0.23.0

### pip list

```
cartopy
jupyterlab
```

### Environment

```sh
python3.11 -m venv .venv
source .venv/bin/activate
python -m pip install --upgrade pip setuptools wheel
python -m pip install cartopy jupyterlab
```

</details>

guillermodotn:
## Rationale

Gridline labels hidden at `rotation=90°` due to `intersects_path()` false positives on collinear edges.

Fixes #2394.

## Implications

Adds a `bbox` pre-check before `intersects_path()`; if `bboxes` don't overlap, skip the path check. Tilted geo labels unaffected.

### DIFF
--- a/lib/cartopy/mpl/gridliner.py
+++ b/lib/cartopy/mpl/gridliner.py
@@ -987,6 +987,7 @@ class Gridliner(matplotlib.artist.Artist):
                     # Updates
                     label.set_visible(visible)
                     label.path = this_path
+                    label.extents = this_path.get_extents()
                     label.xy = xylabel
                     label.loc = loc
                     self._labels.append(label)
@@ -1287,6 +1288,10 @@ class Label:
         return self.artist.get_visible()
 
     def check_overlapping(self, label):
+        # NOTE: Workaround for intersects_path() false positives on collinear edges.
+        # See https://github.com/matplotlib/matplotlib/issues/6076
+        if not self.extents.overlaps(label.extents):
+            return False
         overlapping = self.path.intersects_path(label.path)
         if overlapping:
             self.set_visible(False)
--- a/lib/cartopy/tests/mpl/test_gridliner.py
+++ b/lib/cartopy/tests/mpl/test_gridliner.py
@@ -519,6 +519,25 @@ def test_gridliner_save_tight_bbox():
     fig.savefig(io.BytesIO(), bbox_inches='tight')
 
 
+def test_gridliner_ylabel_rotation_90_tight_bbox():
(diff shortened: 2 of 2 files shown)


rcomer:
Thanks for this @guillermodotn.  My only worry is that calling `get_extents` here will slow down what is already a pretty slow draw.  Since we call `check_overlapping` on every pair of visible labels, the time increases with order N**2.  Could we instead call `get_extents` once for each label and attach the extents to the label alongside the path [here](https://github.com/SciTools/cartopy/blob/29ebd329d722c867f65031eec663e24ddb2e4a29/lib/cartopy/mpl/gridliner.py#L989)?  I tried locally by timing [this particularly slow test](https://github.com/SciTools/cartopy/blob/29ebd329d722c867f65031eec663e24ddb2e4a29/lib/cartopy/tests/mpl/test_gridliner.py#L522-L542) and it does seem to make a difference (unless I made a mistake...)

rcomer:
Also you have a typo in your commit/title: 9 -> 90

guillermodotn:
Thanks for pointing to where to cache it and the benchmark test.

Fixed both.

rcomer (approved):
Sorry, this fell off my to-do list.

Thanks for your work on this @guillermodotn!

PR #2672, opened 2026-04-26
MNT: don't access Formatter.locs

rcomer:
## Rationale

The `Formatter.locs` attribute will be [deprecated at Matplotlib 3.11](https://matplotlib.org/devdocs/api/next_api_changes/deprecations/31416-TH.html).  Since [`Formatter.set_locs` only assigns the variable we pass it](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/ticker.py#L291-L298) we can just use the variable we already have hold of.

### DIFF
--- a/lib/cartopy/mpl/ticker.py
+++ b/lib/cartopy/mpl/ticker.py
@@ -145,10 +145,13 @@ class _PlateCarreeFormatter(Formatter):
         self._target_projection = ccrs.PlateCarree(globe=self._source_projection.globe)
 
     def set_locs(self, locs):
-        Formatter.set_locs(self, locs)
+        super().set_locs(locs)
         if not self._auto_hide:
             return
-        self.locs, degs, mins, secs = self._get_dms(self.locs)
+        locs, degs, mins, secs = self._get_dms(locs)
+
+        # locs dtype may have changed.
+        super().set_locs(locs)
         secs = np.round(secs, self._precision - 3).astype('i')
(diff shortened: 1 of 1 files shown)


lib/cartopy/mpl/ticker.py
```diff
@@ -148,7 +148,10 @@ def set_locs(self, locs):
         Formatter.set_locs(self, locs)
         if not self._auto_hide:
             return
-        self.locs, degs, mins, secs = self._get_dms(self.locs)
+        locs, degs, mins, secs = self._get_dms(locs)
+
+        # locs dtype may have changed.
+        Formatter.set_locs(self, locs)
```
rcomer: I added this because we previously assigned the output of `_get_dms` to the attribute, but no tests fail if I don't do this.

greglucas: Kind of odd that this is not just setting locs but also doing all the auto_hide logic.

Should this be calling `super().set_locs(locs)` instead?

greglucas (approved):
Looks good as-is, but I did leave one potential nit suggestion since it stuck out to me. Feel free to self merge.

lib/cartopy/mpl/ticker.py
rcomer: Probably.  I was just following what was done on line 148 but, since that is 8 years old, maybe I shouldn't trust it!

PR #2676, opened 2026-05-18
FIX: remove internal wrapped collection when removing quadmesh

### LINKED ISSUE #2675: Memory leak when shifting `central_longitude`
I've started seeing this because I was running many plotting processes in parallel, and the memory usage was EXPLODING, but when isolating to a single for loop I can still see the same behaviour, which I narrowed down to using a `central_longitude` different than the standard. 

This is a minimal reproducible example (it downloads netcdf data and cache it in your temp directory). 

```python
import matplotlib
matplotlib.use("Agg")
import xarray as xr
import fsspec
from tqdm import tqdm
import pandas as pd
import cartopy.crs as ccrs
import numpy as np
import time
from aiohttp.client_exceptions import ServerDisconnectedError

dates = pd.date_range("2015-05-01", "2016-05-01")

fig, ax = plt.subplots(
    figsize=(12, 12),
    subplot_kw={
        "projection": ccrs.PlateCarree(central_longitude=180)
        },
)
ax.add_feature(
    cfeature.LAND.with_scale("110m"), edgecolor="black", linewidth=0.25, zorder=2
)

for date in tqdm(dates):
    url = f"https://www.star.nesdis.noaa.gov/pub/socd/mecb/crw/data/5km/v3.1_op/nc/v1.0/daily/sst/{date.year}/coraltemp_v3.1_{date.strftime('%Y%m%d')}.nc"
    while True:
        try:
            file = fsspec.open_local(
                f"simplecache::{url}", simplecache={"cache_storage": "/tmp/"}
            )
            ds = xr.open_dataset(file, engine="netcdf4").squeeze()
            break
        except ServerDisconnectedError:
            time.sleep(0.5)

    lon2d, lat2d = np.meshgrid(ds['lon'], ds['lat'])
    p = ax.pcolormesh(
        lon2d,
        lat2d,
        ds["analysed_sst"].values,
        transform=ccrs.PlateCarree(),
        cmap="jet",
        shading="nearest",
    )
    plt.savefig(f"frame_{date.strftime('%Y%m%d')}.png", bbox_inches="tight")
    p.remove()
```

Notice that I'm reusing the axis to avoid having to define the projection every time. If you let this run you will notice memory consumption steadily increasing. Here is a screenshot from istat (I know, that's not the most scientific way of doing memory profiling but you get the gist of it..).

Image
Image

Now, if I change the plot projection to `ccrs.PlateCarree()` and I let it run for the same amount of time, memory usage is constant. 

Image
Image

Weirdly enough, if I try to shift the data beforehand and use the same projection for the data and the plot 

```python
new_lon = ((ds.lon - 180 + 180) % 360) - 180
ds = ds.assign_coords(lon=new_lon).sortby("lon")

...

transform=ccrs.PlateCarree(central_longitude=180)
```

the memory leak is still there, which is weird because I thought the memory leak was related to the transformation due to the shifting of the 0 line. 

The issue seems to be present also on other crs, not just the default PlateCarree. 

I'm using cartopy = 0.24.0 and pyproj = 3.7.2. 

I didn't have time to investigate more where the memory leak is actually coming from. 

Do you have any suggestions?

greglucas:
GeoQuadMesh has two collections, the default collection and the wrapped collection. We weren't cleaning up the wrapped collection properly before because the default remove() method would only remove the actual colleciton and not the wrapped one we added.

closes #2675

### DIFF
--- a/lib/cartopy/mpl/geocollection.py
+++ b/lib/cartopy/mpl/geocollection.py
@@ -93,6 +93,13 @@ class GeoQuadMesh(QuadMesh):
         # Update color limits for the rest of the cells.
         super().set_clim(vmin, vmax)
 
+    def remove(self):
+        # Also remove the wrapped collection fix if it exists, otherwise it
+        # will be orphaned in the axes and cause a memory leak.
+        if hasattr(self, '_wrapped_collection_fix'):
+            self._wrapped_collection_fix.remove()
+        super().remove()
+
     def get_datalim(self, transData):
         # Return the corners that were calculated in
         # the pcolormesh routine.
--- a/lib/cartopy/tests/mpl/test_mpl_integration.py
+++ b/lib/cartopy/tests/mpl/test_mpl_integration.py
@@ -692,6 +692,45 @@ def test_pcolormesh_nan_wrap():
     assert len(pcolor.get_paths()) == 2
 
 
+def test_pcolormesh_remove():
+    # When a wrapped GeoQuadMesh is removed, its internal _wrapped_collection_fix
+    # must also be removed to prevent a memory leak.
+
+    # wrapped mesh
+    xs, ys = np.meshgrid([120, 160, 200], [-30, 0, 30])
+    data = np.ones((2, 2)) * np.nan
+
+    ax = plt.axes(projection=ccrs.PlateCarree())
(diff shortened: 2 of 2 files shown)


rcomer (approved):
The conda failure is unrelated.

PR #2684, opened 2026-06-26
MNT: Remove checks for old Shapely

QuLogic:
## Rationale

We already require Shapely 2.0, so no longer need to maintain compatibility code for older versions.

## Implications

Less code.


rcomer:
There is a ruff failure in this one.

dopplershift (approved):

QuLogic:
> There is a ruff failure in this one.

Fixed.

PR #2623, opened 2026-01-11
MNT: bump minimum dependency versions

rcomer:
## Rationale

Bump minimum supported python version to 3.11, and other dependency versions to something newer but always at least 2 years old.  I only looked at packages that we pin in the minimum version tests, although we do specify minimums for other packages.

## Implications

Dropping support for mpl 3.7 allows us to delete a bunch of old code :sunglasses:

Issue #2245, opened 2023-09-19
Using Gridliner and Figure(layout="constrained") together misbehaves


mx-moth:
### Description

I am trying to use the new constrained layout engine from matplotlib in combination with a GeoAxes with gridlines. I can not find a combination that works properly. All combinations I try fail in some way or another, some with poor styling and some with actual crashes.

I need to make plots of a specific size, and I would like the axes, titles, colour bars, etc to fill the available space. A `Figure(layout="constrained")` combined with `axes.set_aspect("equal", "datalim")` works perfectly for this. The figure is always of the correct size, and the axes are reshaped to fill the figure. By setting the appropriate xlim/ylim, or by using `axes.autoscale()` the elements to be plotted are arranged within the axes correctly.

I would also like to include grid lines on the axes. No matter which options I choose, I can not get the grid lines to behave.

#### Code to reproduce

I installed the latest versions of Python, cartopy, matplotlib:

```shell
$ conda install -c conda-forge python cartopy matplotlib
```

I then added the following to a script and ran it:

```python
#!/usr/bin/env python3

import cartopy.crs
import numpy
from matplotlib import pyplot as plt
from matplotlib.collections import PolyCollection

# Make a figure with a specific size
figure = plt.figure(figsize=(8, 4), layout="constrained")

# Make some axes that will fill the available space while maintaining correct aspect ratio
axes = figure.add_subplot(projection=cartopy.crs.PlateCarree())
axes.set_aspect(aspect='equal', adjustable='datalim')

# Add some polygon to the map, with a colour bar
collection = PolyCollection(
    verts=[
        [[0, 0], [1, 0], [1, 1], [0, 1]],
        [[1, 0], [2, 0], [2, 1], [1, 1]],
        [[0, 1], [1, 1], [1, 2], [0, 2]],
        [[1, 1], [2, 1], [2, 2], [1, 2]],
    ],
    array=[1, 2, 3, 4],
)
axes.add_collection(collection)
figure.colorbar(collection, ax=axes, location='right', label='index')

# Set up the axes data limits to keep the polygon in view 
axes.autoscale()

# Add some gridlines
gridlines = axes.gridlines(draw_labels=["bottom", "left"])

figure.savefig('figure.png')
```

This results in the grid lines not filling the available space correctly:

![figure](https://github.com/SciTools/cartopy/assets/132147/1bf119d9-88b9-445e-be5f-abe65f3a9c9f)

[...]

PR #2238, opened 2023-08-22
Consolidate mpl version checks

rcomer:
## Rationale

Closes #2237.  I should have been more explicit with those imports when I added them.

I'm wondering if it's worth moving all these mpl version checks to a central place, like [dask does with its numpy version checks](https://github.com/dask/dask/blob/5f456cc899890d11da0bf528412858e19665eccd/dask/array/numpy_compat.py#L10-L14).  I guess it depends how often we need this sort of version gating.

PR #2685, opened 2026-06-26
Update test figures for Matplotlib 3.11

### LINKED ISSUE #2666: Nightlies failures: text issues

QuLogic:
## Rationale

As we all know, Matplotlib changed text rendering, as well as some image resampling code, so tests are all failing when running against the latest version.

I went for replacing the images and adding a tolerance for older versions. This does result in _quite_ high tolerances in some cases, so it is possible we may wish to ship two images for some time instead.

## Implications

Compatibility with latest release of Matplotlib.

I also compressed the results with `oxipng -o max`, which saved:
```
Files processed: 57/57
Input size: 4.45 MiB (4663311 bytes)
Output size: 2.02 MiB (2121057 bytes)
Total saved: 2.42 MiB (54.52%)
```

Fixes #2666


QuLogic:
I've pushed a followup that changes tests that _were already affected_ to use the mpl20 style. This is more modern and also minimizes the RMS tolerance we need in most cases. Probably these two commits can be squashed, but I kept them separate if you wish to review them by themselves.

lib/cartopy/tests/mpl/test_ticks.py
```diff
@@ -6,29 +6,36 @@
...
+from cartopy.mpl import _MPL_311
 from cartopy.mpl.ticker import LatitudeFormatter, LongitudeFormatter
 
 
 @pytest.mark.natural_earth
-@pytest.mark.mpl_image_compare(filename='xticks_no_transform.png')
+@pytest.mark.mpl_image_compare(filename='xticks_no_transform.png', style='mpl20',
+                               tolerance=5.20 if not _MPL_311 else 0.5)
 def test_set_xticks_no_transform():
+    plt.rc('figure.subplot', left=0.125, right=0.9, bottom=0.1, top=0.9)
```
QuLogic, line 17: Note in `test_images`, I went with changing the `axes()` call to `add_axes` with this rect:
https://github.com/SciTools/cartopy/pull/2685/changes#diff-ef0d6ab6699444c9c1bdbb5f5003c73a10781b3d70cec64f3009344a4d3a647dR129

I don't know if one way looks better than the other.

lib/cartopy/tests/mpl/test_ticks.py
```diff
@@ -72,8 +85,11 @@ def test_set_yticks_non_cylindrical():
 
 
 @pytest.mark.natural_earth
-@pytest.mark.mpl_image_compare(filename='xyticks.png')
+@pytest.mark.mpl_image_compare(filename='xyticks.png', style='mpl20',
+                               tolerance=5.42 if not _MPL_311 else 0.5)
 def test_set_xyticks():
+    plt.rc('figure.subplot', left=0.125, right=0.9, bottom=0.1, top=0.9)
+    plt.rc('axes.formatter', limits=(-7, 7))
```
QuLogic, line 92: This prevents the y-axis from changing to a smallish number with a 1e6 multiplier.

lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/gridliner_constrained_adjust_datalim.png
rcomer, line 1: This came from #2245 where the axes layout was completely broken, so I think we don’t need the actual labels here and could use `text_placeholders`.

lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/gridliner_labels_tight.png
rcomer, line 1: This test makes sure `tight_layout` accounts for the gridliner, so could use `text_placeholders`.

lib/cartopy/tests/mpl/baseline_images/mpl/test_gridliner/gridliner_labels_title_adjust.png
rcomer, line 1: This tests that the titles land in the right position, so could use `text_placeholders`.

lib/cartopy/mpl/__init__.py
```diff
@@ -2,3 +2,10 @@
 #
 # This file is part of Cartopy and is released under the BSD 3-clause license.
 # See LICENSE in the root of the repository for full licensing details.
+
+import matplotlib
+import packaging.version
```
rcomer, line 7: We removed `packaging` from the general dependencies at #2623.

rcomer (commented):
Thanks for doing this @QuLogic.

I think we can get rid of the largest of these tolerances by using your `text_placeholders` fixture (see my comments on the images).  I'm not sure about the others: I never got the hang of what counts as too high for a tolerance.

lib/cartopy/mpl/__init__.py
QuLogic: Good point, maybe we only want it for tests?

QuLogic:
Ah, it looks like the placeholders didn't handle rotation, so the boxes stick together a little differently on macOS. I'll try and see if I can get the placeholders to rotate as well.

lib/cartopy/mpl/__init__.py
rcomer: It has moved about.  When I arrived it was just in `lib/cartopy/tests/mpl/__init__.py `.

I think #2169 started using `packaging` in the main code (and forgot to add it to deps).  Then mpl v3.8 changes needed a lot of version gating in the main code #2238.

rcomer:
Will we always need to vendor the fixture or will we be able to import it from Matplotlib when we require a high enough  version?

QuLogic:
After a bit of CI debugging, I've found that on Ubuntu (x86_64) vs macOS (arm64), there are some small differences in the map boundary and subsequent gridline calculations. The map boundary difference is only somewhere below the 12th decimal place though, so that would seem not very relevant. What does differ more significantly is the calculated `MultiLineString` of the intersection of the latitude line with the map boundary.

On Ubuntu (x86_64), the 20°S, 0, and 40°N lines are split into 4 segments, whereas 20°N, 60°N, and 80°N are only 2 segments. On the hand, with macOS (arm64), the 60°N and 80°N lines are also split into 4. This occurs roughly at the ends +/-180° longitude, which on the LCC projection for the test is at about 60° from the top-left edge. You can see a comparison here:
projections

It may be a bit difficult to see, but there are a few extra colours along that split for the two innermost latitudes. For `MultiLineString`, the code then picks up to the eighth-last (not entirely sure why that limit; it should maybe be a distance long the path?) element as a possible position (see the last `for` loop below). When we have 4 parts, it picks a position in one of the small ones, and some slightly different points in the longer one than it would if there were only 2 parts. This causes the labels to seem to overlap, and get turned off.

https://github.com/SciTools/cartopy/blob/c1b512ae768dda9a24bca4de7e4360758620c7d8/lib/cartopy/mpl/gridliner.py#L813-L839

Now, it turns out that when we get 4 parts, each smaller part is just two elements long and an end is coincident with an end of a longer part. I'm uncertain why we end up with these extra parts (as at least the transformed latitude line is the same on both platforms, and the split is nowhere near the map boundary), but if we merge them together, then we can get consistent potential-label-positions on all platforms and they end up all being visible in the same way. As you can see in the above code, we _already_ merge the lines if there are more than 4 parts.

I do not have a strong idea of whether we should always have _one_ `LineString`, so I have only applied the merge everywhere, and not the coalescence into a single one. The only loss is to Robinson where the top 180 labels have gone missing.

QuLogic:
> Will we always need to vendor the fixture or will we be able to import it from Matplotlib when we require a high enough version?

I didn't end up applying any rotation changes, so yes, we can import it when we require a high enough version.

QuLogic:
> The only loss is to Robinson where the top 180 labels have gone missing.

Sigh, and now macOS produces a different result for that one... Let me see what the calculations are like.

QuLogic:
So I guess I see this Robinson case is the reason why we have that merging code for `MultiLineString`. The 180 lines are nearly coincident with the boundary and produce ~90-part `MultiLineString` of mostly two-point `LineString`s. Unfortunately, the `line_merge` appears to sort them in a different order, which is in line with [the "mild" canonical form](https://shapely.readthedocs.io/en/stable/geometry.html#canonical-form) that Shapely produces, but causes us to pick slightly different spots to attempt to label. I have tried [calling `normalize` on the result](https://github.com/QuLogic/cartopy/commit/75dac262d86e4f3fe52ce58eddf3be5f57ac551b) before  the manual concatenation, but it's still inconsistent. I'm beginning to wonder if there's a Shapely or GEOS bug across platforms now.

QuLogic:
OK, I've had to isolate the changes in the last commit, so that if after line merging we _still_ have to combine all the sub-geometries manually, then do that on the pre-line-merged intersection. This avoids the platform-specific re-ordering that Shapely/GEOS seems to be doing. If I can make a reproducer, I will report this to whichever upstream, but in the meantime, this will allow us to get tests working.

lib/cartopy/mpl/__init__.py
rcomer: I don't really have an opinion whether this should be moved to the test module, or we should have `packaging` as a main dependency again.

rcomer:
Thanks for the detailed explanations @QuLogic.  These changes all makes sense to me.  I think we just need to decide what to do with the `packaging` import and then this is good to go.

lib/cartopy/mpl/__init__.py
QuLogic: Matplotlib (since 3.9) depends on it and we depend on >=3.8, so while there's a slight mismatch, I think it should be fine for us to keep an unconditional dependency for now.

rcomer (approved):
Please handle the merge however you would like.

QuLogic:
I re-arranged the commits a little bit so that the image changes could be minimized. The main one is squashing the `mpl20` style change into the 3.11 update, but also setting up the placeholders before the update as well.

PR #2687, opened 2026-06-26
TST: explicitly add pip in conda environment

rcomer:
## Rationale

Recently the conda tests have been failing in the Cartopy install step.  I notice that `pip` is not in the conda list in these runs, and the dependencies are getting pip-installed in the Cartopy install step.  Perhaps we are using the wrong `pip`, so this PR is an attempt to make sure we get `pip` in the conda environment and use it.


rcomer:
Conda tests are green ✅

QuLogic (approved):

PR #2688, opened 2026-06-28
FIX: rotate text placeholders

rworkman4:
This fixes the macOS image-test regression by making the shared text placeholder fixture respect text rotation.

Root cause:
- The placeholder rectangles used for layout-sensitive MPL tests were drawn without rotation.
- That caused small geometry drift on macOS, which showed up in the `ci / lint-and-test` run.

Checks:
- `python -m py_compile lib/cartopy/tests/mpl/conftest.py`
- `python -m ruff check lib/cartopy/tests/mpl/conftest.py`


lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -542,8 +542,10 @@ def test_gridliner_ylabel_rotation_90_tight_bbox():
 
 
 @pytest.mark.natural_earth
-@pytest.mark.mpl_image_compare(filename='gridliner_labels_title_adjust.png',
-                               style='mpl20')
+@pytest.mark.mpl_image_compare(
+    filename='gridliner_labels_title_adjust.png', style='mpl20',
+    tolerance=13.0,
```
dopplershift, line 547: Do you have the example image showing what's going on to need this large tolerance?

rcomer:
Closing for lack of response/update and because the target branch has moved on.  We did not need rotation after all.

PR #2693, opened 2026-07-05
PERF: vectorize shapely_to_path coordinate extraction

greglucas:
Rather than pulling out each sub-geometry .xy's coordinates, we can do one get_coordinates() call and fill in the codes all at once. Using cumsum to incrementally keep track of which ring we are in for the code placement.

I get about 50-100x speedup in the following example.

```python
import time
import numpy as np, shapely.geometry as sgeom
from cartopy.mpl.path import shapely_to_path

t = np.linspace(0, 2 * np.pi, 50_000)
holes = [np.column_stack([np.cos(t[::100]), np.sin(t[::100])]) * 2 + [x, 0]
            for x in np.linspace(-40, 40, 20)]
poly = sgeom.Polygon(np.column_stack([np.cos(t), np.sin(t)]) * 100, holes)

t0 = time.perf_counter()
shapely_to_path(poly)
t1 = time.perf_counter()
print(f"Time taken: {t1 - t0:.5f} seconds")
```

### DIFF
--- a/lib/cartopy/mpl/path.py
+++ b/lib/cartopy/mpl/path.py
@@ -14,6 +14,7 @@ and `Matplotlib Path API <https://matplotlib.org/stable/api/path_api.html>`_.
 
 from matplotlib.path import Path
 import numpy as np
+import shapely
 import shapely.geometry as sgeom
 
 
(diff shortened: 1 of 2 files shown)


lib/cartopy/mpl/path.py
```diff
@@ -106,20 +107,20 @@ def shapely_to_path(shape):
     if shape.is_empty:
         return Path(np.empty([0, 2]))
     elif isinstance(shape, sgeom.LinearRing):
-        return Path(np.column_stack(shape.xy), closed=True)
+        return Path(shapely.get_coordinates(shape), closed=True)
```
greglucas, line 110: get_coordinates() defaults to 2d arrays, so we can use it directly here even if someone passes a geometry with Z.

https://shapely.readthedocs.io/en/2.1.2/reference/shapely.get_coordinates.html

lib/cartopy/tests/mpl/test_path.py
```diff
@@ -102,3 +103,32 @@ def test_non_polygon_path_closing(self, path):
...
         assert isinstance(closed_path, Path)
         assert closed_path.vertices.size == 0
+
+
+class Test_shapely_to_path:
+    def test_polygon_with_multiple_interiors(self):
+        exterior = sgeom.box(0, 0, 12, 12).exterior.coords
+        interiors = [sgeom.box(1, 1, 2, 2, ccw=False).exterior.coords,
+                    sgeom.box(4, 4, 5, 6, ccw=False).exterior.coords,
+                    sgeom.box(8, 8, 9, 10, ccw=False).exterior.coords]
```
QuLogic: Funny alignment.

QuLogic (approved):

lib/cartopy/tests/mpl/test_path.py
greglucas: Fixed it up, relying on ruff auto-formatting too much anymore and I forgot it isn't on this project.

PR #2698, opened 2026-07-14
Drop use of backcompat modules

QuLogic:
## Rationale

See individual commits for details, but the main points are:
- pyproj made `CustomConstructorCRS` public in 3.2, and we depend on 3.6.
- Many symbols in deeper modules have been exposed at the top-level `shapely` level with 2.0. While the former haven't been removed, they are no longer documented.

## Implications

Keeping up with documented interfaces makes it easier to cross-references and follow along with upstream examples/documentation. This also _fixes_ the cross-references in Sphinx to `shapely` stuff.

### DIFF
--- a/lib/cartopy/mpl/patch.py
+++ b/lib/cartopy/mpl/patch.py
@@ -16,7 +16,7 @@ import warnings
 
 from matplotlib.path import Path
 import numpy as np
-import shapely.geometry as sgeom
+import shapely
 
 import cartopy.mpl.path as cpath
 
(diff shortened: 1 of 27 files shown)


QuLogic:
I guess I forgot there's no automated rebuilds without Meson...

lib/cartopy/mpl/patch.py
```diff
@@ -144,9 +145,10 @@ def path_to_geos(path, force_ccw=False):
     Returns
     -------
     A list of instances of the following type(s):
-        :class:`shapely.geometry.polygon.Polygon`,
-        :class:`shapely.geometry.linestring.LineString` and/or
-        :class:`shapely.geometry.multilinestring.MultiLineString`.
+
+    - :class:`shapely.Polygon`,
+    - :class:`shapely.LineString` and/or
+    - :class:`shapely.MultiLineString`.
```
rcomer: These have come out with a double bullet in the built docs.  Maybe they need more indentation?  These functions are anyway deprecated in favour of the equivalents in the `path` module - do you want to put bulleted lists in those docstrings?

lib/cartopy/trace.pyx
rcomer, line 1: I assume we can also get rid of `from __future__ import print_function`.

lib/cartopy/trace.pyx
```diff
@@ -247,7 +244,7 @@ cdef State get_state(const Point &point, object gp_domain, bool geom_fully_insid
         # Fast-path return because the geometry is fully inside
         return POINT_IN
     if isfinite(point.x) and isfinite(point.y):
-        state = POINT_IN if shapely.intersects_xy(gp_domain.context, point.x, point.y) else POINT_OUT
+        state = POINT_IN if shapely.intersects_xy(gp_domain, point.x, point.y) else POINT_OUT
```
rcomer, line 245: This is the only bit I’m not sure about as I am struggling to find information about what the `context` attribute is/was.  I assume it previously provided extra information that is now a available directly from the input geometry?

rcomer (commented):
Nice cleanup!

lib/cartopy/trace.pyx
QuLogic: Quite likely, yes.

lib/cartopy/trace.pyx
QuLogic: It's added onto a prepared geometry to reference the original:
https://github.com/shapely/shapely/blob/264792832fc6c2d2df6c931ef8178146d4897ab8/shapely/prepared.py#L28

But we shouldn't need that, as Shapely should select the prepared/not prepared geometry itself automatically.

rcomer (approved):

PR #2699, opened 2026-07-16
TST: Drop unnecessary explicit baseline image names

QuLogic:
## Rationale

Since pytest-mpl saves its results with the test name, it can be confusing to correlate the result with the original baseline image, unless you go through the test files at the same time. There's no reason to keep the original names other than perhaps shortening the original migration PR.

However, when tests reuse baseline images, they do need to use the explicit name. Note that this applies to parametrized tests that output the same result. I have still renamed those so that they'd match.

Also, drop some tolerances that were less than the default of 0.5.

## Implications

It's a bit clearer when looking at pytest-mpl's results which test image was broken.


QuLogic:
Also combined all the `test_feature_artist` tests into one.

lib/cartopy/tests/mpl/test_mpl_integration.py
```diff
@@ -920,7 +916,7 @@ def test_quiver_regrid_with_extent():
 
 @requires_scipy
 @pytest.mark.natural_earth
-@pytest.mark.mpl_image_compare(filename='barbs_plate_carree.png')
+@pytest.mark.mpl_image_compare
 def test_barbs():
```
rcomer: It might be nice to make this `test_barbs_plate_carree`, consistent with the `test_quiver_plate_carree` above.

rcomer (commented):
I agree, I have also been finding it slightly awkward to navigate the baseline images because of inconsistent naming in places.

I have one small suggestion, but I do not feel strongly about it.

rcomer:
I don't understand why coveralls is unhappy.

QuLogic:
Not sure, but it seems happy this time.

rcomer (approved):

PR #2700, opened 2026-07-16
Update linting with prek, yamllint and zizmor

QuLogic:
## Rationale

Like Matplotlib did before, switch from pre-commit to prek. Also add yamllint and zizmor, fixing issues pointed out by them.

## Implications

Faster pre-commit checks, more secure workflows, etc.

However, due to https://github.com/zizmorcore/zizmor/issues/2197 we may not want to merge this just yet. I've put it in draft for now.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every weekday
       interval: "daily"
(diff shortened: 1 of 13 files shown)


QuLogic:
I rebased and added an ignore for zizmor; this can be removed after their next release.

greglucas (approved):

PR #2717, opened 2026-08-10
ci: Set default shell to bash everywhere

### LINKED ISSUE #2715: [TST] Upcoming dependency test failures
The build with nightly wheels from matplotlib, pyproj, scipy, shapely and
their dependencies has failed. Check the logs for any updates that need to
be made in cartopy.
https://github.com/SciTools/cartopy/actions/runs/31348330864

QuLogic:
## Rationale

This should fix the nightly builds as well.

## Implications

Fixes #2715

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -11,6 +11,10 @@ on:
 
 permissions: {}
 
+defaults:
+  run:
+    shell: bash
+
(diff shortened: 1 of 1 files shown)


QuLogic:
Looks like we need to use a login shell for conda still, so added a comment about it.

dopplershift (approved):

Issue #2279, opened 2023-10-18
transform_vector is wrong / not linear for some projections


csubich:
### Description

I'm using (attempting to use) `cartopy.crs.transform` to transform map-coordinate wind vectors from one projection to another, currently from a rotated lat-lon projection to an ordinary lat-lon projection.  In so doing, I could never quite get the wind speeds to match up after converting from (presumably) degrees/sec back to meters/second.

In drilling down to simpler and simpler reproductions, I've found that for at least `cartopy.crs.RotatedPole`, `transform_vector` _isn't even linear_.  I could understand some scale-based difference between the output of `transform_vector` and a finite difference approximation via `transform_points`, but I can't think of any reason that the transformation should behave nonlinearly.

#### Code to reproduce

```python
import cartopy
import numpy as np
print(cartopy.__version__)
latlon_crs = cartopy.crs.PlateCarree()
rot_crs = cartopy.crs.RotatedPole(pole_longitude=180,pole_latitude=45.0)
azero = np.array([0],dtype=np.float32)
aone = np.array([1],dtype=np.float32)

print('[1,0] transformed', np.concatenate(latlon_crs.transform_vectors(rot_crs,azero,azero,aone,azero)))
print('[0,1] transformed', np.concatenate(latlon_crs.transform_vectors(rot_crs,azero,azero,azero,aone)))
print('[1,1] transformed', np.concatenate(latlon_crs.transform_vectors(rot_crs,azero,azero,aone,aone)))

# Via finite difference
def do_fd(x,y):
    fd_eps = 1e-2
    fd_origin = rot_crs.transform_points(latlon_crs,azero,azero)[0,:2]    
    fd_delta = (rot_crs.transform_points(latlon_crs,azero+fd_eps*x,azero+fd_eps*y)[0,:2] - fd_origin)/fd_eps
    return fd_delta

print('---')
    
print('[1,0] via finite difference', do_fd(1,0))
print('[0,1] via finite difference', do_fd(0,1))
print('[1,1] via finite difference', do_fd(1,1))
```

#### Output

```
0.22.0
[1,0] transformed [ 1.00000000e+00 -6.17066907e-06]
[0,1] transformed [-6.1880869e-08  1.0000000e+00]
[1,1] transformed [1.15470764 0.81648649]
---
[1,0] via finite difference [1.41421352e+00 8.72664586e-05]
[0,1] via finite difference [0.         0.99999998]
[1,1] via finite difference [1.41396673 1.00008723]
[...]

PR #2718, opened 2026-08-10
Fix vector transform linearity for angular coordinate systems

domfrbassett:
## Rationale

`CRS.transform_vectors()` estimates output directions by perturbing each point in the direction of the full input vector. This can make the result non-linear in the input components. For example, in a `RotatedPole -> PlateCarree` transform, transforming `(1, 0) + (0, 1)` does not necessarily match transforming `(1, 1)`.

This changes angular/geodetic-like source or target CRS cases to estimate the transformed local eastward and northward basis directions separately, then apply the resulting local rotation to the input vector components. The existing approximation is retained as a fallback for projected-to-projected and singular cases.

## Implications

This improves vector directions for rotated latitude/longitude and near-pole cases used by quiver, barbs, and streamplot. The affected image baselines were updated.

The new path is slower because it uses two basis perturbations rather than one vector-direction perturbation. In local benchmarks it was about 1.5–1.8x slower for affected cases, but still sub-millisecond for typical quiver-sized fields.

## Validation

- Added regression tests for vector linearity and near-pole PlateCarree vectors.
- Ran relevant CRS/vector/matplotlib tests: `65 passed`.
- Ran `ruff check` on changed Python files.


QuLogic:
@greglucas Does this relate to #1920?

greglucas:
Thanks for reminding me about those PRs @QuLogic, they had dropped off my radar. This PR is an improvement over #1920, but I still think it can be generalized even more and doesn't need the fallback/cutout. I've updated the other PRs I had addressing that as an alternative to this approach if anyone wants to take a look at those generalizations in #2725.

domfrbassett:
@greglucas I agree #2725 is the more general version of this fix. I’ll take a look there and see whether I can help with the remaining test/benchmark coverage, especially for the #2279 linearity case.

PR #2720, opened 2026-08-12
Enable testing and building of freethreaded/3.15 wheels

QuLogic:
## Rationale

Nothing seems to prevent this from working, AFAICT.

## Implications

Users can install wheels on 3.14t/3.15t.


QuLogic:
Also added 3.15 testing, as cibuildwheel just enabled it since it's in rc and #2719 adds it for us.

QuLogic:
Hmm, looks like we're blocked by pyproj right now https://github.com/pyproj4/pyproj/issues/1572

QuLogic:
And for 3.15, https://github.com/pyproj4/pyproj/issues/1621

.github/workflows/ci-testing.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97  # v7.0.0
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: ${{ startsWith(matrix.python-version, '3.15') }}
```
greglucas: Do we need the restriction here or can we just let it be True so we don't have to bump this each time?

.github/workflows/ci-testing.yml
```diff
@@ -97,6 +98,9 @@ jobs:
         if: matrix.use-network
         # we need to force bash to use line continuations on Windows
```
greglucas: I just noticed this when looking, but it looks like this comment may be out of date or out of place?

greglucas (commented):
I think we should have pyproj wheels now.

Do we need to think about what we want to say we support in terms of free threading? i.e. should we be putting locks around the entry/exit to trace or geometry transforms.

.github/workflows/ci-testing.yml
QuLogic: I suppose not, since we have a fixed set of releases listed, so we shouldn't accidentally get 3.16 or anything.

.github/workflows/ci-testing.yml
QuLogic: Should've been removed in #2717.

QuLogic:
> Do we need to think about what we want to say we support in terms of free threading? i.e. should we be putting locks around the entry/exit to trace or geometry transforms.

Well, I marked it Beta so I wouldn't need to think about it too hard just yet :wink:

QuLogic:
Looks like we need Matplotlib wheels (working on that now, though it doesn't break anything yet) and Shapely wheels, being debated here https://github.com/shapely/shapely/issues/2413

QuLogic:
Matplotlib wheels are ready now, and Shapely has them on the nightly index, so I've pulled them from there for now.

greglucas (approved):

PR #2726, opened 2026-09-03
Fix issue #2562 - incorrect documentation on PlateCarree

### LINKED ISSUE #2562: PlateCarree documentation incorrect
The documentation for the PlateCarree class is defined with two parameters:
cartopy.crs.PlateCarree(central_longitude=0.0, globe=None)

But the parameter description lists "proj4_params" and "globe".  Instead of "proj4_params" I believe it should list "central_longitude".

Parameters:
proj4_params (iterable of key-value pairs) – The proj4 parameters required to define the desired CRS. The parameters should not describe the desired elliptic model, instead create an appropriate Globe instance. The proj4_params parameters will override any parameters that the Globe defines.
globe – If omitted, the default Globe instance will be created. 

https://cartopy.readthedocs.io/stable/reference/projections.html#cartopy.crs.PlateCarree

DavidVadnais:
## Rationale
Fixes: #2562

Im making this change to improve the documentation in cartopy.

## Implications
Should be low.  Just doc strings.

### DIFF
--- a/lib/cartopy/crs.py
+++ b/lib/cartopy/crs.py
@@ -1957,9 +1957,27 @@ class LambertAzimuthalEqualArea(Projection):
 
 
 class Miller(_RectangularProjection):
+    """
+    A Miller cylindrical projection.
+
+    This projection is a modified Mercator projection where polar distortion
+    is reduced as compared to Mercator. It is is a useful compromise
+    for world maps between equal-area and conformal but it isnt either.
+
+    """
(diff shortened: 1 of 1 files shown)


QuLogic (approved):

QuLogic:
Thanks @DavidVadnais! Congratulations on your first PR to Cartopy :tada: We hope to hear from you again.

PR #2733, opened 2026-09-09
ENH: nudge every azimuth PROJ cannot build in ObliqueMercator

greglucas:
PROJ cannot build an omerc whose centre line runs due east or west, and that covers every azimuth congruent to 90 modulo 180, not just 90 itself: -90, 270 and 450 all fail at some central latitudes.

Also add tests against a real hotine EPSG definition.


lib/cartopy/crs.py
```diff
@@ -3260,9 +3261,12 @@ def __init__(self, central_longitude=0.0, central_latitude=0.0,
 
         """
 
-        if np.isclose(azimuth, 90):
-            # Exactly 90 causes coastline 'folding'.
-            azimuth -= 1e-3
+        # PROJ cannot build an omerc whose centre line runs due east or west:
+        # its origin comes from asin(G * tan(gamma0)), and at sin(alpha) = 1
+        # the tangent overruns the outer asin. Nudge off the singular value.
```
QuLogic: This comment leaves out something, because it says the second equation causes an overrun in the first, but that one doesn't directly include the second.

greglucas: OK, I made it a bit more generic and left out the math and just said we need to nudge away from singular values now.

QuLogic (approved):

PR #2738, opened 2026-09-09
Auto-detect RGB vs RGBA tile format instead of hardcoding it

### LINKED ISSUE #2729: Add auto-detect tile helper for loading as RGB vs RGBA
Rather than dealing with having users specify the desired tile forms like in #2728 I think we could add a helper to auto-load the image into the format needed instead. We would likely need to change all the `desired_tile_format=None` to intercept when someone explicitly wants a tile form and respect that, but when it is None we could just auto-load properly based on whether the loaded image has an alpha channel.

prototype example:

```python
def detect_and_convert(image_path):
    img = Image.open(image_path)
    
    # 1. Direct alpha check for RGBA, LA (Luminance + Alpha), etc.
    # 2. Check for GIF/PNG Palette images ('P') with transparency metadata
    has_alpha = "A" in img.mode or (img.mode == "P" and "transparency" in img.info)
    
    if has_alpha:
        return img.convert("RGBA")
    else:
        return img.convert("RGB")
```

greglucas:
desired_tile_form now defaults to None on GoogleWTS and all its subclasses (and on NestedImageCollection). When left as None, the format is detected per image from its PIL mode/info: RGBA if it carries an alpha channel or palette transparency, otherwise RGB. Explicitly passing "RGB" or "RGBA" still forces that form for every tile.

Closes #2729

### DIFF
--- a/lib/cartopy/io/__init__.py
+++ b/lib/cartopy/io/__init__.py
@@ -61,6 +61,19 @@ def fh_getter(fh, mode='r', needs_filename=False):
     return fh, filename
 
 
+def _ensure_tile_form(img, desired_tile_form=None):
+    """Convert a PIL image to ``desired_tile_form``.
+
+    If ``desired_tile_form`` is None, auto-detect: 'RGBA' if the image
+    has an alpha channel or palette transparency, else 'RGB'.
+    """
+    if desired_tile_form is None:
+        has_alpha = 'A' in img.mode or (
+            img.mode == 'P' and 'transparency' in img.info)
+        desired_tile_form = 'RGBA' if has_alpha else 'RGB'
+    return img.convert(desired_tile_form)
+
+
 class DownloadWarning(Warning):
     """Issued when a file is being downloaded by a :class:`Downloader`."""
     pass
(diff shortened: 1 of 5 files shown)


lib/cartopy/io/__init__.py
```diff
@@ -61,6 +61,19 @@ def fh_getter(fh, mode='r', needs_filename=False):
...
 
 
+def _ensure_tile_form(img, desired_tile_form=None):
+    """Convert a PIL image to ``desired_tile_form``.
+
+    If ``desired_tile_form`` is None, auto-detect: 'RGBA' if the image
+    has an alpha channel or palette transparency, else 'RGB'.
+    """
+    if desired_tile_form is None:
+        has_alpha = 'A' in img.mode or (
```
QuLogic: This would also match `LAB` and miss `RGBa` & `La`.

greglucas: Updated with an explicit list of options now. Also apparently La can't be directly converted to RGBA, so there is now an extra conversion step to go to LA first.

QuLogic (approved):

PR #2743, opened 2026-09-15
FIX: respect gridliner visibility and inherited rotation defaults

greglucas:
Honor the parent artist visibility when drawing and positioning titles, and clear stale state after a draw.

OSGB/OSNI inherited a different rotate labels than they should have which this fixes. I chose not to update the image files and instead force the image comparison test to use the old rotate_labels choice. We test the rotate_labels assertions in the added non-image tests now.


lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -640,3 +646,46 @@ def test_gridliner_geo_labels_respect_side_visibility():
...
     assert labels == ['60°S', '60°N']
+
+
+def test_gridliner_visibility_after_draw():
+    fig = plt.figure()
+    ax = fig.add_subplot(projection=ccrs.PlateCarree())
+    ax.set_global()
+    ax.set_title('Map title')
+    fig.canvas.draw()
+    bare = np.asarray(fig.canvas.buffer_rgba()).copy()
```
QuLogic:
```suggestion
    bare = np.asarray(fig.canvas.buffer_rgba(), copy=True)
```

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -640,3 +646,46 @@ def test_gridliner_geo_labels_respect_side_visibility():
...
+    bare_title = ax.title.get_position()
+
+    grid = ax.gridlines(draw_labels=True)
+    fig.canvas.draw()
+    assert not np.array_equal(bare, np.asarray(fig.canvas.buffer_rgba()))
```
QuLogic: This should automatically cast, I think:
```suggestion
    assert not np.array_equal(bare, fig.canvas.buffer_rgba())
```

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -640,3 +646,46 @@ def test_gridliner_geo_labels_respect_side_visibility():
...
+    assert not grid.stale
+
+    grid.set_visible(False)
+    fig.canvas.draw()
+    np.testing.assert_array_equal(bare, np.asarray(fig.canvas.buffer_rgba()))
```
QuLogic: Ditto.
```suggestion
    np.testing.assert_array_equal(bare, fig.canvas.buffer_rgba())
```

lib/cartopy/tests/mpl/test_gridliner.py
```diff
@@ -640,3 +646,46 @@ def test_gridliner_geo_labels_respect_side_visibility():
...
+    assert ax.title.get_position() == bare_title
+
+    grid.set_visible(True)
+    fig.canvas.draw()
+    assert not np.array_equal(bare, np.asarray(fig.canvas.buffer_rgba()))
```
QuLogic:
```suggestion
    assert not np.array_equal(bare, fig.canvas.buffer_rgba())
```

QuLogic (approved):

lib/cartopy/tests/mpl/test_gridliner.py
greglucas: Updated.

lib/cartopy/tests/mpl/test_gridliner.py
greglucas: Good call, this and all of the below are updated.

lib/cartopy/tests/mpl/test_gridliner.py
greglucas: Apparently asarray doesn't have the copy keyword in our minimum version. I changed to a bare `array(..., copy=True)` instead here.

PR #2744, opened 2026-09-16
Deprecate old lat/lon tick formatters

QuLogic:
## Rationale

This is a re-creation of #1066, but with a completely new implementation using proper module-level deprecation.

## Implications

Closes #1066


rcomer (approved):
Well I've learned something.  I didn't know you could do this but I've played about with it locally and, clearly, it works.

QuLogic:
We do [this in Matplotlib as well](https://github.com/matplotlib/matplotlib/blob/main/lib/matplotlib/_api/__init__.py#L288), but it goes through the slightly more complicated route of decorating a class that gets turned into the function.

